PR #4, opened 2017-05-21
JAV-17 Replace package name from com.huawei.paas.cse to io.servicecomb.

TankTian:


seanyinx:
the package renaming was done and in master

PR #11, opened 2017-05-26
[issueid # JAV-43] spring boot tranport demo code optimized as using …

acsukesh:
…embedded servlet we directly use spring boot container and few modules not supported in open source branch so removed unused code like highwat,tcc..and added AT code

Signed-off-by: sukesh a c <sukeshac@huawei.com>


demo/demo-spring-boot-transport/demo-spring-boot-pojo-client/src/main/java/io/servicecomb/demo/TestMgr.java
```diff
@@ -32,9 +32,9 @@
  * @since  [产品/模块版本]
  */
 public class TestMgr {
-    private static Logger LOGGER = LoggerFactory.getLogger(TestMgr.class);
```
seanyinx: this is a duplicate file of the one in demo-schema. please remove this file.

PR #13, opened 2017-05-27
JAVA-49 publish java-chassis binary version package.

TankTian:


java-chassis-distribution/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.servicecomb</groupId>
+            <artifactId>cse-transport-rest-vertx</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
```
seanyinx: do we only assembly a zip package when releasing a version? if so, it's better to put this plugin in a release profile like the parent pom

TankTian: i has add profile. when you need release version, you can use command: mvn -P release .....

PR #23, opened 2017-06-11
JAV-70 removed missing dependencies and aggregated coverage report

seanyinx:


integration-tests/pom.xml
```diff
@@ -11,6 +11,193 @@
...
+            <artifactId>spring-boot-starter-provider</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.servicecomb</groupId>
+            <artifactId>spring-boot-starter-transport</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.servicecomb.demo</groupId>
+            <artifactId>pojo-client</artifactId>
+            <version>0.1.0-m2-SNAPSHOT</version>
```
TankTian: why , need version?

TankTian: why , need version?

seanyinx: because demo dependencies are not managed anywhere else

PR #26, opened 2017-06-13
Demo run profile

wujimin:
JAV-92 create a pom profile to make demos can run by java -jar xxx.jar

mvn install -Pdemo-run-release
  will create demo-bin directory in the root of java-chassis


.gitignore
```diff
@@ -75,3 +75,4 @@ __pycache__/
 
 # SVN metadata
 .svn/
+demo-bin
```
WillemJiang: Can we  put the demo-bin into the target directory which is ignored by the git.

demo/pom.xml
```diff
@@ -165,4 +165,74 @@
...
+	
+	<profiles>
+		<profile>
+			<id>demo-run-release</id>
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.commonjava.maven.plugins</groupId>
+						<artifactId>directory-maven-plugin</artifactId>
+						<version>0.1</version>
```
seanyinx: the version numbers of the 3 plugins can be inherited from parent

.gitignore
wujimin: changed to root/demo/target

demo/pom.xml
wujimin: removed these versions

PR #38, opened 2017-06-18
JAV-104 clean up unused methods whose tests fail on windows

seanyinx:


foundation-common/src/main/java/io/servicecomb/foundation/common/utils/FortifyUtils.java
```diff
@@ -16,43 +16,14 @@
 
 package io.servicecomb.foundation.common.utils;
 
-import java.io.File;
```
liubao68: we'd better to remove this class in the future, some of methods in this class, like xml processing can also be deleted.

seanyinx: planning to do so

PR #41, opened 2017-06-22
[JAV-127]retry的场景在自己的线程池执行

liubao68:


handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -53,6 +57,15 @@
 public class LoadbalanceHandler extends AbstractHandler {
     private static final Logger LOGGER = LoggerFactory.getLogger(LoadbalanceHandler.class);
 
+    private static final ExecutorService RETRY_POOL = Executors.newCachedThreadPool(new ThreadFactory() {
```
WillemJiang: It could be better to use fix size thread pool to avoid the resource leak.

handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -181,7 +194,9 @@ private void sendWithRetry(Invocation invocation, AsyncResponse asyncResp,
             newExecutor = new Executor() {
                 @Override
                 public void execute(Runnable command) {
-                    command.run(); // retry的场景，对于同步调用， 需要在网络线程中进行。同步调用的主线程已经被挂起，无法再主线程中进行重试。
+                    // retry的场景，对于同步调用, 同步调用的主线程已经被挂起，无法再主线程中进行重试;
```
seanyinx: can you reproduce this issue in a test and ensure it's fixed in the new solution?

liubao68: It's quite easy to reproduce according to the issue discretion, and I have test it.

liubao68: start a producer to listen to both roc/highway, and a rest api that throws an exception. start a consumer with load balance retry enabled, and call this method

handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
liubao68: This Is not a resource leak stuff, but a performance stuff I think. And I.think it's better to use a cached pool to reuse in high volume request and when no requests thread will be destroyed  automatically

handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
seanyinx: great, could you please add an automated test to ensure it won't be broken by changes in the future?

liubao68: added

demo/demo-springmvc/springmvc-server/src/main/java/io/servicecomb/demo/springmvc/server/ControllerImpl.java
```diff
@@ -52,6 +52,9 @@ public String saySomething(String prefix, @RequestBody Person user) {
     @RequestMapping(path = "/sayhi", method = RequestMethod.GET)
     public String sayHi(HttpServletRequest request) {
         String[] values = request.getParameterValues("name");
+        if (values != null && values.length > 0 && values[0].equals("throwexception")) {
```
WillemJiang: 为什么不能直接从框架内部扔出异常。

PR #42, opened 2017-06-23
add samples of java chassis

zhenghaibo1008:
add pojo、jaxrs、springmvc、codefirst、customer-handler sample to java chassis


samples/jaxrs-sample/src/main/java/io/servicecomb/samples/jaxrs/client/JaxrsClient.java
```diff
@@ -0,0 +1,50 @@
...
+
+@Component
+public class JaxrsClient {
+
+    @RpcReference(microserviceName = "jaxrs", schemaId = "jaxrsHello")
+    private static Hello hello;
+
+    public static void main(String[] args) throws Exception {
+        init(args);
+        System.out.println(hello.sayHi("Java Chassis"));
```
liubao68: we can't call @RestScheam service with RPC call, must call these method with RestTeplate( REST style api). This can work only because you define only POST methods in your application.

samples/jaxrs-sample/src/main/java/io/servicecomb/samples/jaxrs/server/JaxrsHelloImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.ws.rs.core.MediaType;
+import java.awt.*;
+
+@RpcSchema(schemaId = "jaxrsHello")
+@Path("/jaxrshello")
+@Produces(MediaType.APPLICATION_JSON)
+public class JaxrsHelloImpl implements Hello {
+
+    @Path("/sayhi")
+    @POST
```
liubao68: There is only POST methods, try some get/put or other to find why client code is wrong

samples/jaxrs-sample/src/main/resources/META-INF.spring/pojo.client.bean.xml
```diff
@@ -0,0 +1,30 @@
...
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://www.springframework.org/schema/p"
+	xmlns:util="http://www.springframework.org/schema/util" xmlns:cse="http://www.huawei.com/schema/paas/cse/rpc"
+	xmlns:context="http://www.springframework.org/schema/context"
+	xsi:schemaLocation="
+		http://www.springframework.org/schema/beans classpath:org/springframework/beans/factory/xml/spring-beans-3.0.xsd
+		http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.0.xsd
+		http://www.huawei.com/schema/paas/cse/rpc classpath:META-INF/spring/spring-paas-cse-rpc.xsd">
+
+	<cse:rpc-reference id="hello" microservice-name="hello"
```
liubao68: microservice-name should be jaxrs

samples/springmvc-sample/src/main/java/io/servicecomb/samples/springmvc/client/SpringmvcClient.java
```diff
@@ -0,0 +1,50 @@
...
+import io.servicecomb.foundation.common.utils.Log4jUtils;
+import io.servicecomb.provider.pojo.RpcReference;
+import io.servicecomb.samples.springmvc.Hello;
+import io.servicecomb.samples.springmvc.models.Person;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SpringmvcClient {
+
+    @RpcReference(microserviceName = "springmvc", schemaId = "springmvcHello")
```
liubao68: same problem as jaxrs examples

samples/README.md
```diff
@@ -0,0 +1,24 @@
+# samples
```
liubao68: notes about the examples:
1. we can call jaxrs/springmvc server implementation with RPC style client code, this is not our intention, but this seams a quite great feature. We can discuss if include this feature in the example. Before that, we must test any other HTTP methods like GET/POST
2. should add client codes use RestTemplate to call jarxs/springmvc service, that's what we have given in other demos.

PR #45, opened 2017-06-24
changed service port to 9980

eric-lee-ltk:
cse service port among all demos except demo-spring-boot-discovery is 9980, is it a good idea to keep it the same among all demos?


WillemJiang:
We just change the service center docker image which use the 30100 port, so this change is not we want.  But you PR just showed a defect of our demo test which doesn't pick up the error of service starting. We should fix this defect ASAP.

PR #48, opened 2017-06-24
JAV-133 RegistryUtils: placeholder in publish address is network interface name, not host name

wujimin:


service-registry/src/test/java/io/servicecomb/serviceregistry/TestRegistry.java
```diff
@@ -396,27 +402,128 @@ public void onMicroserviceInstanceChanged(MicroserviceInstanceChangedEvent chang
...
         String address = RegistryUtils.getPublishAddress();
-        Assert.assertNotNull(address);
-        System.setProperty("cse.service.publishAddress", "{eth0}");
-        address = RegistryUtils.getPublishAddress();
-        Assert.assertNotNull(address);
+        Assert.assertEquals("1.1.1.1", address);
+
+        new Expectations(DynamicPropertyFactory.getInstance()) {
+            {
+                DynamicPropertyFactory.getInstance().getStringProperty("cse.service.publishAddress", "");
```
WillemJiang: It could be better if we use constant value for the "cse.service.publishAddress".

wujimin: done

PR #50, opened 2017-06-26
[JAV-136]在发生异常，执行fallback的时候，没有打印原始异常，问题定位非常不方便

bismy:


handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/BizkeeperHandler.java
```diff
@@ -52,6 +54,16 @@
...
             LOG.warn("Hystrix properties already registerd. Dynamic configuration may not work.");
         }
+        try {
+            HystrixPlugins.getInstance().registerCommandExecutionHook(new HystrixCommandExecutionHook() {
+                public <T> Exception onExecutionError(HystrixInvokable<T> commandInstance, Exception e) {
+                    LOG.warn("bizkeeper execution error", e);
+                    return e; //by default, just pass through
+                }
+            });
+        } catch (Exception e) {
```
seanyinx: only catch exception type thrown by the method called

liubao68: no exception defined by this method, only RuntimeException will throw by this method. Here we ignore any runtimeexception thrown by this method.

seanyinx: it throws IllegalStateException and we shall only catch this one.

liubao68: I don't think it's a good practice. Because IllegalStateException is thrown internally, we don't know what's it if we don't check the source code. This rule is acceptable when the exception is declared by the method implicitly and not a RuntimeException. When we deal with RuntimeException, we'd better to use generic exception type, It's better for code quality.

WillemJiang: As you don't log the actual exception message, it could mislead the user if the RuntimeException is not related to HystrixCommandExcecution hook setting.

PR #56, opened 2017-06-27
JAV-102 find resource root path

wujimin:
if a microservice.yaml path is : file:/....../target/classes/microservice.yaml
then root path is : ....../target/classes
if a class (io.servicecomb.demo.jaxrs.server.ComputeImpl) is in file:/....../target/classes/io/servicecomb/demo/jaxrs/server/
then root path is : ....../target/classes
when resource is in a jar, then the root is the jar path.
file:/....../repository/commons-lang/commons-lang/2.6/commons-lang-2.6.jar!/org/apache/commons/lang/
root path is : ....../repository/commons-lang/commons-lang/2.6/commons-lang-2.6.jar

### DIFF
--- a/foundation-config/src/main/java/io/servicecomb/config/archaius/sources/AbstractConfigLoader.java
+++ b/foundation-config/src/main/java/io/servicecomb/config/archaius/sources/AbstractConfigLoader.java
@@ -22,7 +22,10 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
+
 import org.springframework.util.ResourceUtils;
+
+import io.servicecomb.foundation.common.utils.FileUtils;
 
 public abstract class AbstractConfigLoader {
     protected String orderKey = "config-order";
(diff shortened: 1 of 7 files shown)


foundation-config/src/main/java/io/servicecomb/config/archaius/sources/AbstractConfigLoader.java
```diff
@@ -54,9 +57,9 @@ public ConfigModel load(URL url, String resourceName) throws IOException {
 
         ConfigModel configModel = new ConfigModel();
         configModel.setUrl(url);
-        //        if (resourceName != null) {
-        //            configModel.setRootPath(FileUtils.findRootPath(url, resourceName));
-        //        }
+        if (resourceName != null) {
```
WillemJiang: If the resourceName is null ,  it's better to throw the exception here.

wujimin: resourceName is null when load additional config.
it's normal, not a exception

PR #57, opened 2017-06-27
[JAV-147]RestTemplate支持扩展

liubao68:


provider-springmvc/src/main/java/io/servicecomb/provider/springmvc/reference/RestTemplateWrapper.java
```diff
@@ -27,26 +29,43 @@
...
 import org.springframework.web.client.RestTemplate;
 
-import io.servicecomb.common.rest.RestConst;
-
 /**
  * 用于同时支持cse调用和非cse调用
  */
 public class RestTemplateWrapper extends RestTemplate {
-    private static RestTemplate cseRestTemplate = new CseRestTemplate();
+    private static List<AcceptableRestTemplate> acceptableRestTemplates = new ArrayList<>();
```
seanyinx: a better design is to declare default rest template and avoid null checking all together
```
class RestTemplateWrapper implements RestOperations {
  private static final RestTemplate defaultTemplate =  new RestTemplate();
  private static List<AcceptableRestTemplate> acceptableRestTemplates = asList(new CseRestTemplate());

    private RestTemplate getRestTemplate(String url) {
        for (AcceptableRestTemplate template : acceptableRestTemplates) {
            if (template.isAcceptable(url)) {
                return template;
            }
        }
        return defaultTemplate;
    }
}
```

seanyinx: and i don't understand why these methods need to be static. we have way too many static classes/methods. this kind of programming model creates tons of tight coupling.

provider-springmvc/src/test/java/io/servicecomb/provider/springmvc/reference/TestRestTemplateBuilder.java
```diff
@@ -16,14 +16,49 @@
...
+        }
+    }
+    
+    @Test
+    public void testRestTemplateBuilderResttemplate() {
+        RestTemplateBuilder.addAcceptableRestTemplate(new MyAcceptableRestTemplate());
+        RestTemplate template = RestTemplateBuilder.create();
+        try {
+            template.delete("http://test");
+            Assert.assertFalse(true);
```
seanyinx: this test only covers the exception case. what about the happy path?

provider-springmvc/src/test/java/io/servicecomb/provider/springmvc/reference/TestRestTemplateBuilder.java
```diff
@@ -57,13 +57,23 @@ public void testRestTemplateBuilderResttemplate() {
...
             Assert.assertFalse(true);
         } catch (RestClientException e) {
+            //custom
             Assert.assertEquals(e.getMessage(), "test error.");
         }
 
         try {
+            template.delete("cse://test/a/b/c");
+            Assert.assertFalse(true);
+        } catch (NullPointerException e) {
```
WillemJiang: It could be better if we can wrap the NPE with some kind of Runtime exception.

PR #58, opened 2017-06-27
JAV-105 create microservice from definition

wujimin:


service-registry/src/main/java/io/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -61,6 +60,9 @@
 public final class RegistryUtils {
     private static final Logger LOGGER = LoggerFactory.getLogger(RegistryUtils.class);
 
+    // temp field
+    private static MicroserviceManager microserviceManager;
```
WillemJiang: It's not thread safe if there are multiple thread init the microserviceManager.

wujimin:
create microservice from microservice definition, and save them in MicroserviceManager

service-registry/src/test/java/io/servicecomb/serviceregistry/config/TestPropertiesLoader.java
```diff
@@ -16,31 +16,31 @@
...
+import io.servicecomb.config.archaius.sources.ConfigModel;
+import io.servicecomb.config.archaius.sources.MicroserviceConfigLoader;
+import io.servicecomb.serviceregistry.api.registry.Microservice;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceManager;
+import io.servicecomb.serviceregistry.definition.DefinitionConst;
+import io.servicecomb.serviceregistry.definition.MicroserviceDefinition;
 
 public class TestPropertiesLoader {
+    private static MicroserviceManager microserviceManager = new MicroserviceManager();;
```
WillemJiang: ;;

service-registry/src/main/java/io/servicecomb/serviceregistry/RegistryUtils.java
wujimin: only be inited when cse start, and only init once.

wujimin: and all about this field code is temp
all will be changed after the follow PR.

PR #59, opened 2017-06-27
Multiple microservices demo

wujimin:


wujimin:
1.allow multiple microservices in one process
2.unit test speed optimized

core/src/main/java/io/servicecomb/core/Const.java
```diff
@@ -34,5 +34,5 @@ private Const() {
 
     public static final String SRC_MICROSERVICE = "x-cse-src-microservice";
 
-    public static final String DEST_MICROSERVICE = "x-cse-dest-microservice";
+    public static final String TARGET_MICROSERVICE = "x-cse-target-microservice";
```
WillemJiang: Is this header used by other outside service?

wujimin: when target process have multiple microserives,  and consumer do not use cse sdk
then consumer must set x-cse-target-microservice in header

WillemJiang: If the header is used internally, it's OK to change the header name, but if it is used from outside, we should keep the old header there.

wujimin: no problem
this is the first time to use it.

PR #68, opened 2017-07-04
[JAV-161]负载均衡模块每次请求都会加载filter, 修改为loadbalance创建的时候创建一次

bismy:


handlers/handler-loadbalance/src/test/java/io/servicecomb/loadbalance/filter/TestIsolationServerListFilter.java
```diff
@@ -19,24 +19,46 @@
...
+
 public class TestIsolationServerListFilter {
 
     IsolationServerListFilter IsolationServerListFilter = null;
 
     LoadBalancerStats loadBalancerStats = null;
 
+    @BeforeClass
+    public static void initConfig() throws Exception {
+        ConfigUtil.installDynamicConfig();
```
seanyinx: it's better to clean up this config after use, or other tests using archaius may fail

PR #74, opened 2017-07-06
JAV-88 added springmvc tests to verify springmvc endpoints only

eric-lee-ltk:
added tests to verify springmvc endpoints only

### DIFF
--- a/integration-tests/springmvc-tests/src/test/java/io/servicecomb/demo/springmvc/tests/SpringMvcIntegrationTest.java
+++ b/integration-tests/springmvc-tests/src/test/java/io/servicecomb/demo/springmvc/tests/SpringMvcIntegrationTest.java
@@ -309,12 +309,11 @@ public class SpringMvcIntegrationTest {
 
   @Test
   public void ensureServerBlowsUp() {
-
     try {
       restTemplate.getForObject(
           controllerUrl + "sayhi?name=throwexception",
           String.class);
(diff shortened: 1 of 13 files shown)


integration-tests/springmvc-tests/src/test/java/io/servicecomb/demo/springmvc/tests/SpringMvcIntegrationTest.java
```diff
@@ -16,19 +16,49 @@
...
 import org.springframework.http.ResponseEntity;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
 import org.springframework.web.client.RestTemplate;
+import org.springframework.web.client.UnknownHttpStatusCodeException;
 
 public class SpringMvcIntegrationTest {
 
+  private final String baseUrl = "http://127.0.0.1:8080/";
   private final RestTemplate restTemplate = new RestTemplate();
```
liubao68: All requests go direct to server, and not pass handler chain. Should we use RestTemplateBuilder to create invocation?

seanyinx: you mean consumer handler chain right? we are trying to test provider (including its handler chain) in isolation.

PR #77, opened 2017-07-11
removed redundant validation logic

eric-lee-ltk:
method `checkMicroserviceName` has already validate whether name is empty or not, there is no need to validate it again, right?

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/definition/MicroserviceDefinition.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/definition/MicroserviceDefinition.java
@@ -95,10 +95,8 @@ public class MicroserviceDefinition {
             Configuration conf = ConfigUtil.createConfig(Arrays.asList(model));
             String name =
                 conf.getString(DefinitionConst.qulifiedServiceNameKey, DefinitionConst.defaultMicroserviceName);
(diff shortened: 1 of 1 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/definition/MicroserviceDefinition.java
```diff
@@ -95,10 +95,8 @@ private void initCombinedFrom(List<ConfigModel> configModels) {
             Configuration conf = ConfigUtil.createConfig(Arrays.asList(model));
             String name =
                 conf.getString(DefinitionConst.qulifiedServiceNameKey, DefinitionConst.defaultMicroserviceName);
-            if (!StringUtils.isEmpty(name)) {
-                checkMicroserviceName(name);
-                combinedFrom.add(name);
```
seanyinx: do we have corresponding test to verify that this is indeed redundant?

eric-lee-ltk: seems like there has no unit tests for this class and no test covers this scene.

seanyinx: combinedFrom will throw NPE when name is null and the logic is different after change when name is empty

PR #85, opened 2017-07-14
JAV-88 added pojo tests

eric-lee-ltk:
added pojo tests to verify highway endpoints only


seanyinx:
please do a git rebase

PR #86, opened 2017-07-14
Jav 163 zuul tracing

seanyinx:


integration-tests/spring-zuul-tracing-tests/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <version>0.1.1-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>spring-zuul-tracing-tests</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.servicecomb.tests</groupId>
+      <artifactId>test-common</artifactId>
```
WillemJiang: If the test common is used by other test, it could be better we just put the code into main/src. 
Because we normally don't release the test jar.

integration-tests/spring-zuul-tracing-tests/src/test/java/io/servicecomb/spring/cloud/zuul/tracing/LoggingFilterConfig.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+        throws IOException, ServletException {
+      logger.info("logged tracing filter");
```
WillemJiang: It could be better if we can log more information about request and response here.

seanyinx: this is just for testing to grab the tracing logs

integration-tests/spring-zuul-tracing-tests/pom.xml
seanyinx: users don't need to use test common, right? it's to share common test scaffolding in our integration tests

WillemJiang: OK, that's fine.

PR #90, opened 2017-07-18
JAV-214 avoid NullPointerException in register task, let it try to register again

ShakespeareLiu:
avoid NullPointerException in register task.

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
@@ -56,7 +56,9 @@ public class MicroserviceRegisterTask extends AbstractRegisterTask {
                     microservice.getServiceName(),
                     microservice.getVersion());
 
-            checkSchemaIdSet();
+            if(!checkSchemaIdSet()) {
+                return false;
+            }
         } else {
             serviceId = srClient.registerMicroservice(microservice);
             if (StringUtils.isEmpty(serviceId)) {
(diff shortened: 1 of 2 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -89,6 +89,10 @@ protected boolean doRegister() {
 
     private void checkSchemaIdSet() {
         Microservice existMicroservice = srClient.getMicroservice(microservice.getServiceId());
+        if (existMicroservice == null) {
```
wujimin: only happened when the microservice been deleted after "srClient.getMicroserviceId".
if this happened, it's wrong to just log and return, because we need to reregister the microservice or just refused boot.

ShakespeareLiu: yes, maybe is not the properly way to just return. I found this problem when the connection to Service Center is not stable.  sdk is failed to update heartbeat to SC or any time when sdk register to SC. but if the register thread exits, it will not cover any more even next time sdk could update heartbeat successfully ,it's an improvement issue,

wujimin: i feel that, it should be treat as register microservice failed, and reset the service center task like register instance failed.

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -103,7 +109,7 @@ private void checkSchemaIdSet() {
                     microservice.getVersion(),
                     localSchemas,
                     existSchemas);
-            return;
+            return true;
```
seanyinx: shall we return true or false in case the schema mismatches?

ShakespeareLiu: this flag  just  judges whether get the microservice schema dates successfully,  another flag "schemaIdSetMatch" judges whether the schema match.

PR #95, opened 2017-07-21
JAV-211 APPLICTION_ID & microservice name can not use any java packag…

wujimin:
1.add prefix _ to numbers
2.add prefix _ to  java keyword
3.change - to _

### DIFF
--- a/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
+++ b/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
@@ -18,9 +18,12 @@ package io.servicecomb.swagger.generator.core.utils;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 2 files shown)


swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
```diff
@@ -42,6 +45,61 @@
 import io.swagger.models.properties.Property;
 
 public final class ClassUtils {
+    // copy from https://docs.oracle.com/javase/tutorial/java/nutsandbolts/_keywords.html
+    private static final Set<String> JAVA_KEYWORDS = new HashSet<>();
+    static {
+        JAVA_KEYWORDS.addAll(Arrays.asList("abstract",
+                "continue",
```
seanyinx: is it valid to use true, false, null in package name?

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
```diff
@@ -217,6 +275,20 @@ public static String correctMethodParameterName(String name) {
     }
 
     public static String correctClassName(String name) {
-        return name.replace("-", "_");
+        String parts[] = name.split("\\.");
+        for (int idx = 0; idx < parts.length; idx++) {
+            String part = parts[idx];
+            if (part.isEmpty()) {
+                parts[idx] = "_";
```
seanyinx: is this case covered in test?

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
wujimin: oh, it's invalid, let me fix it. and maybe lost other scenes also......

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
wujimin: at first, i don't think anyone will use a name like "1..."
  and add "part.isEmpty()" is because "part.charAt(0)" in the next block.

but, ok, i will add test case for this.

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
wujimin: add true/false/null, and change JAVA_KEYWORDS to JAVA_RESERVED_WORDS

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/utils/ClassUtils.java
wujimin: change string.split(regex) to string.split(regex, -1) to remains empty part
and add test case

PR #96, opened 2017-07-24
Configuration of config center url

wujimin:
1.read config center url from local config data, that include microservice.yaml/additional config/environment/command parameters
　2.discuss with config-cc, modify SPI interface, not use WatchedConfigurationSource directly, but extend it, add method: void init(Configuration localConfig), to avoid create another configration instance and reduce api of create configration coupling

### DIFF
--- a/foundations/foundation-config/src/main/java/io/servicecomb/config/ConfigUtil.java
+++ b/foundations/foundation-config/src/main/java/io/servicecomb/config/ConfigUtil.java
@@ -16,7 +16,12 @@
 
 package io.servicecomb.config;
 
+import static io.servicecomb.foundation.common.base.ServiceCombConstants.CONFIG_CSE_PREFIX;
+import static io.servicecomb.foundation.common.base.ServiceCombConstants.CONFIG_SERVICECOMB_PREFIX;
+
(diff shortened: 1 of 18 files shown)


foundations/foundation-config/src/main/java/io/servicecomb/config/ConfigUtil.java
```diff
@@ -49,7 +45,7 @@
 public final class ConfigUtil {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigUtil.class);
 
-    private static final String configCenterUrlKey = "cse.config.client.serverUri";
+    protected static final String configCenterUrlKey = "cse.config.client.serverUri";
```
liubao68: 对于final class，不应该有protected方法

wujimin: share for unit test.

WillemJiang:
Hi jiming,  there are some changes on the ConfigUtil. Could you update the code and resolve the conflicts?

wujimin:
resolved conflicts
and fixed a JAV-129 bug

PR #97, opened 2017-07-24
JAV-221 make vertx eventloop thread name related to vertx name

wujimin:
previous name is "vert.x-eventloop-thread-{number}"
new name is "{name}-vert.x-eventloop-thread-{number}"


wujimin:
yes
but vertx did not provide mechanism to do this
and we have automatic test case to ensure this is no problem.

liubao68:
return Vertx.vertx(vertxOptions) is a factory method , after change, we new a object our self, most of time UT can not guarantee this

wujimin:
UT do not  care for this , thread  name only  for  runtime.

wujimin:
just  only  one  test case check  the  thread name , this  case  added  in this PR

PR #100, opened 2017-07-25
JAV-222 when register schema failed, can not find the reason from log.

wujimin:


service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -122,6 +123,36 @@ private boolean retry(RequestContext requestContext, Handler<RestResponse> respo
...
+
+    // temporary copy from syncHandler
+    // we will use swagger invocation to replace RestUtils later.
+    private Handler<RestResponse> syncHandlerEx(CountDownLatch countDownLatch, Holder<ResponseWrapper> holder) {
+        return restResponse -> {
+            RequestContext requestContext = restResponse.getRequestContext();
+            HttpClientResponse response = restResponse.getResponse();
+            if (response == null) {
+                // invoke failed, call another SC instance
+                if (!retry(requestContext, syncHandlerEx(countDownLatch, holder))) {
```
WillemJiang: How about the service center doesn't send the response back. 
Do we have timeout mechanism for CountDownLatch?
BTW, is it enough if we just retry once?

wujimin: httpClientRequest.setTimeout to control timeout.
if no response, then timeout and callback with new RestResponse(requestContext, null),that means response is null
retry just copy from old code, infact maybe no need to retry, because registerSchema failed means registerMicroservice failed, and will auto retry.

WillemJiang: If we don't need the retry here,  we can consider to remove this part of code.

PR #101, opened 2017-07-25
[JAV-235]解决query/path参数的转码问题；以及body的编码问题

liubao68:


common/common-rest/src/main/java/io/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
```diff
@@ -58,7 +58,8 @@ public Object getValue(RestServerRequest request) throws Exception {
 
                 String contentType = request.getContentType();
                 if (contentType != null && contentType.startsWith(MediaType.TEXT_PLAIN)) {
```
WillemJiang: You can get the encoding information from the content-type , just like this "Content-Type: text/html; charset=utf-8" . 
I don't think using the "UTF-8"  without checking the charset is good idea. We can use the default setting as UTF-8, but when user specify the charset, we need to take it in the consideration.

liubao68: I add a TODO to fix this later. But we use a hard code now. Now UTF-8 is the correct one, because most URI encoding use UTF-8 and we use json as body, they are all UTF-8

liubao68: https://servicecomb.atlassian.net/browse/JAV-245

PR #103, opened 2017-07-28
[JAV-246]sc code reliable issues, check issue for details

liubao68:


service-registry/src/main/java/io/servicecomb/serviceregistry/config/ServiceRegistryConfig.java
```diff
@@ -126,9 +128,9 @@ public int getConnectionTimeout() {
     public int getIdleConnectionTimeout() {
         DynamicIntProperty property =
             DynamicPropertyFactory.getInstance()
-                    .getIntProperty("cse.service.registry.client.timeout.idle", DEFAULT_TIMEOUT_IN_MS);
+                    .getIntProperty("cse.service.registry.client.timeout.idle", DEFAULT_TIMEOUT_IN_SECONDS);
```
wujimin: changed from millisecond to second
make sure that nobody used this config.

liubao68: this config is not documented yet, we will add it to wiki when neccesary(now no need to give this. because 30 seconds is the best and not neccessary to change it .

PR #106, opened 2017-08-04
[JAV-264]support zone aware load balance features

liubao68:


handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/filter/ZoneAwareServerListFilterExt.java
```diff
@@ -0,0 +1,72 @@
...
+import io.servicecomb.serviceregistry.RegistryUtils;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ZoneAwareServerListFilterExt implements ServerListFilterExt {
+
+  @Override
+  public List<Server> getFilteredListOfServers(List<Server> list) {
+    List<Server> result = new ArrayList<>();
```
wujimin: is it possible to create cache for this?
avoid loop and build the list each time?

liubao68: not possible here because doing that we have to iterate 'list' to check if it is changed here.

wujimin: if want to use cache, must:
1.the filter must be the first filter
2.when create instance, save the cache reference
3.when getFilteredListOfServers, ask cache if it is changed, if changed, then rebuild zone cache

it's a little complex, if this filter caused a performance problem in the furture, then can use this method to optimize.

PR #113, opened 2017-08-10
[JAV-286]add protect feature when call remote calls in bean initializ…

liubao68:
…ation

### DIFF
--- a/core/src/main/java/io/servicecomb/core/SystemBootListener.java
+++ b/core/src/main/java/io/servicecomb/core/SystemBootListener.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 Huawei Technologies Co., Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicecomb.core;
+
+import org.springframework.stereotype.Component;
+
+@Component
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/servicecomb/core/SystemBootListener.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package io.servicecomb.core;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class SystemBootListener implements BootListener {
+
+  private static boolean ready = false;
```
WillemJiang: The default value of boolean is false, so you don't need assign the value.

core/src/test/java/io/servicecomb/core/TestSystemBootListener.java
```diff
@@ -0,0 +1,36 @@
...
+import io.servicecomb.serviceregistry.RegistryUtils;
+import io.servicecomb.serviceregistry.ServiceRegistry;
+import io.servicecomb.serviceregistry.registry.ServiceRegistryFactory;
+import mockit.Deencapsulation;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
+ * Created by l00168639 on 2017/8/10.
```
WillemJiang: Please update the template for it.

demo/demo-pojo/pojo-client/src/main/java/io/servicecomb/demo/pojo/client/BeanRpcTest.java
```diff
@@ -0,0 +1,39 @@
+package io.servicecomb.demo.pojo.client;
+
+import io.servicecomb.demo.server.Test;
+import io.servicecomb.provider.pojo.RpcReference;
+import org.springframework.stereotype.Component;
+
+/**
+ * Created by l00168639 on 2017/8/10.
```
WillemJiang: same issue.

core/src/main/java/io/servicecomb/core/SystemBootListener.java
liubao68: I see many guidelines saying better to give a default value implicitly. e.g. some eclipse compiler and coverity tools

PR #114, opened 2017-08-12
Unified servlet init

wujimin:


demo/demo-server-servlet/src/main/webapp/WEB-INF/web.xml
```diff
@@ -27,17 +27,30 @@
 		</param-value>
 	</context-param>
 
+	<!-- usage 1 equals usage 2 -->
```
WillemJiang: If these two cases are same, we don't need to keep the case 2 here.

spring-boot-starter/spring-boot-starter-transport/src/main/java/io/servicecomb/springboot/starter/transport/RestServletInitializer.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void onStartup(ServletContext servletContext) throws ServletException {
+        if (getPort() == 0) {
+            LOGGER.warn(
+                    "spring boot embed web container listen port is 0, serviceComb will not use container's port to handler RESTful request.");
+            return;
+        }
+
+        // web container did not did listen now.
+        // so mock to listen, and then close.
```
WillemJiang: I guess you just want to test if the port is open for use.

wujimin: if dynamic inject servlet, depend on if web container listened
but for embed web container, did not listen in this time,  for unified the inject logic, we do listen and close in this time.

demo/demo-server-servlet/src/main/webapp/WEB-INF/web.xml
wujimin: document had not finished now, so remain the information here
after document finished, move the description to the document

spring-boot-starter/spring-boot-starter-transport/src/main/java/io/servicecomb/springboot/starter/transport/RestServletInitializer.java
WillemJiang: Can we just put the "RestServletInjector.defaultInject(servletContext);"  there instead of checking the port?
It's much simple, you don't need to check the port here.

wujimin: can not.
if serviceComb work with spring boot embed web container, and serviceComb use another port(that means use vertx rest transport), if this time
if we register our servlet, will cause embed web container failed

PR #117, opened 2017-08-17
JAV-304 No error when provider starts up with unsupported return type…

wujimin:
…s in schema

### DIFF
--- a/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/extend/ModelResolverExt.java
+++ b/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/extend/ModelResolverExt.java
@@ -26,6 +26,7 @@ import org.springframework.util.StringUtils;
 
 import com.fasterxml.jackson.databind.JavaType;
 
+import io.servicecomb.foundation.common.exceptions.ServiceCombException;
 import io.servicecomb.swagger.converter.property.StringPropertyConverter;
 import io.servicecomb.swagger.extend.property.ByteProperty;
 import io.servicecomb.swagger.extend.property.ShortProperty;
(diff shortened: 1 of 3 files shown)


swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/extend/ModelResolverExt.java
```diff
@@ -94,11 +100,11 @@ private void checkType(JavaType type) {
         }
 
         if (type.getRawClass().isInterface()) {
-            throw new Error(type.getTypeName() + " is interface.");
+            throw new ServiceCombException(type.getTypeName() + " is interface. " + msg);
         }
 
         if (Modifier.isAbstract(type.getRawClass().getModifiers())) {
-            throw new Error(type.getTypeName() + " is abstract.");
+            throw new ServiceCombException(type.getTypeName() + " is abstract. " + msg);
```
WillemJiang: Can we change it to "is abstract class"?

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/extend/ModelResolverExt.java
```diff
@@ -81,6 +82,11 @@ private void checkType(JavaType type) {
             return;
         }
 
+        String msg = "must be a clear type.";
```
wujimin: done

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/extend/ModelResolverExt.java
wujimin: done

PR #120, opened 2017-08-18
Assert rpc before sdk init

wujimin:
old implementation had bugs:
1.we set ready in EventType.AFTER_REGISTER
and we suggest developers begin their logic by listen EventType.AFTER_REGISTER too
these two logic's order is undefined

2.invocationFactory is not the invoke entry, we have 3 entries:
a.Invoker
b.CseClientHttpRequest.createRequestMeta
c.InvokerUtils
if developers make remote call in invalid stage, sdk maybe throw exception before invocationFactory


providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/Invoker.java
```diff
@@ -54,8 +55,8 @@ public Invoker(String microserviceName, String schemaId, Class<?> consumerIntf)
         this.consumerIntf = consumerIntf;
     }
 
-    public void prepare() {
-        referenceConfig = CseContext.getInstance().getConsumerProviderManager().getReferenceConfig(microserviceName);
+    protected void prepare() {
```
WillemJiang: It's not binary compatible change,  it's better to keep the method signature there.

wujimin: prepare create in yesterday

WillemJiang: OK

PR #123, opened 2017-08-24
[JAV-166]注册微服务读取paths配置信息

liubao68:


wujimin:
this should based on #122 
basePath in microservice.yaml should support relative path

in #122, urlPrefix saved in system properties before serviceRegistry init

PR #141, opened 2017-09-08
[JAV-364]add more details message when laodbalance not configured in …

liubao68:
…consumer chain. add thread id in console log.

### DIFF
--- a/core/src/main/java/io/servicecomb/core/Invocation.java
+++ b/core/src/main/java/io/servicecomb/core/Invocation.java
@@ -74,6 +74,9 @@ public class Invocation extends SwaggerInvocation {
   }
 
   public Transport getTransport() {
+    if (endpoint == null) {
+      throw new IllegalStateException("Endpoint is empty. Forget to configure \"loadbalance\" in consumer handler chain?");
+    }
     return endpoint.getTransport();
   }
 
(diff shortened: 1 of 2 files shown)


eric-lee-ltk:
Can this warning show up when the application starts instead of runtime?

core/src/main/java/io/servicecomb/core/Invocation.java
```diff
@@ -74,6 +74,9 @@ private void init(OperationMeta operationMeta, Object[] swaggerArguments) {
   }
 
   public Transport getTransport() {
+    if (endpoint == null) {
+      throw new IllegalStateException("Endpoint is empty. Forget to configure \"loadbalance\" in consumer handler chain?");
```
WillemJiang: Can we change the message to "Endpoint is empty. Please  configure \"loadbalance\" in the consumer handler chain.

eric-lee-ltk: I guess this exception should be detected at startup time and should not be thrown at run time. Validate the `microservice.yaml` file before the application starts.

wujimin: we don't know which handler is loadBalance
because cutomer can dev their own loadBalance

liubao68: Yes, we don't know at which handler will a developer to set the address, and they can not use 'loadbalance' and define a address resolver themself. So this message is a best guess for the developer.

PR #147, opened 2017-09-13
Sorted spi services

wujimin:


core/src/test/java/io/servicecomb/core/transport/TestTransportManager.java
```diff
@@ -176,21 +176,21 @@ public void testChooseOneTransportSecond(@Mocked Transport t1, @Mocked Transport
     new Expectations() {
       {
         t1.getOrder();
-        result = 1;
+        result = Integer.MAX_VALUE;
```
liubao68: it's better to add it as a new test case

PR #148, opened 2017-09-14
Servlet rest signature

wujimin:
when server is servlet rest, then support inject client/server filter to transport.


common/common-rest/src/main/java/io/servicecomb/common/rest/filter/HttpClientFilter.java
```diff
@@ -0,0 +1,35 @@
...
+import io.vertx.core.http.HttpClientRequest;
+import io.vertx.core.http.HttpClientResponse;
+
+public interface HttpClientFilter {
+  int getOrder();
+
+  // bodyBytes maybe be null
+  void beforeSendRequest(Invocation invocation, HttpClientRequest clientRequest, Buffer requestBodyBuffer);
+
+  // if finished, then return a none null response
```
liubao68: It's better to give a java doc comment.  spell  non null ? or a response that is not null

PR #161, opened 2017-09-25
Extract rest invocation

wujimin:


common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -0,0 +1,160 @@
...
+  protected void invoke() {
+    try {
+      this.initProduceProcessor();
+      this.setContext();
+      invocation.getHandlerContext().put(RestConst.REST_REQUEST, requestEx);
+
+      for (HttpServerFilter filter : httpServerFilters) {
+        Response response = filter.afterReceiveRequest(invocation, requestEx);
+        if (response != null) {
+          sendResponseQuietly(response);
```
WillemJiang: it looks like the filter could hijack the response message which calling the doInvoke method.
I don't think it's good idea, there are too much assumption.

common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestServer.java
```diff
@@ -1,228 +0,0 @@
...
-import io.servicecomb.foundation.common.utils.JsonUtils;
-import io.servicecomb.foundation.common.utils.SPIServiceUtils;
-import io.servicecomb.foundation.vertx.http.HttpServletRequestEx;
-import io.servicecomb.foundation.vertx.http.HttpServletResponseEx;
-import io.servicecomb.foundation.vertx.stream.BufferOutputStream;
-import io.servicecomb.serviceregistry.RegistryUtils;
-import io.servicecomb.swagger.invocation.Response;
-import io.servicecomb.swagger.invocation.exception.InvocationException;
-
-public class AbstractRestServer {
```
WillemJiang: Is it safe to remove the AbstractRestServer?

common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
wujimin: modify response is one purpose of filter definition：
  producer generator signature and write it to response header.

common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestServer.java
wujimin: this is our inner class, not a interface to developers

common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
WillemJiang: Please add some comments in the HttpServerFilter.afterReceiveRequst,  to make sure user know about this kind of feature.

wujimin: already have comments in HttpServerFilter.afterReceiveRequst and HttpClientFilter.afterReceiveResponse in current PR code.

image

image

WillemJiang: OK.

PR #166, opened 2017-09-28
JAV-418 transport rest client optimize

eric-lee-ltk:
Signed-off-by: eric-lee-ltk <lidagang1@huawei.com>

PR #167, opened 2017-09-28
[JAV-411]add PATCH support

liubao68:


eric-lee-ltk:
This PR will cause conflicts with PR #166 as vertx rest client can use vertx's request method instead of using a specific method, e.g. get, patch.

liubao68:
@eric-lee-ltk ok. but please merge all test cases of this PR to #166

liubao68:
move to #166

PR #168, opened 2017-09-29
[JAV-421]add support for template type for compatibility

liubao68:


swagger/swagger-generator/generator-core/src/test/java/io/servicecomb/swagger/generator/core/TestInvalidType.java
```diff
@@ -43,54 +43,56 @@ public void testAbstractClass() {
         InvalidType.class,
         "testAbstractClass");
   }
+//this is an invalid case. need to discuss in JVA-422
+//  @Test
```
WillemJiang: Please use the Ignore annotation for us to revisit this issue later.

PR #169, opened 2017-09-29
Jav 389 custom tracing

seanyinx:


tracing/tracing-zipkin/src/main/java/io/servicecomb/tracing/zipkin/ZipkinSpanAspect.java
```diff
@@ -0,0 +1,48 @@
...
+import org.aspectj.lang.reflect.MethodSignature;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import brave.Tracing;
+import io.servicecomb.tracing.Span;
+
+@Aspect
+class ZipkinSpanAspect {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
liubao68: naming: public static final parameters should be LOGGER

tracing/pom.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
WillemJiang: Please add the license header.

seanyinx: done

tracing/tracing-zipkin/src/main/java/io/servicecomb/tracing/zipkin/ZipkinSpanAspect.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private final ZipkinTracingAdviser adviser;
+
+  ZipkinSpanAspect(Tracing tracing) {
+    this.adviser = new ZipkinTracingAdviser(tracing.tracer());
+  }
+
+  @Around("execution(@io.servicecomb.tracing.Span * *(..)) && @annotation(spanAnnotation)")
+  public Object advise(ProceedingJoinPoint joinPoint, Span spanAnnotation) throws Throwable {
+    Method method = ((MethodSignature) joinPoint.getSignature()).getMethod();
```
WillemJiang: I think we can add some field in the Span to  let the user override the span name.

seanyinx: yes, maybe wait for user's request, since they ask for more than just span name custimization

PR #180, opened 2017-10-10
fixed typos and removed unnecessary imports

eric-lee-ltk:
Signed-off-by: eric-lee-ltk <lidagang1@huawei.com>

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/definition/DefinitionConst.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/definition/DefinitionConst.java
@@ -25,13 +25,13 @@ public interface DefinitionConst {
 
   String nameKey = "name";
 
-  String qulifiedServiceNameKey = serviceDescriptionKey + "." + nameKey;
+  String qualifiedServiceNameKey = serviceDescriptionKey + "." + nameKey;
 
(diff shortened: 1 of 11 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/definition/DefinitionConst.java
```diff
@@ -25,13 +25,13 @@
 
   String nameKey = "name";
 
-  String qulifiedServiceNameKey = serviceDescriptionKey + "." + nameKey;
```
WillemJiang: It could be better if we use the up case letter to specify the constant string name. Such as QUALIFIED_SERVICE_NAME_KEY can let the user know it's a constant string.

eric-lee-ltk: It seems constants in `DefinitionConst.java` are all in camel case. Besides, there are some duplicate constants in `DefinitionConst.java` and `ServiceCombConstants.java`. Maybe another PR for this as it may affect others and this PR is mainly focus on misspell issues?

WillemJiang: OK， you can fill a JIRA to check this issue.

PR #181, opened 2017-10-10
[JAV-397] Add automatic signing of release on tagged builds

asifdxtreme:


pom.xml
```diff
@@ -85,19 +85,35 @@
       <id>release</id>
       <build>
         <plugins>
+        <plugin>
```
WillemJiang: why do we need to add the config of maven-deploy-plugin?

.travis.settings.xml
```diff
@@ -7,4 +7,14 @@
             <password>${env.SONATYPE_PASSWORD}</password>
         </server>
     </servers>
+<profiles>
```
liubao68: This file is not properly indented

pom.xml
asifdxtreme: yes it is redundant as already while calling mvn deploy we are setting skiptest so this configuration is not required, I will update the changes

.travis.settings.xml
asifdxtreme: Thanks @liubao68 , I have updated the file with proper indentedation

.travis.yml
```diff
@@ -1,17 +1,22 @@
...
-  - mvn deploy -DskipTests --settings .travis.settings.xml
+ deploy:
+  provider: script
+  script: 
+   - openssl aes-256-cbc -K $encrypted_acbbc88fb3ab_key -iv $encrypted_acbbc88fb3ab_iv -in gpg-sec.tar.enc -out gpg-sec.tar -d
+   - tar xvf gpg-sec.tar
+   - mvn deploy -DskipTests -Prelease -Ppassphrase --settings .travis.settings.xml
+  on:
+   tags: true
+   all_branches: true
```
seanyinx: does this mean deploying changes from all branches?

.travis.yml
```diff
@@ -1,17 +1,22 @@
...
 services:
-  - docker
-
+- docker
 script:
-  - mvn clean install -Ddocker.showLogs -Pdocker -Pjacoco -Pit -Pcoverage coveralls:report
-
+- mvn clean install -Ddocker.showLogs -Pdocker -Pjacoco -Pit -Pcoverage coveralls:report
 after_success:
-  - mvn deploy -DskipTests --settings .travis.settings.xml
```
seanyinx: we need snapshots deployed too on each build, in order to allow testing in demo projects to discover potential bugs before release

.travis.yml
asifdxtreme: if there is any tagged commit from any branch then it will deploy with auto-signing

.travis.yml
WillemJiang: Can we just keep deploying the snapshot there?

asifdxtreme: Currently I have made the changes to retain old behaviour and on tagged commit the auto signing will be performed for deploy.

.travis.yml
```diff
@@ -1,17 +1,24 @@
...
+- mvn deploy -DskipTests --settings .travis.settings.xml
+deploy:
+ provider: script
+ script: 
+  - openssl aes-256-cbc -K $encrypted_acbbc88fb3ab_key -iv $encrypted_acbbc88fb3ab_iv -in gpg-sec.tar.enc -out gpg-sec.tar -d
+  - tar xvf gpg-sec.tar
+  - mvn deploy -DskipTests -Prelease -Ppassphrase --settings .travis.settings.xml
+ on:
+  tags: true
+  all_branches: true
```
seanyinx: is it better to limit release to tag on master only?

asifdxtreme: okay, I will update it

PR #187, opened 2017-10-11
Report error instead of warning if rest/highway address format is invalid

MabinGo:
Signed-off-by: MabinGo bin.ma@huawei.com


foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -151,18 +149,18 @@ public static IpPort parseIpPortFromURI(String uriAddress) {
    */
   public static String getRealListenAddress(String schema, String address) {
     if (address == null) {
-      return address;
+      return null;
```
WillemJiang: It could be better to throw the exception out.

PR #188, opened 2017-10-11
JAV-413 Vertx http dispatcher

wujimin:


common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -62,16 +62,12 @@ public void setHttpServerFilters(List<HttpServerFilter> httpServerFilters) {
     this.httpServerFilters = httpServerFilters;
   }
 
-  protected boolean initProduceProcessor() {
```
liubao68: pull request message must include JIRA issue id for reference, e.g JAV-416

wujimin: done

PR #189, opened 2017-10-11
JAV-318 added switch for tracing

eric-lee-ltk:
Signed-off-by: Eric Lee <lidagang1@huawei.com>

### DIFF
--- a/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/base/ServiceCombConstants.java
+++ b/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/base/ServiceCombConstants.java
@@ -36,6 +36,8 @@ public interface ServiceCombConstants {
 
   String CONFIG_TRACING_COLLECTOR_ADDRESS = "servicecomb.tracing.collector.address";
 
+  String CONFIG_TRACING_ENABLED_KEY = "servicecomb.tracing.enabled";
+
   String CONFIG_TRACING_COLLECTOR_PATH = "/api/v1/spans";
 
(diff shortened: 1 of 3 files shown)


integration-tests/spring-zuul-tracing-tests/src/test/java/io/servicecomb/spring/cloud/zuul/tracing/SpringCloudZuulTracingDisableTest.java
```diff
@@ -0,0 +1,51 @@
...
+import io.servicecomb.foundation.common.utils.BeanUtils;
+
+@DirtiesContext
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = TracedZuulMain.class, webEnvironment = RANDOM_PORT)
+public class SpringCloudZuulTracingDisableTest {
+
+  @BeforeClass
+  public static void init() throws Exception {
+    System.setProperty(CONFIG_TRACING_ENABLED_KEY, "false");
```
WillemJiang: Please make sure you clean up the system property once you finish the test.

WillemJiang:
@eric-lee-ltk Please take a look at the failed PR.

eric-lee-ltk:
@WillemJiang The failed log can be resolved by adding the setUpLocalRegistry. However, test classes annotates with SpringBootTest under the same package share the context, which makes the second test class fail. Besides, there are few documents about the usage of SpringBootTest annotation in this case. If we do not use the SpringBootTest annotation like the previous spring-springmvc-tests section, then the autowired TestRestTemplate will become invalid and we have to find another way to get the zuul's running port.

WillemJiang:
Maybe you can setup the test without tracing enable on the other package to workaround this issue.

PR #192, opened 2017-10-13
JAV-402 add model field json ignore test

zhengyangyong:
增加了Model Field @JsonIgnore的test case

### DIFF
--- a/demo/demo-schema/src/main/java/io/servicecomb/demo/CodeFirstRestTemplate.java
+++ b/demo/demo-schema/src/main/java/io/servicecomb/demo/CodeFirstRestTemplate.java
@@ -33,6 +33,7 @@ import io.servicecomb.demo.ignore.InputModelForTestIgnore;
 import io.servicecomb.demo.ignore.OutputModelForTestIgnore;
 import io.servicecomb.demo.server.User;
 import io.servicecomb.serviceregistry.RegistryUtils;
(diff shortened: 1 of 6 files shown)


demo/demo-schema/src/main/java/io/servicecomb/demo/CodeFirstRestTemplate.java
```diff
@@ -57,6 +59,11 @@ public void testCodeFirst(RestTemplate template, String microserviceName, String
       //            testCodeFirstRawJsonString(template, cseUrlPrefix);
       testCodeFirstSayHello(template, cseUrlPrefix);
       testCodeFirstReduce(template, cseUrlPrefix);
+
+      // TODO: highway unsupported until JAV-394 completed
+      if(transport.equals("rest")) {
```
wujimin: if transport is "", that means auto select a transport, that maybe highway

zhengyangyong: Checked，DemoConst.transports = new String[] {"highway", "rest", Const.ANY_TRANSPORT}; 
Const.ANY_TRANSPORT = "";
So，testModelFieldIgnore will be call once , I think that's enough ?

wujimin: ok

PR #195, opened 2017-10-16
JAV-409 Automate the signing of release on tagged commit

asifdxtreme:


scripts/travis.sh
```diff
@@ -0,0 +1,33 @@
+#bin/sh
+
+##Check if the commit is tagged commit or not
+TAGGEDCOMMIT=$(git tag -l --contains HEAD)
```
WillemJiang: It's better to check the tag with commit message.

PR #201, opened 2017-10-18
JAV-453 support version rule

wujimin:


service-registry/src/test/java/io/servicecomb/serviceregistry/version/TestVersionRuleLatestParser.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import io.servicecomb.serviceregistry.definition.DefinitionConst;
+
+public class TestVersionRuleLatestParser {
+  VersionRuleParser parser = new VersionRuleLatestParser();
+
+  VersionRule versionRule = parser.parse(DefinitionConst.VERSION_RULE_LATEST);
```
WillemJiang: How can we tell the latest version? It looks like very version is OK.

service-registry/src/main/java/io/servicecomb/serviceregistry/version/VersionRuleLatestParser.java
```diff
@@ -0,0 +1,41 @@
...
+
+// latest
+public class VersionRuleLatestParser implements VersionRuleParser {
+  class LatestVersionRule extends VersionRule {
+    public LatestVersionRule(String versionRule) {
+      super(versionRule);
+    }
+
+    public boolean isMatch(Version version) {
+      return true;
```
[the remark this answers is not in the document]
wujimin: done.

service-registry/src/test/java/io/servicecomb/serviceregistry/version/TestVersionRuleLatestParser.java
wujimin: firstly, we must have a version set, and then we can determine which one is the latest.
"latest" can not be a simple rule, it must work with data.

i have changed the implementation

PR #207, opened 2017-10-23
JAV-454 enhanced cache manager

wujimin:
1.support multiple versionRules for one microservice
2.can get latest microservice information inside a versionRule
3.can get latest instances inside a versionRule, and can determined if intances changed.
4.work with pull mode.
5.trigger pull when serviceRegistry connection recovery

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/AppManager.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/AppManager.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2017 Huawei Technologies Co., Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.servicecomb.serviceregistry.consumer;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.google.common.eventbus.EventBus;
+
+public class AppManager {
+  private EventBus eventBus;
+
+  private MicroserviceVersionFactory microserviceVersionFactory = new DefaultMicroserviceVersionFactory();
+
+  // key: appId
+  private Map<String, MicroserviceManager> apps = new ConcurrentHashMap<>();
+
+  public AppManager(EventBus eventBus) {
+    this.eventBus = eventBus;
+  }
+
+  public EventBus getEventBus() {
+    return eventBus;
+  }
+
+  public MicroserviceVersionFactory getMicroserviceVersionFactory() {
+    return microserviceVersionFactory;
+  }
+
+  public void setMicroserviceVersionFactory(MicroserviceVersionFactory microserviceVersionFactory) {
+    this.microserviceVersionFactory = microserviceVersionFactory;
+  }
+
+  // microserviceName maybe normal name or alias name
+  public MicroserviceVersionRule getOrCreateMicroserviceVersionRule(String appId, String microserviceName,
+      String versionRule) {
+    MicroserviceManager microserviceManager = apps.computeIfAbsent(appId, id -> {
+      return new MicroserviceManager(this, appId);
+    });
+
+    return microserviceManager.getOrCreateMicroserviceVersionRule(microserviceName, versionRule);
+  }
+}
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceManager.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceManager.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 Huawei Technologies Co., Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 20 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/AppManager.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public void setMicroserviceVersionFactory(MicroserviceVersionFactory microserviceVersionFactory) {
+    this.microserviceVersionFactory = microserviceVersionFactory;
+  }
+
+  // microserviceName maybe normal name or alias name
+  public MicroserviceVersionRule getOrCreateMicroserviceVersionRule(String appId, String microserviceName,
+      String versionRule) {
+    MicroserviceManager microserviceManager = apps.computeIfAbsent(appId, id -> {
```
WillemJiang: id is useless here.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public MicroserviceManager(AppManager appManager, String appId) {
+    this.appManager = appManager;
+    this.appId = appId;
+  }
+
+  public MicroserviceVersionRule getOrCreateMicroserviceVersionRule(String microserviceName,
+      String versionRule) {
+    MicroserviceVersions microserviceVersions =
+        versionsByName.computeIfAbsent(microserviceName, name -> {
```
WillemJiang: same issue there, I suggest to use name variable in the construction method of MicroserviceVersion.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.servicecomb.serviceregistry.consumer;
+
+public interface MicroserviceVersionFactory {
+  MicroserviceVersion create(String microserviceName, String microserviceId);
```
WillemJiang: How can we specify the version number here?

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionRule.java
```diff
@@ -0,0 +1,145 @@
...
+
+    resetLatest();
+    LOGGER.info("delete microserviceVersion, appId={}, microserviceName={}, version={}, versionRule={}.",
+        microserviceVersion.getMicroservice().getAppId(),
+        microserviceVersion.getMicroservice().getServiceName(),
+        microserviceVersion.getVersion().getVersion(),
+        versionRule.getVersionRule());
+  }
+
+  protected void resetLatest() {
```
WillemJiang: The name is better to be resetLatestVersion.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionRule.java
```diff
@@ -0,0 +1,145 @@
...
+  public void setInstances(Collection<MicroserviceInstance> newInstances) {
+    if (newInstances == null) {
+      return;
+    }
+
+    instances = newInstances.stream().filter(instance -> {
+      MicroserviceVersion microserviceVersion = versions.get(instance.getServiceId());
+      return microserviceVersion != null
+          && versionRule.isMatch(microserviceVersion.getVersion(), latestVersion.getVersion());
+    }).filter(instance -> {
```
WillemJiang: Please use forEach instead of filter.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -0,0 +1,178 @@
...
+        DefinitionConst.VERSION_RULE_ALL);
+    if (pulledInstances == null) {
+      return;
+    }
+
+    setInstances(pulledInstances);
+  }
+
+  private void setInstances(List<MicroserviceInstance> pulledInstances) {
+    synchronized (lock) {
```
WillemJiang: Do we have cach or copy on write list for it?
I'm not a big fan of synchronized.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/AppManager.java
wujimin: id will be set to microserviceManager and so on 
that make it easy to log correct information and more clear for debug

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceManager.java
wujimin: same to above

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionFactory.java
wujimin: each appId + microserviceName + version has  a special microserviceId

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
wujimin: can not use that 
as i comment in field , what we protected is not one container, but  is "process pulled instances and create versionRule"

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionRule.java
wujimin: forEach is a "terminal operation"
forEach is not the purpose, Collectors.toMap is the purpose.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersionRule.java
wujimin: done

PR #211, opened 2017-10-24
JAV-464 support annotation ApirRsponse's  description

jeho0815:


WillemJiang:
@jeho0815  The patch is not merged, so please don't close it yourself.

PR #214, opened 2017-10-25
JAV-470 resolve enum param validate fail has no response

jeho0815:


liubao68:
What's the difference with https://github.com/ServiceComb/ServiceComb-Java-Chassis/pull/211? It seems a wrong patch.

swagger/swagger-invocation/invocation-core/src/main/java/io/servicecomb/swagger/invocation/arguments/ArgumentsMapperFactory.java
```diff
@@ -99,10 +99,12 @@ protected boolean isSwaggerWrapBody(ArgumentsMapperConfig config, List<ProviderP
       return false;
     }
 
-    swaggerType = ((Class<?>) swaggerType).getFields()[0].getGenericType();
-    Converter converter = converterMgr.findConverter(type, firstProviderParam, swaggerType);
-    if (ConverterCommon.class.isInstance(converter)) {
-      return false;
+    if (((Class<?>) swaggerType).getFields().length > 0) {
```
wujimin: must have a field in this place
if there is no field, that means it's bug, MUST not think this is a normal scenes.

jeho0815: has remove this modify

jeho0815:
@liubao68 this is resolve exception in method runOnExecutor() has no response

common/common-rest/src/test/java/io/servicecomb/common/rest/TestRestProducerInvocation.java
```diff
@@ -202,6 +202,29 @@ void invoke() {
...
+      @Mock
+      void invoke() {
+        invokeNoParam = true;
+      }
+    }.getMockInstance();
+    initRestProducerInvocation();
+    try {
+      restProducerInvocation.runOnExecutor();
+    } catch (Exception e) {
+      e.printStackTrace();
```
WillemJiang: If the exception is thrown, you need to write assert for the exception and don't just print it out.

liubao68: Assert is done above the expectation, method invoke is not executed and result is false. I think it's his logic.

jeho0815: yes, this is simulate a exception

wujimin: best practice is use org.junit.rules.ExpectedException

jeho0815:
use another pr trace

PR #216, opened 2017-10-26
JAV-365 able to recognize body type of normal http server exception

eric-lee-ltk:
Signed-off-by: Eric Lee <lidagang1@huawei.com>

### DIFF
--- a/handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/BizkeeperCommand.java
+++ b/handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/BizkeeperCommand.java
@@ -16,6 +16,8 @@
 
 package io.servicecomb.bizkeeper;
 
+import static io.servicecomb.swagger.invocation.Response.isValid5xxServerError;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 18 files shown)


wujimin:
developers can define ResponseType of any statusCode in contract.
what's rule of new implementation?

eric-lee-ltk:
@wujimin Currently, java chassis is not fully compatible with spring when the old application uses spring's default error ResponseType, namely String. This makes it a bit complicate when migrating spring application to java chassis.  You can see the description of JAV-365 for details. Besides, casting string directly to Throwable will raise an exeception.

wujimin:
image

maybe this is better(just a demo, not full resolution).
and you should test not only RESTful transport, but also highway

wujimin:
and maybe we can add a rule: unknown statusCode in contract, responseType equals 200
i'm not sure if it's ok

eric-lee-ltk:
@wujimin The biggest problem is the `isFailed` method intercepts all error status code(including the user-defined one), and wraps the error with `InvocationException`. How about we add the following condition:
* if user defines a custom response code and its response type, then it will bypass the `isFailed` condition.

handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/BizkeeperCommand.java
```diff
@@ -79,7 +81,15 @@ protected String getCacheKey() {
           if (isFailedResponse(resp)) {
             // e should implements toString
             LOG.warn("bizkeeper command failed due to:" + resp.getResult());
-            f.onError(resp.getResult());
+            Object result = resp.getResult();
+            // bypass normal http server exception
+            if (isValid5xxServerError(resp.getStatusCode()) && (result instanceof String)) {
```
liubao68: This logic is better put in isFailedResponse implementation

liubao68:
I'd prefer not to merge this PR. And comments are given in JAV-365

eric-lee-ltk:
Yes, the default behavior of error handling is to wrap with InvocationException, changing the rest error handling will make the behavior inconsistent with the highway's default behavior. wujimin's solution seems to be better.

PR #226, opened 2017-10-27
JAV-472 fix rest produce select bug

wujimin:
operation declared to produce text/plain
consumer use accept widecard to send request

but producer produced json format response

### DIFF
--- a/common/common-rest/src/test/java/io/servicecomb/common/rest/definition/TestRestOperationMeta.java
+++ b/common/common-rest/src/test/java/io/servicecomb/common/rest/definition/TestRestOperationMeta.java
@@ -129,6 +129,17 @@ public class TestRestOperationMeta {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


common/common-rest/src/test/java/io/servicecomb/common/rest/definition/TestRestOperationMeta.java
```diff
@@ -129,6 +129,17 @@ public void testCreateProduceProcessorsNotSupported() {
   }
 
   @Test
+  public void testCreateProduceProcessorsTextAndWidecard() {
```
eric-lee-ltk: spelling mistake? Widecard -> Wildcard

wujimin: done

PR #227, opened 2017-10-27
JAV-473 fix usage of HttpServerFilter.afterReceiceRequest result

wujimin:
when HttpServerFilter.afterReceiceRequest return none null, do not call into business operation

### DIFF
--- a/common/common-rest/src/test/java/io/servicecomb/common/rest/TestAbstractRestInvocation.java
+++ b/common/common-rest/src/test/java/io/servicecomb/common/rest/TestAbstractRestInvocation.java
@@ -202,6 +202,7 @@ public class TestAbstractRestInvocation {
     restInvocation = new AbstractRestInvocation() {
       @Override
       protected void doInvoke() throws Throwable {
+        result.value = Response.ok("not run to here");
       }
 
       @Override
--- a/edge/edge-core/src/test/java/io/servicecomb/edge/core/TestEdgeInvocation.java
+++ b/edge/edge-core/src/test/java/io/servicecomb/edge/core/TestEdgeInvocation.java
@@ -22,6 +22,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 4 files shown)


common/common-rest/src/test/java/io/servicecomb/common/rest/TestAbstractRestInvocation.java
```diff
@@ -202,6 +202,7 @@ public void invokeFilterHaveResponse(@Mocked HttpServerFilter filter) {
     restInvocation = new AbstractRestInvocation() {
       @Override
       protected void doInvoke() throws Throwable {
+        result.value = Response.ok("not run to here");
```
eric-lee-ltk: how about change to "This should not run" or "expect not to run here"?

edge/edge-core/src/test/java/io/servicecomb/edge/core/TestEdgeInvocation.java
```diff
@@ -275,6 +279,45 @@ public void invokeNormal(@Mocked MicroserviceVersionMeta latestMicroserviceVersi
...
+      }
+
+      @Override
+      protected void sendResponse(Response response) throws Exception {
+        result.value = response;
+      }
+
+      @Override
+      protected void doInvoke() throws Throwable {
+        result.value = Response.ok("do not run to here");
```
eric-lee-ltk: same as above

common/common-rest/src/test/java/io/servicecomb/common/rest/TestAbstractRestInvocation.java
wujimin: i feet that this is not necessary

edge/edge-core/src/test/java/io/servicecomb/edge/core/TestEdgeInvocation.java
wujimin: same as above

PR #230, opened 2017-10-27
JAV-47 resolve if Invocation has an exception, then has no respose

jeho0815:


jeho0815:
@WillemJiang @liubao68 @wujimin @eric-lee-ltk please review

PR #232, opened 2017-10-28
JAV-469 Edge lb new instance cache manager

wujimin:
edge use new InstanceCacheManager, normal consumer use old InstanceCacheManager
if there is no problem, then old InstanceCacheManager will be deleted.

### DIFF
--- a/edge/edge-core/src/main/resources/microservice.yaml
+++ b/edge/edge-core/src/main/resources/microservice.yaml
@@ -1,0 +1,3 @@
+cse-config-order: -500
+servicecomb:
+  microserviceVersionFactory: io.servicecomb.core.definition.PrivateMicroserviceVersionMetaFactory
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
@@ -33,10 +33,13 @@ import io.servicecomb.serviceregistry.api.registry.Microservice;
 import io.servicecomb.serviceregistry.api.registry.MicroserviceFactory;
 import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
 import io.servicecomb.serviceregistry.cache.InstanceCacheManager;
+import io.servicecomb.serviceregistry.cache.InstanceCacheManagerNew;
 import io.servicecomb.serviceregistry.cache.InstanceCacheManagerOld;
 import io.servicecomb.serviceregistry.client.IpPortManager;
 import io.servicecomb.serviceregistry.client.ServiceRegistryClient;
 import io.servicecomb.serviceregistry.config.ServiceRegistryConfig;
+import io.servicecomb.serviceregistry.consumer.AppManager;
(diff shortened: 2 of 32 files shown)


core/src/main/java/io/servicecomb/core/definition/classloader/DefaultMicroserviceClassLoaderFactory.java
```diff
@@ -20,7 +20,7 @@
   public static final MicroserviceClassLoaderFactory INSTANCE = new DefaultMicroserviceClassLoaderFactory();
 
   @Override
-  public ClassLoader create(String microserviceName, String version) {
+  public ClassLoader create(String appId, String microserviceName, String version) {
     return Thread.currentThread().getContextClassLoader();
```
WillemJiang: If we just return the TCC, maybe we need to update the method name for it.

edge/edge-core/src/main/resources/microservice.yaml
```diff
@@ -0,0 +1,3 @@
+cse-config-order: -500
```
WillemJiang: Please add the license header here.

service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
```diff
@@ -95,6 +98,36 @@ public void init() {
     eventBus.register(this);
   }
 
+  protected void initAppManager() throws InstantiationException, IllegalAccessException, ClassNotFoundException {
+    appManager = new AppManager(eventBus);
+
+    String microserviceVersionFactoryClass = serviceRegistryConfig.getMicroserviceVersionFactory();
+    if (microserviceVersionFactoryClass == null) {
+      return;
```
WillemJiang: It's better to throw exception here.

core/src/main/java/io/servicecomb/core/definition/classloader/DefaultMicroserviceClassLoaderFactory.java
wujimin: this just a Default factory, another PrivateXXXFactory will do a real create

service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
wujimin: in this version, we did not remove old cacheManager, and new cacheManager only used by edge.
so microserviceVersionFactoryClass is null, means not work with edge, will use old cacheManager, can not throw exception

WillemJiang: Can we add a comment to the code for it?

wujimin: done.

edge/edge-core/src/main/resources/microservice.yaml
wujimin: discussed, no need to add license here.

PR #239, opened 2017-11-03
JAV-228 hidden method should be skipped in swagger method lookup

eric-lee-ltk:
Signed-off-by: eric-lee-ltk <lidagang1@huawei.com>


foundations/foundation-common/pom.xml
```diff
@@ -70,5 +70,9 @@
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 		</dependency>
-	</dependencies>
+    <dependency>
```
liubao68: minor : this file is not properly indented

foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/utils/ReflectUtils.java
```diff
@@ -55,6 +57,10 @@ public static void setField(Class<?> cls, Object instance, String fieldName, Obj
   public static Method findMethod(Class<?> cls, String methodName) {
     for (Method method : cls.getMethods()) {
       if (method.getName().equals(methodName)) {
+        if (method.isAnnotationPresent(ApiOperation.class) &&
+            method.getAnnotation(ApiOperation.class).hidden()) {
```
wujimin: not  so good  that associate swagger with reflection
you  can do this  in swagger generator project

PR #247, opened 2017-11-06
JAV-489 resolve Edge service return Cotent-Length error

jeho0815:
if return the content-lengh from server, can't change the body, so remove this header


common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -141,7 +141,9 @@ protected void sendResponse(Response response) throws Exception {
     if (response.getHeaders().getHeaderMap() != null) {
       for (Entry<String, List<Object>> entry : response.getHeaders().getHeaderMap().entrySet()) {
         for (Object value : entry.getValue()) {
-          responseEx.addHeader(entry.getKey(), String.valueOf(value));
+          if (!entry.getKey().equalsIgnoreCase(HttpHeaders.CONTENT_LENGTH)) {
```
wujimin: maybe it's better to remove HttpHeaders.CONTENT_LENGTH after the loop

jeho0815: has no method to do this, to do this, has add an another method

PR #250, opened 2017-11-08
JAV-493 send form data must use urlcoding and remove null value to send

jeho0815:


common/common-rest/src/test/java/io/servicecomb/common/rest/codec/param/TestRestClientRequestImpl.java
```diff
@@ -0,0 +1,56 @@
+package io.servicecomb.common.rest.codec.param;
```
WillemJiang: Please add license header here.

PR #251, opened 2017-11-09
JAV-429 If the schema already exists, it will be re-registered

wWX495170:
As a developer, I need to get informed if my contract in local doesn't match the content already in SC so that I can change my code to handle it and avoid the incorrect running behavior


wujimin:
must not do this

only allow this operation in dev mode.

PR #259, opened 2017-11-13
removed useless abstract class

qiuxd:
We could use default method in interface instead of creating an abstract class that does the same thing.


WillemJiang:
Thanks for the PR, as the AbstractHandler may be used by user, we cannot just remove this class simply without any warning message.  

My suggestion is we just keep the class there and add the Deprecated annotation to let the user know about it, and then we can remove it in the next release.

PR #261, opened 2017-11-15
JAV-479 Discovery optimize

wujimin:
details is in https://servicecomb.atlassian.net/browse/JAV-479

in this issue, create new filter mechanism, old filters will rewrite in new Filter later.


service-registry/src/main/java/io/servicecomb/serviceregistry/discovery/DiscoveryTree.java
```diff
@@ -0,0 +1,138 @@
...
+    return doDiscovery(context, parent);
+  }
+
+  protected DiscoveryTreeNode doDiscovery(DiscoveryContext context, DiscoveryTreeNode parent) {
+    for (int idx = 0; idx < filters.size();) {
+      DiscoveryFilter filter = filters.get(idx);
+      context.setCurrentNode(parent);
+
+      DiscoveryTreeNode child = filter.discovery(context, parent);
+      if (child == null) {
```
liubao68: If child is null, will generate a NullPointerException, it's the same

wujimin: yes, but from null exception, we can  not know which filter throw it.

foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/cache/VersionedCache.java
```diff
@@ -0,0 +1,134 @@
...
+
+  protected Object data;
+
+  protected IsEmpty isEmpty = this::isCommonEmpty;
+
+  private boolean isCommonEmpty() {
+    return data == null;
+  }
+
+  private boolean isMapEmpty() {
```
WillemJiang: If the data is not a map , we will get the class case exception. 
I don't like these three method.

wujimin: yes, i don't like this too.
but data in cache is not always the same type.
eg: in discoveryTree, before endpointFilter, discoveryTreeNode data is Map<String, MicroserviceInstance>, after endpointFilter, discoverTreeNode data is Collection<Endpoint> or Collection<CseServer>

caller know what's data type in cache, and call different method, if invoke wrong method, then it's a bug, developers should fix it.

wujimin: isXXXEmpty methods have no problems, because they are private, and used by data type.
not so good methods are: mapData/collectionData/arrayData

PR #262, opened 2017-11-15
JAV-495 set url pattern default value

yhs0092:
See details in https://servicecomb.atlassian.net/browse/JAV-495

If the RestServletInitializer is invoked and URL pattern is not specified, set a default URL pattern "/*" to avoid the initialization failure of JavaChassis servlet.


yhs0092:
@wujimin 这样修改行吗？RestServletInjector在RestServletInitializer和CseXmlWebApplicationContext有调用，现在的修改方案应该只影响到RestServletInitializer。

demo/demo-spring-boot-convert/demo-spring-boot-springmvc-server-convert/src/main/java/io/servicecomb/demo/springboot/convert/springmvc/handler/MyHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MyHandler extends AbstractHandler {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MyHandler.class);
+
+  @Override
+  public void handle(Invocation invocation, AsyncResponse asyncResponse) throws Exception {
+    LOGGER.info("If you see this log, that means this demo project has been converted to ServiceComb framework.");
```
wujimin: please make the case automatically
eg:
　if response type is User
　in handler, assign user.name = "myHandler"
　then consumer determine, if user.name is empty, means did not run with serviceComb handlers

yhs0092:
This PR is abandoned. And I will create another PR after improving the demo.

PR #264, opened 2017-11-15
JAV-492 Edge service not properly handle 404 requests

wWX495170:


edge/edge-core/src/test/java/io/servicecomb/edge/core/TestEdgeInvocation.java
```diff
@@ -258,6 +267,11 @@ protected void doInvoke() throws Throwable {
 
       doInvokeClassLoader = Thread.currentThread().getContextClassLoader();
     }
+  
+    @Override
+    public void sendFailResponse(Throwable throwable) {
+      doSendFailResponse = Thread.currentThread().getContextClassLoader();
```
WillemJiang: It's better to use a flag instead of using the classloader to check if the method is called or not.

PR #266, opened 2017-11-15
JAVA-421 use the new style of loadbalance rule invalid

laijianbin:


handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -70,17 +70,22 @@ public Thread newThread(Runnable r) {
   private final Object lock = new Object();
 
   private String policy = null;
+  private String strategy = null;
 
   @Override
   public void handle(Invocation invocation, AsyncResponse asyncResp) throws Exception {
     String p = Configuration.INSTANCE.getPolicy(invocation.getMicroserviceName());
```
WillemJiang: It could better if we have an interface to notify the configure change to save some time in the handle method.

PR #268, opened 2017-11-16
JAV-358 Operation filter

wujimin:
edge service do not dispatch microservice invocation to the old instance that do not support the new operation


core/src/main/resources/META-INF/spring/cse.bean.xml
```diff
@@ -33,6 +33,7 @@
 	<bean class="io.servicecomb.core.CseApplicationListener">
 	</bean>
 
-	<bean id="cse.executor.default" class="io.servicecomb.core.executor.FixedThreadExecutor"></bean>
+	<bean id="cse.executor.groupThreadPool" class="io.servicecomb.core.executor.FixedThreadExecutor"></bean>
+    <alias name="cse.executor.groupThreadPool" alias="cse.executor.default"/>
```
liubao68: indention

demo/demo-edge/edge-service/src/main/java/io/servicecomb/demo/edge/service/EdgeDispatcher.java
```diff
@@ -41,10 +44,13 @@ public void init(Router router) {
   protected void onRequest(RoutingContext context) {
     Map<String, String> pathParams = context.pathParams();
     String microserviceName = pathParams.get("param0");
-    String path = "/" + pathParams.get("param1");
+    String pathVersion = pathParams.get("param1");
+    String path = context.request().path().substring(4);
```
WillemJiang: How can we tell the path by removing the first 4 characters?

demo/demo-edge/hiboard-consumer/src/main/java/io/servicecomb/demo/edge/consumer/Consumer.java
```diff
@@ -55,17 +69,53 @@ public Consumer() {
...
+    invoke("/v1/dec", 2, 1, decV1Result);
+    invoke("/v1/dec", 3, 1, decV1Result);
+
+    invoke("/v2/add", 2, 1, addV2Result);
+    invoke("/v2/add", 3, 1, addV2Result);
+
+    invoke("/v2/dec", 2, 1, decV2Result);
+    invoke("/v2/dec", 3, 1, decV2Result);
+
+    printResults("v1/add", addV1Result);
```
WillemJiang: We should use assert instead of printResults to check the result data.

demo/demo-edge/edge-service/src/main/java/io/servicecomb/demo/edge/service/EdgeDispatcher.java
wujimin: this is a demo
and the path rule is :/api/ms/version/......

edge/edge-core/src/main/java/io/servicecomb/edge/core/EdgeBootListener.java
```diff
@@ -0,0 +1,45 @@
...
+  @Override
+  public void onBootEvent(BootEvent event) {
+    if (!EventType.BEFORE_PRODUCER_PROVIDER.equals(event.getEventType())) {
+      return;
+    }
+
+    if (DynamicPropertyFactory.getInstance()
+        .getStringProperty(ExecutorManager.KEY_EXECUTORS_DEFAULT, null)
+        .get() != null) {
+      return;
```
WillemJiang: we may need to do some log here.

edge/edge-core/src/main/java/io/servicecomb/edge/core/EdgeBootListener.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // change default to reactive mode
```
WillemJiang: log the edge service works in reactive mode.

core/src/main/resources/META-INF/spring/cse.bean.xml
wujimin: done

demo/demo-edge/hiboard-consumer/src/main/java/io/servicecomb/demo/edge/consumer/Consumer.java
wujimin: done

edge/edge-core/src/main/java/io/servicecomb/edge/core/EdgeBootListener.java
wujimin: done

edge/edge-core/src/main/java/io/servicecomb/edge/core/EdgeBootListener.java
wujimin: done

PR #271, opened 2017-11-16
JAV-496 Async servlet timeout will cause 500 error

wWX495170:


transports/transport-rest/transport-rest-servlet/src/test/java/io/servicecomb/transport/rest/servlet/TestRestAsyncListener.java
```diff
@@ -111,11 +111,10 @@ void flushBuffer() throws IOException {
 
   @Test
   public void onTimeoutExecuting() throws IOException {
```
wujimin: after remove RestConst.REST_STATE_EXECUTING, this case is useless, remote it.

transports/transport-rest/transport-rest-servlet/src/test/java/io/servicecomb/transport/rest/servlet/TestRestAsyncListener.java
```diff
@@ -110,15 +110,6 @@ void flushBuffer() throws IOException {
   }
 
   @Test
-  public void onTimeoutExecuting() throws IOException {
-    request.setAttribute(RestConst.REST_STATE_EXECUTING, true);
-
-    listener.onTimeout(event);
```
liubao68: coverd by onTimeoutCommitted & onTimeoutNotCommitted

PR #276, opened 2017-11-18
[JAV-504] instance not unregistered when shutdown and auto-discovery …

liubao68:
1.instance not unregistered when shutdown and auto-discovery  not work because not use the correct new sc api
2. and we choose any address configured first in a random fashion

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/api/Const.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/api/Const.java
@@ -142,8 +142,6 @@ public final class Const {
 
   public static final String REGISTRY_SERVICE_NAME = "SERVICECENTER";
 
-  public static final String REGISTRY_VERSION = "latest";
-
   public static final String APP_SERVICE_SEPARATOR = ":";
 
   public static final String PATH_CHECKSESSION = "checksession";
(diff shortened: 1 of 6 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/api/Const.java
```diff
@@ -37,6 +37,15 @@ private Const() {
 
     public static final String V4_PREFIX = String.format("/v4/%s/registry", DOMAIN_NAME);
 
+    public static final String REGISTRY_VERSION;
+    static {
+      if (VERSION_V3.equals(CURRENT_VERSION)) {
+        REGISTRY_VERSION = "3.0.0"; // old is strange
+      } else {
+        REGISTRY_VERSION = "0.4.0";
```
WillemJiang: I think we just need keep tracking the API version instead of the service center version.

liubao68: I am going to change this to latest according to wujimin's suggestion

service-registry/src/main/java/io/servicecomb/serviceregistry/api/Const.java
```diff
@@ -142,8 +142,6 @@ private Const() {
 
   public static final String REGISTRY_SERVICE_NAME = "SERVICECENTER";
 
-  public static final String REGISTRY_VERSION = "3.0.0";
```
WillemJiang: It's not a good practise to delete the public static constant.

PR #278, opened 2017-11-20
JAV-505 HttpServletReqeust.getContextPath returns the default value ""

coolhongluo:
return "", not throw a exception


coolhongluo:
@wujimin @WillemJiang pls review

PR #289, opened 2017-11-21
jav-479 optimizeLB add filter to distiguish instance enviroment and u…

mt-monster:
…p stage ,this is relate to cse-554

### DIFF
--- a/handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/filter/StageDistinguishFilter.java
+++ b/handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/filter/StageDistinguishFilter.java
@@ -1,0 +1,30 @@
+package io.servicecomb.loadbalance.filter;
+
+import com.netflix.loadbalancer.Server;
+import io.servicecomb.loadbalance.CseServer;
+import io.servicecomb.loadbalance.ServerListFilterExt;
+import io.servicecomb.serviceregistry.RegistryUtils;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstanceStatus;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class StageDistinguishFilter implements ServerListFilterExt {
+    @Override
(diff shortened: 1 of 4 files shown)


handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/filter/StageDistinguishFilter.java
```diff
@@ -0,0 +1,30 @@
...
+import io.servicecomb.loadbalance.CseServer;
+import io.servicecomb.loadbalance.ServerListFilterExt;
+import io.servicecomb.serviceregistry.RegistryUtils;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
+import io.servicecomb.serviceregistry.api.registry.MicroserviceInstanceStatus;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class StageDistinguishFilter implements ServerListFilterExt {
```
wujimin: please use DiscoverFilter to impl this

and i heard that, SC will only return same stage instances, so this filter is useless?

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -109,25 +109,20 @@ public void pullInstances() {
     setInstances(pulledInstances);
   }
 
-  private void setInstances(List<MicroserviceInstance> pulledInstances) {
-    synchronized (lock) {
-      instances = pulledInstances
-          .stream()
-          .filter(instance -> {
-            return MicroserviceInstanceStatus.UP.equals(instance.getStatus());
```
liubao68: It's better a add this logic here and filter out all other env's instances. Because this is a required feature, and filters are optional ones.

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/filter/StageDistinguishFilter.java
liubao68: this feature given by SC, they will return all instances of different env's. STAGE is another scope of separation.

PR #297, opened 2017-11-23
JAV-524 Optimize edge executor

wujimin:
use the same mechanism of normal microservice producer

before optimize, edge's decode and filter run in net thread.

### DIFF
--- a/common/common-rest/src/main/java/io/servicecomb/common/rest/RestProducerInvocation.java
+++ b/common/common-rest/src/main/java/io/servicecomb/common/rest/RestProducerInvocation.java
@@ -18,13 +18,6 @@ package io.servicecomb.common.rest;
 
 import java.util.List;
 
-import javax.ws.rs.core.Response.Status;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import io.servicecomb.common.rest.codec.RestCodec;
-import io.servicecomb.common.rest.definition.RestOperationMeta;
 import io.servicecomb.common.rest.filter.HttpServerFilter;
 import io.servicecomb.common.rest.locator.OperationLocator;
 import io.servicecomb.common.rest.locator.ServicePathManager;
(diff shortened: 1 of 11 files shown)


common/common-rest/src/main/java/io/servicecomb/common/rest/RestProducerInvocation.java
```diff
@@ -65,115 +52,26 @@ public void invoke(Transport transport, HttpServletRequestEx requestEx, HttpServ
...
-      QueueMetricsData reqQueue = MetricsServoRegistry.getOrCreateLocalMetrics()
-          .getOrCreateQueueMetrics(restOperationMeta.getOperationMeta().getMicroserviceQualifiedName());
-      reqQueue.incrementTotalCount();
-      Long timeInQueue = metricsData.getQueueEndTime() - metricsData.getQueueStartTime();
-      reqQueue.setTotalTime(reqQueue.getTotalTime() + timeInQueue);
-      reqQueue.setMinLifeTimeInQueue(timeInQueue);
-      reqQueue.setMaxLifeTimeInQueue(timeInQueue);
-      reqQueue.decrementCountInQueue();
-    }
+  protected OperationLocator locateOperation(ServicePathManager servicePathManager) {
```
WillemJiang: What will happen if we cannot find the OperationLocator?

wujimin: if locate failed, throw 404 or 405 exception.

transports/transport-rest/transport-rest-servlet/src/main/java/io/servicecomb/transport/rest/servlet/RestServletProducerInvocation.java
```diff
@@ -17,20 +17,17 @@
...
 import io.servicecomb.foundation.vertx.http.StandardHttpServletRequestEx;
 
 public class RestServletProducerInvocation extends RestProducerInvocation {
   @Override
-  protected RestOperationMeta findRestOperation() {
-    RestOperationMeta restOperationMeta = super.findRestOperation();
+  protected void findRestOperation() {
+    super.findRestOperation();
 
     boolean cacheRequest = collectCacheRequest(restOperationMeta.getOperationMeta());
```
WillemJiang: this variable is useless.

transports/transport-rest/transport-rest-servlet/src/main/java/io/servicecomb/transport/rest/servlet/RestServletProducerInvocation.java
```diff
@@ -17,20 +17,17 @@
...
 
 import io.servicecomb.common.rest.RestProducerInvocation;
-import io.servicecomb.common.rest.definition.RestOperationMeta;
 import io.servicecomb.common.rest.filter.HttpServerFilter;
 import io.servicecomb.core.definition.OperationMeta;
 import io.servicecomb.foundation.vertx.http.StandardHttpServletRequestEx;
 
 public class RestServletProducerInvocation extends RestProducerInvocation {
   @Override
-  protected RestOperationMeta findRestOperation() {
```
WillemJiang: It looks like the new findRestOperation method is just throw the exception if it cannot find the operation.  I'm not sure if this change has some side effect on the user side code.

transports/transport-rest/transport-rest-servlet/src/main/java/io/servicecomb/transport/rest/servlet/RestServletProducerInvocation.java
wujimin: which one?
cacheRequst?   it's setted to requestEx,  if copy request body controlled by this avriable.

transports/transport-rest/transport-rest-servlet/src/main/java/io/servicecomb/transport/rest/servlet/RestServletProducerInvocation.java
wujimin: not change, just move from one class to another class.

transports/transport-rest/transport-rest-servlet/src/main/java/io/servicecomb/transport/rest/servlet/RestServletProducerInvocation.java
WillemJiang: I missed the second line , it makes sense now.

PR #310, opened 2017-11-25
Cse-722 content-type convert to small-spelling problem

mt-monster:
this resolve the content-type convert to small-spelling problem

### DIFF
--- a/common/common-rest/src/main/java/io/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
+++ b/common/common-rest/src/main/java/io/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
@@ -63,7 +63,7 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
       }
 
       String contentType = request.getContentType();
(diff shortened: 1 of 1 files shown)


common/common-rest/src/main/java/io/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
```diff
@@ -63,7 +63,7 @@ public Object getValue(HttpServletRequest request) throws Exception {
       }
 
       String contentType = request.getContentType();
-      if (contentType != null && !contentType.startsWith(MediaType.APPLICATION_JSON)) {
+      if (contentType != null && !contentType.toLowerCase().startsWith(MediaType.APPLICATION_JSON)) {
```
wujimin: 1.toLowerCase(Locale.US)?
2.where is accept bug fix?

mt-monster: I prefer to use Locale.ROOT(recommended by java 8) than Locale.US or Locale.ENGLISH ( by java 7 and below ) ,but this two would be alright.

PR #316, opened 2017-11-27
JAV-501 provide authentication between service

coolhongluo:
As a developer, I need to have authentication in service provider to protect the microservice from unauthorized access


WillemJiang:
The new added UT is not enough for increase the code coverage. Please add an integration test for it.

PR #317, opened 2017-11-27
Jav 529 dubbo migration

seanyinx:


seanyinx:
@liubao68 we already added a migration guide here: https://github.com/ServiceComb/ServiceComb.github.io/pull/133

PR #321, opened 2017-11-27
fixed typos of TcpResonseCallback

m20082008m:


foundations/foundation-vertx/src/main/java/io/servicecomb/foundation/vertx/client/tcp/TcpResponseCallback.java
```diff
@@ -18,6 +18,6 @@
 
 import io.servicecomb.foundation.vertx.AsyncResultCallback;
 
-public interface TcpResonseCallback extends AsyncResultCallback<TcpData> {
+public interface TcpResponseCallback extends AsyncResultCallback<TcpData> {
```
WillemJiang: We need to still keep the old wrong API with deprecated annotation to warn the user.

m20082008m: fixed, thanks for pointing out.

wujimin:
keep the old interface
and create new interface, but all input parameter changed to new interface

i think it's not compatible.
but maybe nobody do this type extension now?
we change the name directly?

WillemJiang:
@wujimin  You are right, the other method is public which could affect to the user. 
If the send method is lower level API, may be we can add some words on the release note to remind user for that.

PR #331, opened 2017-11-28
JAV-429 Modify the startup policy of SDK

wWX495170:
JAV-429 Modify the startup policy of SDK
added schema summary,
added batch registration,
added to determine whether the servicecenter environment's runmode is dev,

### DIFF
--- a/providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/schema/PojoProducers.java
+++ b/providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/schema/PojoProducers.java
@@ -39,6 +39,14 @@ public class PojoProducers implements BeanPostProcessor {
     return pojoMgr.values();
   }
 
+  /**
+   * @deprecated  Replaced by {@link #getProducers()}
+   */
+  @Deprecated
+  public Collection<PojoProducerMeta> getProcucers() {
+    return getProducers();
+  }
+
   @Override
   public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
     return bean;


service-registry/src/main/java/io/servicecomb/serviceregistry/api/Sha.java
```diff
@@ -0,0 +1,66 @@
...
+package io.servicecomb.serviceregistry.api;
+
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ *Created by   on 2017/11/22.
+ */
+public class Sha {
```
wujimin: it seems guava already has this utils?

service-registry/src/main/java/io/servicecomb/serviceregistry/client/LocalServiceRegistryClientImpl.java
```diff
@@ -334,4 +335,22 @@ public boolean updateInstanceProperties(String microserviceId, String microservi
     }
     return true;
   }
+
+  @Override
+  public ServiceCenter getServiceCenter() {
```
wujimin: left these methods empty not good.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -265,6 +328,51 @@ public boolean registerSchema(String microserviceId, String schemaId, String sch
     return false;
   }
 
+  @Override
+  public boolean syncSchemasAll(Microservice microservice) {
```
wujimin: if there are 600 schemas, will this request too big?

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -142,12 +152,57 @@ private boolean registerSchemas() {
...
     }
 
     return true;
   }
+
+  private boolean serviceCenterEnv() {
+    if (Const.SERVICECENTER_RUNMODE_DEV.equals(srClient.getServiceCenter().getRunMode())) {
+      LOGGER.warn("The current servicecenter environment runmode is {}", Const.SERVICECENTER_RUNMODE_DEV);
+      return true;
+    } else if (Const.SERVICECENTER_RUNMODE_PROD.equals(srClient.getServiceCenter().getRunMode())) {
```
wujimin: already return, no need else

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -142,12 +152,57 @@ private boolean registerSchemas() {
...
+            }
+          }
         }
       }
     }
 
     return true;
   }
+
+  private boolean serviceCenterEnv() {
```
wujimin: serviceCenterEnv is not clear
it seems that to determine if it is dev mode.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/ServiceRegistryClient.java
```diff
@@ -68,12 +75,24 @@
    */
   boolean registerSchema(String microserviceId, String schemaId, String schemaContent);
 
+  /**
+   * 
+   * 同步本地schema
+   */
+  boolean syncSchemasAll(Microservice microservice);
```
wujimin: we need a batch api to determine if some schemas exist.

service-registry/src/main/java/io/servicecomb/serviceregistry/api/Sha.java
wWX495170: yes,use guava is better

service-registry/src/main/java/io/servicecomb/serviceregistry/client/LocalServiceRegistryClientImpl.java
wWX495170: ok，i'll use some information to mark

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
wWX495170: ok，i'll modify it to return false directly,without using else

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
wWX495170: Use String instead of boolean to do the return value type, and then judge?

wWX495170: only in the dev can perform changes currently, so determine if it is dev mode in the method,we can only care about it'll return true or false

service-registry/src/main/java/io/servicecomb/serviceregistry/client/ServiceRegistryClient.java
wWX495170: this time i use the batch upload interface that SC provided,if not good enough, we can use the better plan

service-registry/src/main/java/io/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
wujimin: yes.
but method name is serviceCenterEnv, not relative to dev mode.

wWX495170: ok,i've changed the metnod's name to serviceCenterEnvIsDev

service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
wWX495170: so how to deal with such a scene? if local schema is not same with SC's

Issue #325, opened 2017-11-28
naming of constant variables should use proper separator between words


eric-lee-ltk:
The naming of constant variables in [Configuration.java](https://github.com/ServiceComb/ServiceComb-Java-Chassis/blob/master/handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/Configuration.java) should obey the [constant naming of google style](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names). However, now the naming use property name without any separators, which is hard to read and not a good practice.

PR #333, opened 2017-11-29
Update Configuration.java

Cheukyin:
change the naming of constant variables in Configuration.java to obey the constant naming of google style


eric-lee-ltk:
fix #325

PR #334, opened 2017-11-29
JAV-543 Fix dns resolve

wujimin:
it's netty's bug.
latest vertx had upgrade netty dependence.
we can upgrade to latest vertx 3.5.0

### DIFF
--- a/foundations/foundation-vertx/src/main/java/io/vertx/ext/web/impl/MimeTypesUtils.java
+++ b/foundations/foundation-vertx/src/main/java/io/vertx/ext/web/impl/MimeTypesUtils.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017 Huawei Technologies Co., Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.vertx.ext.web.impl;
+
+import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


foundations/foundation-vertx/src/main/java/io/vertx/ext/web/impl/MimeTypesUtils.java
```diff
@@ -0,0 +1,69 @@
+package io.vertx.ext.web.impl;
```
WillemJiang: Please add license header.

foundations/foundation-vertx/src/test/java/io/vertx/ext/web/impl/TestMimeTypesUtils.java
```diff
@@ -0,0 +1,81 @@
+package io.vertx.ext.web.impl;
```
WillemJiang: License header.

java-chassis-dependencies/pom.xml
```diff
@@ -1,1001 +1,1001 @@
...
+    <artifactId>java-chassis</artifactId>
+    <version>0.4.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>java-chassis-dependencies</artifactId>
+  <packaging>pom</packaging>
+
+  <properties>
+    <jackson.version>2.8.10</jackson.version>
+    <vertx.version>3.5.0</vertx.version>
```
WillemJiang: This file is reformated, Can we just update the version ?

foundations/foundation-vertx/src/main/java/io/vertx/ext/web/impl/MimeTypesUtils.java
wujimin: done

foundations/foundation-vertx/src/test/java/io/vertx/ext/web/impl/TestMimeTypesUtils.java
wujimin: done

java-chassis-dependencies/pom.xml
wujimin: first commit in yellow cloud, forgot if reformat it.
it's not important, right?

PR #336, opened 2017-11-29
JAV-535 provide an access log printer handler

yhs0092:
see details in [https://servicecomb.atlassian.net/browse/JAV-535](https://servicecomb.atlassian.net/browse/JAV-535)

### DIFF
--- a/core/src/main/java/io/servicecomb/core/handler/impl/AccessLogHandler.java
+++ b/core/src/main/java/io/servicecomb/core/handler/impl/AccessLogHandler.java
@@ -1,0 +1,28 @@
+package io.servicecomb.core.handler.impl;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.servicecomb.core.Const;
+import io.servicecomb.core.Handler;
+import io.servicecomb.core.Invocation;
+import io.servicecomb.swagger.invocation.AsyncResponse;
+
+/**
+ * access log printer handler
+ */
+public class AccessLogHandler implements Handler {
+  private static final Logger LOGGER = LoggerFactory.getLogger("accesslog");
+
+  @Override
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/servicecomb/core/handler/impl/AccessLogHandler.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * access log printer handler
+ */
+public class AccessLogHandler implements Handler {
+  private static final Logger LOGGER = LoggerFactory.getLogger("accesslog");
+
+  @Override
+  public void handle(Invocation invocation, AsyncResponse asyncResp) throws Exception {
+    invocation.next(response -> {
+      LOGGER.info("{} - {} \"{}\" {}",
```
wujimin: 1.handler can only log "normal" invocation, 404 or decode failed invocation can not be logged.
   is this what we want?
2.log when invocation finished not started?
3.if some handler throw exception, then will not log?

core/src/main/java/io/servicecomb/core/invocation/InvocationFactory.java
```diff
@@ -39,6 +39,7 @@ public static Invocation forConsumer(ReferenceConfig referenceConfig, OperationM
         operationMeta,
         swaggerArguments);
     invocation.addContext(Const.SRC_MICROSERVICE, getMicroserviceName());
+    invocation.addContext(Const.SRC_INSTANCE, getInstanceId());
```
wujimin: maybe need to add microserviceId too.

yhs0092:
will give another PR according to [https://servicecomb.atlassian.net/browse/JAV-561](https://servicecomb.atlassian.net/browse/JAV-561)

PR #337, opened 2017-11-30
remove unused code of TestJavassistUtils and make it readable 

wldandan:
* remove unused method definition.
* add IDEA project to gitignore file
* make it readable about {}

### DIFF
--- a/common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
+++ b/common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
@@ -80,7 +80,6 @@ public class TestJavassistUtils {
     String intfName = "cse.ut.TestAddParameter";
     classConfig.setClassName(intfName);
 
-    // List<String> method(Map<String, String> map, Set<String> set)
     MethodConfig methodConfig = new MethodConfig();
     methodConfig.setName("method");
     methodConfig.setResult(TypeFactory.defaultInstance().constructCollectionType(List.class, String.class));
@@ -192,10 +191,8 @@ public class TestJavassistUtils {
 
   @Test
   public void managerClassPool() {
-    ClassLoader classLoader1 = new ClassLoader() {
-    };
-    ClassLoader classLoader2 = new ClassLoader() {
-    };
+    ClassLoader classLoader1 = new ClassLoader() { };
+    ClassLoader classLoader2 = new ClassLoader() { };
 
     ClassPool p1 = Deencapsulation.invoke(JavassistUtils.class, "getOrCreateClassPool", classLoader1);
     ClassPool p2 = Deencapsulation.invoke(JavassistUtils.class, "getOrCreateClassPool", classLoader2);
--- a/common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
+++ b/common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
@@ -54,29 +54,23 @@ public class TestJavassistUtils {
 
     Class<?> cls = JavassistUtils.createClass(classConfig);
 
-    // intField
     Field field = cls.getField("intField");
     Assert.assertEquals(Integer.class, field.getType());
 
     Method method = cls.getMethod("getIntField");
     Assert.assertEquals(Integer.class, method.getReturnType());
-    method = cls.getMethod("setIntField", Integer.class);
 
-    // intArrayField
     field = cls.getField("intArrayField");
     Assert.assertEquals(int[].class, field.getType());
 
     method = cls.getMethod("getIntArrayField");
     Assert.assertEquals(int[].class, method.getReturnType());
-    method = cls.getMethod("setIntArrayField", int[].class);
 
-    // listStringField
     field = cls.getField("listStringField");
     Assert.assertEquals("java.util.List<java.lang.String>", field.getGenericType().getTypeName());
 
     method = cls.getMethod("getListStringField");
     Assert.assertEquals("java.util.List<java.lang.String>", method.getGenericReturnType().getTypeName());
-    method = cls.getMethod("setListStringField", List.class);
   }
 
   @Test
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,8 @@ __pycache__/
 # IntelliJ, based on http://devnet.jetbrains.net/docs/DOC-1186
 .idea/
 *.iml
+*.ipr
+*.iws
 
 # logs and trace
 *.log


common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
```diff
@@ -54,29 +54,23 @@ public void testField() throws Exception {
...
     Assert.assertEquals(int[].class, method.getReturnType());
-    method = cls.getMethod("setIntArrayField", int[].class);
 
-    // listStringField
     field = cls.getField("listStringField");
     Assert.assertEquals("java.util.List<java.lang.String>", field.getGenericType().getTypeName());
 
     method = cls.getMethod("getListStringField");
     Assert.assertEquals("java.util.List<java.lang.String>", method.getGenericReturnType().getTypeName());
-    method = cls.getMethod("setListStringField", List.class);
```
wujimin: deleted "getMethod" is to ensure setter is exist and has correct signature
if has any problems, will throw exception.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
```diff
@@ -86,7 +80,6 @@ public void testAddParameter() {
     String intfName = "cse.ut.TestAddParameter";
     classConfig.setClassName(intfName);
 
-    // List<String> method(Map<String, String> map, Set<String> set)
```
wujimin: this comments is what the target method will be.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
```diff
@@ -198,10 +191,8 @@ public void testGetNameForGenerateCode() {
 
   @Test
   public void managerClassPool() {
-    ClassLoader classLoader1 = new ClassLoader() {
-    };
-    ClassLoader classLoader2 = new ClassLoader() {
-    };
+    ClassLoader classLoader1 = new ClassLoader() { };
```
wujimin: after format code, it will rollback to be the deleted style.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
WillemJiang: We need to use assert the verify the call result.

wldandan: If "getMethod" has logic as you mentioned, I think it should be verified rather than just let it be a caller. For example, It is better if we can add the following asset:

method = cls.getMethod("setListStringField", List.class);
assertNotNull(cls.getMethod("setListStringField", List.class))

I will do that in another PR

WillemJiang: @wldandan Please add a new commit for it.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wldandan: OK. fair enough. 
It is not a big issue, but I do not understand why we need to a line break if there is nothing inside.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wujimin: infact, if cls.getMethod return null, will throw exception inside getMethod
so the assert is invalid.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wujimin: here, we create a new class in memory, and add a new method in the class
if we did not comments the targt method signature, it's hard to know what's the purpose

so i thinks, comments can be better, but delete it is not good.

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wujimin: agree.
but it seems that hard to write this rule both in IntelliJ and eclipse

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wujimin: if you care this so much, you can invoke org.springframework.util.ReflectionUtils.findMethod(Class<?>, String, Class<?>...)
then if method not exists, will get a null result
but this method invoked getDeclaredMethod inside it, not invoked getMethod, so after assert null, you must assert if it's public.

wldandan: Ok. my thought is straightforward. 
If some methods are called, it should be verified about the effects. That is the purpose of test case. 
As you mentioned, "deleted "getMethod" is to ensure setter is exist and has correct signature if has any problems, will throw exception."

Back the discussion, what the best way you think we should do to verify if do call the following method, could you please suggest? 

method = cls.getMethod("setListStringField", List.class);

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wldandan: Is it reasonable if we define another method to hold the structure or use test case to clarify it? I think it is more understandable and believable. 
Or should we add more detailed comments to mention that rather than just a method definition?

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
wldandan: Leave it as an issue and put it in the queue. :)

common/common-javassist/src/test/java/io/servicecomb/common/javassist/TestJavassistUtils.java
WillemJiang: We still need to verify the method, if the getMethod throw the exception JUnit can still know about and tell us something is wrong.  As a client we want to make sure is the method is not null and we can still use it.   If we want to verify the non-exist method, we also need to add expected exception into JUnit test case to make sure the exception is thrown.

wldandan: @WillemJiang +1. To keep the scope limited, I think we can add another test case to verify "getmethod" for the scenario about throwing exception.
I will do that soon. 
Thanks all guys.

PR #345, opened 2017-12-03
make name simplified & add private constructor for singleton pattern.

wldandan:
* MicroserviceClassLoader has explicit meaning about Microservice relevant, so change the property from microserviceName to name.
* Add private constructor for singleton pattern.

### DIFF
--- a/core/src/main/java/io/servicecomb/core/definition/classloader/PrivateMicroserviceClassLoaderFactory.java
+++ b/core/src/main/java/io/servicecomb/core/definition/classloader/PrivateMicroserviceClassLoaderFactory.java
@@ -17,7 +17,10 @@
 package io.servicecomb.core.definition.classloader;
 
 public class PrivateMicroserviceClassLoaderFactory implements MicroserviceClassLoaderFactory {
+
   public static final MicroserviceClassLoaderFactory INSTANCE = new PrivateMicroserviceClassLoaderFactory();
+
+  private PrivateMicroserviceClassLoaderFactory() {}
 
   @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/servicecomb/core/definition/classloader/MicroserviceClassLoader.java
```diff
@@ -24,27 +24,27 @@
 
   private final String appId;
 
-  private final String microserviceName;
```
WillemJiang: As the microserviceName is used for a while, we cannot change the name easily due to back compatible reason.

wldandan: From business perspective, I think it is clear that microserceName is an argument.
For the classloader, it is core part and be separating from the argument outside. 
Additionally, it already named MicroserviceClassloader.

If it is already a convention, I am fine.

core/src/main/java/io/servicecomb/core/definition/classloader/PrivateMicroserviceClassLoaderFactory.java
```diff
@@ -19,6 +19,8 @@
 public class PrivateMicroserviceClassLoaderFactory implements MicroserviceClassLoaderFactory {
   public static final MicroserviceClassLoaderFactory INSTANCE = new PrivateMicroserviceClassLoaderFactory();
 
+  private PrivateMicroserviceClassLoaderFactory() {}
```
WillemJiang: This change is OK. Please resubmit the patch.

PR #346, opened 2017-12-03
JAV-546 register an anonymous subscriber to listen to instance registry event and trigger AFTER_REGISTRY event

yhs0092:
I've thought about several solutions on this issue.
- move sending AFTER_REGISTRY logic into ServiceCenterTask. But this means ServiceCenterTask need to get the instance of CseApplicationListener, and triggerEvent() of CseApplicationListener has to be changed to public.
- move sending AFTER_REGISTRY logic into a call-back function and pass it to ServiceCenterTask. When instance registry is completed, this function should be called. But this will change several Class from RegistryUtils to ServiceCenterTask.

I think both of them are not good. So I choose to write an anonymous subscriber and let it to listen to MicroserviceInstanceRegisterTask. Maybe this is the solution with the minimum change of code.

### DIFF
--- a/core/src/main/java/io/servicecomb/core/CseApplicationListener.java
+++ b/core/src/main/java/io/servicecomb/core/CseApplicationListener.java
@@ -129,25 +129,9 @@ public class CseApplicationListener
           schemaListenerManager.notifySchemaListener();
 
           triggerEvent(EventType.BEFORE_REGISTRY);
-          /*
-           * 由于注册实例是异步过程，所以后续的AFTER_REGISTRY消息不能立即发送，必须等到实例第一次注册成功时才能发送。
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/servicecomb/core/CseApplicationListener.java
```diff
@@ -124,9 +129,26 @@ public void onApplicationEvent(ApplicationEvent event) {
           schemaListenerManager.notifySchemaListener();
 
           triggerEvent(EventType.BEFORE_REGISTRY);
+          /*
```
wujimin: it's better to write new comments in english.
and maybe this logic can be a new method, only invoke in this place.

yhs0092: I've extracted this part into a new private method, and translated the comments into English. Please review it.

PR #348, opened 2017-12-04
JAV-550 support use _ to represent system environment variable config levels

qizha:
check https://servicecomb.atlassian.net/browse/jav-550 for details.. I also submit a discussion topic in google group for this issue


WillemJiang:
@qizha  Can you add an unit test for this new added feature?

qizha:
sure, I just added UT for the change

PR #352, opened 2017-12-05
fixed typos of regsiter and PojoProducer.getProcucers

### LINKED ISSUE #326: typos need to be fixed
There are some obvious typos in the code that need to be fixed. e.g. resiter -> register, getProcucers -> getProducers.

yhj7727:
fixed #326

### DIFF
--- a/providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/schema/PojoProducers.java
+++ b/providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/schema/PojoProducers.java
@@ -39,6 +39,14 @@ public class PojoProducers implements BeanPostProcessor {
     return pojoMgr.values();
   }
 
+  /**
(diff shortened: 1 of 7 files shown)


providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/schema/PojoProducers.java
```diff
@@ -35,7 +35,7 @@ public void registerPojoProducer(PojoProducerMeta pojoProducer) {
     pojoMgr.register(pojoProducer.getSchemaId(), pojoProducer);
   }
 
-  public Collection<PojoProducerMeta> getProcucers() {
```
WillemJiang: This is public method , we need to deprecated this method first and add new updated method for it.

yhj7727: already added it back

PR #353, opened 2017-12-05
JAV-551 fix address binding logic, 0.0.0.0 should be replaced with a valid address

yhs0092:
see details in [https://servicecomb.atlassian.net/browse/JAV-551](https://servicecomb.atlassian.net/browse/JAV-551)

### DIFF
--- a/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/NetUtils.java
+++ b/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/NetUtils.java
@@ -25,6 +25,7 @@ import java.net.ServerSocket;
 import java.net.SocketException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.UnknownHostException;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
@@ -35,6 +36,8 @@ import org.slf4j.LoggerFactory;
 public final class NetUtils {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(NetUtils.class);
+
+  private static final String COLON = ":";
 
   // one interface can bind to multiple address
   // we only save one ip for each interface name.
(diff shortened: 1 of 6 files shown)


yhs0092:
seems that test configuration in integration-tests need to be updated with this issue. Checking file ...

yhs0092:
After modifying some configuration in integration-tests, CI has passed. Evaluation on this modification is needed to avoid losing effectiveness of these tests or other side effects.

integration-tests/spring-zuul-tracing-tests/src/test/resources/microservice.yaml
```diff
@@ -7,7 +7,7 @@ cse:
     registry:
       address: http://127.0.0.1:30100
   rest:
-    address: 0.0.0.0:8080
+    address: 127.0.0.1:8080
```
liubao68: 为什么需要修改这里的地址？

yhs0092: 因为现在监听IP的时候不允许监听0.0.0.0了。我们的测试里面全都是provider配的监听0.0.0.0，consumer连接用的127.0.0.1，现在这样会连不上。
其他的测试我已经改成consumer端用NetUtils获取本机IP了。但是有的测试consumer是从配置文件中获取provider的IP做连接的，不好改，所以我就改成两方都用127.0.0.1了。

foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -201,4 +209,33 @@ public static boolean canTcpListen(InetAddress address, int port) {
...
+    }
+    String ip = address.substring(0, address.indexOf(COLON));
+    InetAddress inetAddress = null;
+    try {
+      inetAddress = InetAddress.getByName(ip);
+    } catch (UnknownHostException e) {
+      LOGGER.error("host {} is unknown.", ip);
+      return null;
+    }
+    if (inetAddress.isAnyLocalAddress()) {
```
WillemJiang: Is there any security reason for this change?
We need to tell add the release-note to tell user they cannot use "0.0.0.0" any more.

yhs0092: I misunderstood the requirement. I've communicated with liubao and jimin, and ensure that our current mechanism is right. This pull request will be closed later.

PR #357, opened 2017-12-05
[JAV-554]Fix bug that when booting use java system properties for ssl configuration

liubao68:
…ation

### DIFF
--- a/foundations/foundation-ssl/src/test/java/io/servicecomb/foundation/ssl/SSLOptionTest.java
+++ b/foundations/foundation-ssl/src/test/java/io/servicecomb/foundation/ssl/SSLOptionTest.java
@@ -37,6 +37,7 @@ import com.netflix.config.DynamicConfiguration;
 import com.netflix.config.DynamicPropertyFactory;
 import com.netflix.config.FixedDelayPollingScheduler;
 
+import io.servicecomb.config.ConfigUtil;
 import io.servicecomb.config.archaius.scheduler.NeverStartPollingScheduler;
 import mockit.Deencapsulation;
 import mockit.Mock;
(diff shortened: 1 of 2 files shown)


foundations/foundation-ssl/src/test/java/io/servicecomb/foundation/ssl/SSLOptionTest.java
```diff
@@ -230,6 +230,46 @@ public void testSSLOptionYamlOption2() throws Exception {
     System.clearProperty("ssl.protocols");
   }
 
+  @Test
+  public void testSSLOptionYamlOptionWithProperyFalse() throws Exception {
+    System.setProperty("ssl.authPeer", "false");
+    DynamicConfiguration configFromYamlFile =
```
WillemJiang: The below code is duplicated code, it's better to extract a method for setting up the configuration from system property.

liubao68: For unit test case, I think it quite important to make each test have only one logic (that is to say, code complexity is one) and code duplication is not a concern. So I used to copy code for new test cases.

foundations/foundation-ssl/src/test/java/io/servicecomb/foundation/ssl/SSLOptionTest.java
```diff
@@ -230,6 +230,46 @@ public void testSSLOptionYamlOption2() throws Exception {
...
+    boolean authPeer = option.isAuthPeer();
+    option.setAuthPeer(authPeer);
+    Assert.assertEquals(false, authPeer);
+    System.getProperties().remove("ssl.authPeer");
+  }
+
+  @Test
+  public void testSSLOptionYamlOptionWithProperyTrue() throws Exception {
+    System.setProperty("ssl.authPeer", "true");
+    DynamicConfiguration configFromYamlFile =
```
wujimin: it seems that finalConfig equals io.servicecomb.config.ConfigUtil.createLocalConfig() result?

PR #366, opened 2017-12-07
JAV-557-adjust the DEFAULT_MAX_CONCURRENT_REQUESTS

mt-monster:
adjust the DEFAULT_MAX_CONCURRENT_REQUESTS

### DIFF
--- a/handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/Configuration.java
+++ b/handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/Configuration.java
@@ -67,7 +67,7 @@ public final class Configuration {
 
   private static final int DEFAULT_ISOLATION_TIMEOUT = 30000;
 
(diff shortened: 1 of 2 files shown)


handlers/handler-bizkeeper/src/main/java/io/servicecomb/bizkeeper/Configuration.java
```diff
@@ -114,7 +114,7 @@ public boolean getIsolationTimeoutEnabled(String type, String microserviceName,
   public int getIsolationMaxConcurrentRequests(String type, String microserviceName,
       String qualifiedOperationName) {
     int concurrentRequests;
-    String p = getProperty("10",
+    String p = getProperty("1000",
```
wujimin: all configuration items convert to string, and then convert to int/boolean and so on
and this convert happened in every request, it's not good
can we create a new issue to optimize this?

WillemJiang: Please create JIRA for it.

PR #376, opened 2017-12-08
[JAVA-961] Dynamic configuration is not valid at some time

laijianbin:

---
When delete  the dynamic configuration  through Config Center, then ServiceComb will use  RoundRobin as the default ,after a while, sending the same configuration i hava deleted ,the configuration will not valid

### DIFF
--- a/handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
+++ b/handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
@@ -92,8 +92,8 @@ public class LoadbalanceHandler implements Handler {
     String p = Configuration.INSTANCE.getPolicy(invocation.getMicroserviceName());
     String strategy = Configuration.INSTANCE.getRuleStrategyName(invocation.getMicroserviceName());
 
-    if (this.policy != null && !this.policy.equals(p)
-        || (this.strategy != null && !this.strategy.equals(strategy))) {
+    if (!isNotChanged(p, strategy)) {
+
       //配置变化，需要重新生成所有的lb实例
       synchronized (lock) {
         loadBalancerMap.clear();
@@ -348,6 +348,18 @@ public class LoadbalanceHandler implements Handler {
       } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
         LOGGER.warn("Unable to load filter class: " + className);
       }
+    }
+  }
+
+  private boolean isNotChanged(String p, String s) {
+    if (p == null && s == null) {
+      return this.policy == null && this.strategy == null;
+    } else if (p == null) {
+      return this.policy == null && s.equals(this.strategy);
+    } else if (s == null) {
+      return this.strategy == null && p.equals(policy);
+    } else {
+      return p.equals(policy) && s.contentEquals(strategy);
     }
   }
 }


handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -350,4 +350,16 @@ private void loadFilter(String filter, LoadBalancer lb) {
       }
     }
   }
+
+  private boolean isNotChanged(String p, String s) {
```
wujimin: use meaning name

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -350,4 +350,16 @@ private void loadFilter(String filter, LoadBalancer lb) {
...
+    if (p == null && s == null) {
+      return this.policy == null && this.strategy == null;
+    } else if (p == null) {
```
wujimin: already return, no need else

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
wujimin: maybe "isChanged" is better

wujimin: UT

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -350,4 +350,16 @@ private void loadFilter(String filter, LoadBalancer lb) {
...
+      return this.policy == null && s.equals(this.strategy);
+    } else if (s == null) {
+      return this.strategy == null && p.equals(policy);
+    } else {
+      return p.equals(policy) && s.contentEquals(strategy);
```
wujimin: why strategy need special contentEquals?

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
wujimin: so complex?
it seems old logic is simpler.

WillemJiang:
@laijianbin The JIRA number is not write, could you double check it?

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
eric-lee-ltk: maybe just implement a compare method like the below:
```
private static boolean isEqual(String str1, String str2) {
    return (str1 == null ? str2 == null : str1.equals(str2));
}
```
Then we can simply call `isEqual(policy, this.policy) && isEqual(strategy, this.strategy)` to get the result without so many if/else.

laijianbin: @eric-lee-ltk  This method（isNotChanged()） only use for a judgement . i think there will not much change.so i think no need.

laijianbin: @wujimin  oh! i make a mistake

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
laijianbin: Yeah, the old logic is simpler. But after delete  configuration ,ServiceComb uses RoundRobin Rule as default ,then i  push configuration like RandomRule .ServiceComb still uses the  default configuration.

laijianbin: Okay, i will try

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
eric-lee-ltk: @laijianbin What you intend to do in the `isNotChanged` looks like just compare whether two variables are equal or not. You implementation looks too complicated!

handlers/handler-loadbalance/src/main/java/io/servicecomb/loadbalance/LoadbalanceHandler.java
wujimin: yes, old logic is simpler but had some problems
but new logic is too confusion

i think eric-lee-ltk's suggestion is good.

laijianbin: oh, i think use "isChanged" will more complex

laijianbin:
I pull a new request ,so i close this PR

PR #393, opened 2017-12-10
Use StringBuilder instead of String in loops for concatenation

### LINKED ISSUE #391: Better use StringBuffer append instead of String concatenation in loops.
As every String concatenation copies the whole String, usually it is preferable to replace it with explicit calls to StringBuilder.append() or StringBuffer.append() in loops.

yz19910419:

---
fix #391

### DIFF
--- a/common/common-rest/src/main/java/io/servicecomb/common/rest/definition/path/URLPathBuilder.java
+++ b/common/common-rest/src/main/java/io/servicecomb/common/rest/definition/path/URLPathBuilder.java
@@ -66,26 +66,26 @@ public class URLPathBuilder {
       rawPath = SLASH + rawPath;
     }
 
-    String tmpPath = "";
+    StringBuilder tmpPath = new StringBuilder();
     for (int idx = 0; idx < rawPath.length(); idx++) {
       char currentChar = rawPath.charAt(idx);
       if (currentChar == '{') {
         if (tmpPath.length() != 0) {
-          this.pathParamWriterList.add(new StaticUrlParamWriter(tmpPath));
-          tmpPath = "";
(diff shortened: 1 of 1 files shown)


common/common-rest/src/main/java/io/servicecomb/common/rest/definition/path/URLPathBuilder.java
```diff
@@ -66,26 +66,26 @@ private void initPathWriterList(String rawPath, Map<String, RestParam> paramMap)
...
-    String tmpPath = "";
+    StringBuilder tmpPath = new StringBuilder();
     for (int idx = 0; idx < rawPath.length(); idx++) {
       char currentChar = rawPath.charAt(idx);
       if (currentChar == '{') {
         if (tmpPath.length() != 0) {
-          this.pathParamWriterList.add(new StaticUrlParamWriter(tmpPath));
-          tmpPath = "";
+          this.pathParamWriterList.add(new StaticUrlParamWriter(tmpPath.toString()));
+          tmpPath = tmpPath.delete(0, tmpPath.length());
```
wujimin: tmpPath.setLength(0) is simpler.

wujimin:
the optimization is OK, but not absolutely necessary, :)
because only invoked during boot, not performance bottlenecks.

common/common-rest/src/main/java/io/servicecomb/common/rest/definition/path/URLPathBuilder.java
yz19910419: thanks @wujimin , that's a brilliant idea.

PR #394, opened 2017-12-10
remove useless Timer

### LINKED ISSUE #387: extract common implementation of getNextTimer and nextTimer
There are some common implementation logic  of `getNextTimeout` and `nextTimeout` in `Timer`. We should extract a common method to process this logic instead of implement it twice in these methods.

eekelvin:

---
fix #387


service-registry/src/main/java/io/servicecomb/serviceregistry/utils/Timer.java
```diff
@@ -79,17 +79,18 @@ public int getCurrent() {
   }
 
   public int getNextTimeout() {
```
wujimin: Timer seems already useless, we can remove it.

PR #397, opened 2017-12-10
[JAV-558] pullTask,use rev to ensure data consistency and reduce data…

weichao666:
Not only InstanceCacheManagerOld, but also InstanceCacheManagerNew was modified.
If the rev of SDK is greater than the rev of SC, it means that the result of SDK is new to SC, SC doesn't need to be processed, and returns 304 to indicate that SDK's cache is not refreshed.
If the rev of SDK is less than SC's rev, SC returns the updated result set to SDK, SDK refreshes the cache and
updates the Rev.

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/cache/InstanceCacheManagerOld.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/cache/InstanceCacheManagerOld.java
@@ -33,7 +33,9 @@ import io.servicecomb.serviceregistry.ServiceRegistry;
 import io.servicecomb.serviceregistry.api.Const;
 import io.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
 import io.servicecomb.serviceregistry.api.response.MicroserviceInstanceChangedEvent;
+import io.servicecomb.serviceregistry.client.http.MicroserviceInstanceRefresh;
 import io.servicecomb.serviceregistry.config.ServiceRegistryConfig;
+import io.servicecomb.serviceregistry.definition.DefinitionConst;
 import io.servicecomb.serviceregistry.task.InstancePullTask;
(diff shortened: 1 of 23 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/cache/InstanceCacheManagerOld.java
```diff
@@ -89,19 +93,32 @@ private InstanceCache create(String appId, String microserviceName, String micro
...
+        serviceRegistry.findServiceInstance(appId, microserviceName, microserviceVersionRule, instanceCache.getRevision());
+    if (microerviceInstanceRefresh == null) {
       return null;
     }
+    if (!microerviceInstanceRefresh.isNeedRefresh()) {
+      return instanceCache;
+    }
+    String revision = microerviceInstanceRefresh.getRevision();
+    String rev = instanceCache.getRevision();
+    if (Integer.valueOf(rev) - Integer.valueOf(revision) >= 0) {
```
wujimin: SC said that rev is a String
is it safe to convert to int?

wWX495170: SC said the rev will get bigger and bigger,so told us saved to String will be better

wujimin: why bigger has relationship with String?

weichao666: let's discuss with SC to make sure

weichao666: still use String

PR #406, opened 2017-12-11
[JAV-566] report type/version when register to SC

weichao666:
1. Sdk report their framework type and version when register to SC
2. Each service with specific version has a certain framework+version
3. Present the type/version information in console in Service+Version level
4. In gov service, send the different command to CC according to the type/version when doing management

### DIFF
--- a/common/common-protobuf/src/main/java/io/servicecomb/codec/protobuf/jackson/ParamSerializer.java
+++ b/common/common-protobuf/src/main/java/io/servicecomb/codec/protobuf/jackson/ParamSerializer.java
@@ -35,8 +35,8 @@ public class ParamSerializer extends JsonSerializer<Object> {
     ProtobufGenerator protobufGenerator = (ProtobufGenerator) gen;
     Iterator<ProtobufField> iter = protobufGenerator.getSchema().getRootType().fields().iterator();
     Object[] values = (Object[]) value;
-    for (int idx = 0; idx < values.length; idx++) {
-      gen.writeObjectField(iter.next().name, values[idx]);
+    for (Object value1 : values) {
+      gen.writeObjectField(iter.next().name, value1);
     }
 
     gen.writeEndObject();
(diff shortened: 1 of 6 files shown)


foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/base/ServiceCombConstants.java
```diff
@@ -49,4 +49,12 @@
...
 
   String CONFIG_KEY_SPLITER = "_";
+
+  String CONFIG_MICROSERVICE_DEVELOPMENT_FRAMEWORK_KEY = "framework.name";
+
+  String DEFAULT_MICROSERVICE_DEVELOPMENT_FRAMEWORK = "UNKNOWN";
+
+  String CONFIG_MICROSERVCIE_DEVELOPMENT_FRAMEWORK_VERSION_KEY = "framework.version";
+
+  String CONFIG_MICROSERVICE_REGISTER_WAY_KEY = "registreBy";
```
WillemJiang: It should be registeredBy

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
```diff
@@ -33,6 +33,10 @@
 public class Microservice {
   private String serviceId;
 
+  private Map<String, String> framework = new HashMap<>();
+  
+  private String registerBy;
```
WillemJiang: same typo here.

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
```diff
@@ -78,6 +82,22 @@ public void setServiceId(String serviceId) {
...
 
+  public Map<String, String> getFramework() {
+    return framework;
+  }
+
+  public void setFramework(Map<String, String> framework) {
+    this.framework = framework;
+  }
+
+  public String getRegisterBy() {
```
WillemJiang: typo

foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/base/ServiceCombConstants.java
weichao666: Sorry, the spelling is wrong here,it has been changed to registerBy,it's sure better to use registeredBy, but this field is provided by SC, i have to align with them...

PR #431, opened 2017-12-13
JAV-570 connect to servicecenter by config proxy

jeho0815:
connect to servicecenter by config proxy


service-registry/src/main/java/io/servicecomb/serviceregistry/config/ServiceRegistryConfig.java
```diff
@@ -244,6 +257,31 @@ public String getSecretKey() {
     return tenantName;
   }
 
+  public Boolean isProxyEnable() {
+    String enable = getProperty("false", PROXY_ENABLE);
+    return Boolean.parseBoolean(enable);
+  }
+
+  public String getProxyHost() {
+    String host = getProperty("127.0.0.1", PROXY_HOST);
```
seanyinx: how about `return getProperty("127.0.0.1", PROXY_HOST);` directly?

same for other properties

PR #433, opened 2017-12-13
[JAV-571] Produce mediatype determine

wujimin:
fix bug:
declare only produce text, and consume by json,javascript,*.*,  but produced json

### DIFF
--- a/common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
+++ b/common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
@@ -216,7 +216,7 @@ public abstract class AbstractRestInvocation {
       }
     }
     responseEx.setStatus(response.getStatusCode(), response.getReasonPhrase());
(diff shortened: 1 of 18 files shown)


common/common-rest/src/main/java/io/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -216,7 +216,7 @@ protected void sendResponse(Response response) throws Exception {
       }
     }
     responseEx.setStatus(response.getStatusCode(), response.getReasonPhrase());
-    responseEx.setContentType(produceProcessor.getName());
+    responseEx.setContentType(produceProcessor.getName()+"; charset=utf-8");
```
WillemJiang: It's better to get the charset from context or other setting to let the user override it.

wujimin: yes
but too many place need to be changed, so it will be a new issue, not in this PR.

PR #439, opened 2017-12-15
[JAV-576] The service instance state can‘t be invoked by the client when it is DOWN

weichao666:
sdk在push和pull时需要作判断,sdk不访问实例状态为DOWN的实例


service-registry/src/main/java/io/servicecomb/serviceregistry/cache/InstanceCacheManagerOld.java
```diff
@@ -149,6 +150,11 @@ public void onInstanceUpdate(MicroserviceInstanceChangedEvent changedEvent) {
       switch (changedEvent.getAction()) {
         case CREATE:
         case UPDATE:
+          if (changedEvent.getInstance().getStatus() != MicroserviceInstanceStatus.UP) {
+            instMap.remove(changedEvent.getInstance().getInstanceId());
+            cacheMap.put(key, new InstanceCache(appId, microserviceName, version, instMap));
+            break;
+          }
           instMap.put(changedEvent.getInstance().getInstanceId(), changedEvent.getInstance());
```
eric-lee-ltk: how about putting this line in the else clause to avoid duplicate statement of line 155 and 159?

weichao666: done

liubao68:
Please check https://github.com/ServiceComb/ServiceComb-Java-Chassis/pull/449 for more details

PR #440, opened 2017-12-15
[JAV-566] report type/version when register to SC

weichao666:
1. SDK report their framework type and version when register to SC
2. Each service with specific version has a certain framework+version
3. Present the type/version information in console in Service+Version level
4. In gov service, send the different command to CC according to the type/version when doing management

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
@@ -23,6 +23,7 @@ import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.servicecomb.serviceregistry.api.Const;
 
@@ -32,6 +33,11 @@ import io.servicecomb.serviceregistry.api.Const;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Microservice {
   private String serviceId;
+
+  private Map<String, String> framework = new HashMap<>();
+
+  @JsonProperty(value = "registerBy")
+  private String registeredBy;
 
   private String appId;
 
@@ -172,4 +178,21 @@ public class Microservice {
   public void setPaths(List<BasePath> paths) {
     this.paths = paths;
   }
+
+  public Map<String, String> getFramework() {
+    return framework;
+  }
+
+  public void setFramework(Map<String, String> framework) {
+    this.framework = framework;
+  }
+
+  public String getRegisteredBy() {
+    return registeredBy;
+  }
+
+  public void setRegisteredBy(String registeredBy) {
+    this.registeredBy = registeredBy;
+  }
+
 }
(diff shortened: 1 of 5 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
```diff
@@ -33,6 +34,11 @@
 public class Microservice {
   private String serviceId;
 
+  private Map<String, String> framework = new HashMap<>();
+
+  @JsonProperty(value = "registerBy")
+  private String registeredBy;
```
eric-lee-ltk: Why the key of the property differs from the variable name?

weichao666: ServiceCenter provider "registerBy" to us, sdk use the "registeredBy" to fit the semantics more

eric-lee-ltk: That makes sense. Thanks for the explanation.

wujimin: means "owner"?
if possible, maybe SC also make the change.

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
```diff
@@ -33,6 +34,11 @@
 public class Microservice {
   private String serviceId;
 
+  private Map<String, String> framework = new HashMap<>();
```
wujimin: why a map, why not a model?

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
```diff
@@ -64,6 +69,16 @@ private Microservice createMicroserviceFromDefinition(Configuration configuratio
       microservice.setAlias(Microservice.generateAbsoluteMicroserviceName(microservice.getAppId(),
           microservice.getServiceName()));
     }
+    
+    Map<String, String> framework = new HashMap<>();
+    framework.put("name", configuration.getString(CONFIG_MICROSERVICE_DEVELOPMENT_FRAMEWORK_KEY, DEFAULT_MICROSERVICE_DEVELOPMENT_FRAMEWORK));
```
wujimin: default to SDK?

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
```diff
@@ -64,6 +69,16 @@ private Microservice createMicroserviceFromDefinition(Configuration configuratio
...
+    framework.put("version", configuration.getString(CONFIG_MICROSERVICE_DEVELOPMENT_FRAMEWORK_VERSION_KEY, ""));
```
wujimin: default to version in pom?

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
```diff
@@ -64,6 +69,16 @@ private Microservice createMicroserviceFromDefinition(Configuration configuratio
...
+    microservice.setFramework(framework);
+    String str = "";
+    if (configuration.getString(CONFIG_MICROSERVICE_REGISTER_WAY_KEY, "") != null) {
```
wujimin: we can get a null value in this place?

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/Microservice.java
wujimin: can not understand.
Microservice class can express self and others
so when your said "registerBy" or "registeredBy", there are all correct and all wrong.

service-registry/src/main/java/io/servicecomb/serviceregistry/api/registry/MicroserviceRegisterWay.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.servicecomb.serviceregistry.api.registry;
+
+/**
+ * Created by   on 2017/12/15.
+ */
+public enum MicroserviceRegisterWay {
```
wujimin: just a string, why we need a enum?
some day, add a type, all sdk must update, otherwise will got a unknown?

weichao666: SC's data dictionary shows it's a enum type

wujimin:
we have servicecomb and cse version, which one  will  we report

PR #453, opened 2017-12-21
 [JAV-508] & [SCB-11] Metrics Auto Publish

zhengyangyong:

 - [x] Make sure there is a [JIRA issue](https://https://servicecomb.atlassian.net/projects/JAV/issues) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
 - [x] Format the pull request title like `[JAV-XXX] Fixes bug in ApproximateQuantiles`, where you replace `JAV-XXX` with the appropriate JIRA issue.

---
JAV-508 & SCB-11 Metrics Auto Publish

If user include metrics in dependency,the metrics service would auto publish with Java Chassis Producer,also use same transport and endpoint config.
For example if user config rest address 0.0.0.0:8080 in microservice.yaml,then the metrics service will available at http://<ip address>:8080/metrics


metrics/metrics-core/src/main/java/io/servicecomb/metrics/core/publish/JsonMetricsPublisher.java
```diff
@@ -0,0 +1,48 @@
...
+public class JsonMetricsPublisher implements MetricsPublisher {
+
+  private final DataSource dataSource;
+
+  public JsonMetricsPublisher(DataSource dataSource) {
+    this.dataSource = dataSource;
+  }
+
+  @Override
+  public String metrics(int pollerIndex) {
```
WillemJiang: It‘s not a good way to just return the string here.

zhengyangyong: I had post a mail discuss it : https://lists.apache.org/thread.html/8c34a561c0c4e34b757e32e58c5a99e205c157326292d479526da4c3@%3Cdev.servicecomb.apache.org%3E

zhengyangyong:
this will be improve after this PR : https://github.com/apache/incubator-servicecomb-java-chassis/pull/451,we had fully discuss via mail

PR #461, opened 2017-12-23
[JAV-591] Completable future response mapper

wujimin:


swagger/swagger-invocation/invocation-core/src/main/java/io/servicecomb/swagger/engine/SwaggerEnvironment.java
```diff
@@ -45,14 +47,20 @@
   @Inject
   private ProducerArgumentsMapperFactory producerArgumentsFactory;
 
-  @Inject
-  private ProducerResponseMapperFactory producerResponseMapperFactory;
+  private ResponseMapperFactorys<ProducerResponseMapper> producerResponseMapperFactorys =
```
WillemJiang: why do we make this kind of change?

wujimin: 1.old mechanism only support one level mapper, like:
`ResponseEntity<String> -> String -> ResponseEntity<String>`

but after we support CompletableFuture, we need multiple level mapper, like:
`CompletableFuture<ResponseEntity<String>> -> String -> CompletableFuture<ResponseEntity<String>>`

2.old mechanism is complex and not flexible, hard to extend

swagger/swagger-invocation/invocation-springmvc/src/main/java/io/servicecomb/swagger/invocation/springmvc/response/SpringmvcProducerResponseMapper.java
```diff
@@ -44,9 +43,16 @@ public Response mapResponse(StatusType status, Object response) {
...
         springmvcResponse.getStatusCode().getReasonPhrase());
-    Response cseResponse = Response.status(responseStatus).entity(springmvcResponse.getBody());
-    HttpHeaders headers = springmvcResponse.getHeaders();
 
+    Response cseResponse = null;
+    if (HttpStatus.isSuccess(responseStatus)) {
+      cseResponse = realMapper.mapResponse(responseStatus, springmvcResponse.getBody());
+    } else {
+      // not support fail response mapper now
+      cseResponse = Response.status(responseStatus).entity(springmvcResponse.getBody());
```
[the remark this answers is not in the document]
wujimin: now, all failed response depend on developers create InvocationException manually in producer impl class or handler
no need to do it in mapper

PR #464, opened 2017-12-26
[JAV-583] Pojo consumer return completable future

wujimin:


providers/provider-pojo/src/main/java/io/servicecomb/provider/pojo/Invoker.java
```diff
@@ -86,20 +87,48 @@ protected void prepare() {
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
     if (swaggerConsumer == null) {
-      prepare();
+      synchronized (this) {
+        if (swaggerConsumer == null) {
```
WillemJiang: swaggerConsumer should be volatile.

wujimin: done

WillemJiang: @wujimin  did you submit the change, I cannot find the change.

wujimin: maybe you did not refresh the page?

PR #470, opened 2017-12-27
[JAV-592] Producer return completable future

wujimin:


demo/demo-pojo/pojo-client/src/main/java/io/servicecomb/demo/pojo/client/CodeFirstPojoClient.java
```diff
@@ -152,7 +152,11 @@ protected void testCodeFirstIsTrue(CodeFirstPojoIntf codeFirst) {
   }
 
   protected void testCodeFirstSayHi2(CodeFirstPojoIntf codeFirst) {
-    String result = codeFirst.sayHi2("world");
+    if (!CodeFirstPojoClientIntf.class.isInstance(codeFirst)) {
```
WillemJiang: There is a  build error in the CI.

wujimin: done

PR #471, opened 2017-12-27
[JAV-575] Exception to response converter

wujimin:


swagger/swagger-invocation/invocation-core/src/main/java/io/servicecomb/swagger/invocation/exception/ExceptionToResponseConverters.java
```diff
@@ -0,0 +1,45 @@
...
+  @SuppressWarnings("unchecked")
+  public ExceptionToResponseConverters() {
+    SPIServiceUtils.getAllService(ExceptionToResponseConverter.class).forEach(converter -> {
+      exceptionToResponseConverters.put(converter.getExceptionClass(), converter);
+    });
+  }
+
+  public Response convertExceptionToResponse(SwaggerInvocation swaggerInvocation, Throwable e) {
+    ExceptionToResponseConverter<Throwable> converter = exceptionToResponseConverters.get(e.getClass());
+    if (converter == null) {
```
WillemJiang: Do we have default converter to handler the exception?

swagger/swagger-invocation/invocation-core/src/test/java/io/servicecomb/swagger/engine/TestSwaggerProducerOperation.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class TestSwaggerProducerOperation {
+  SwaggerProducerOperation swaggerProducerOperation = new SwaggerProducerOperation();
+
+  @Test
+  public void processException_convert(@Mocked SwaggerInvocation invocation) {
+    Error error = new Error("abc");
+
+    Response response = swaggerProducerOperation.processException(invocation, error);
+    Assert.assertSame(Status.OK, response.getStatus());
```
WillemJiang: If there is an exception, do we need to set the status to be OK?

wujimin: this is just a test demo.
converter will not limit the convert logic, some customers(eg: music) will convert a exception to normal response, and error information be setted to body model.

swagger/swagger-invocation/invocation-core/src/main/java/io/servicecomb/swagger/invocation/exception/ExceptionToResponseConverters.java
wujimin: default logic is in swaggerProducerOperation
that's not so good, i will extract it to be a default converter

wujimin: done

PR #472, opened 2017-12-29
[JAV-558] pullTask,use rev to ensure data consistency and reduce service center work load

weichao666:
…ice center work load

---
If the rev of SDK is greater than or equal to the rev of SC, it means that  SDK no need to refresh, SC doesn't need to be processed, and returns 304 to indicate that SDK's cache is not refreshed.
If the rev of SDK is less than SC's rev, SC returns the updated result set to SDK, SDK refreshes the cache and updates the Rev.

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/RegistryUtils.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/RegistryUtils.java
@@ -20,7 +20,6 @@ package io.servicecomb.serviceregistry;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.List;
 import java.util.Map;
 
 import org.apache.http.client.utils.URIBuilder;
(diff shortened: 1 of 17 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -202,9 +202,9 @@ private static IpPort genPublishIpPort(String schema, IpPort ipPort) {
     return new IpPort(publicAddressSetting, publishPort);
   }
 
-  public static List<MicroserviceInstance> findServiceInstance(String appId, String serviceName,
-      String versionRule) {
-    return serviceRegistry.findServiceInstance(appId, serviceName, versionRule);
+  public static MicroserviceInstances findServiceInstance(String appId, String serviceName,
```
wujimin: cause not compatible?

weichao666:
create new methods，not modify the exist methods

PR #479, opened 2018-01-03
[SCB-209] pullTask,use rev to ensure data consistency and reduce service center work load

weichao666:
…ice center work load

---
Not modify the existed methods,add the new methods.
If the rev of SDK is greater than or equal to the rev of SC, it means that SDK no need to refresh, SC doesn't need to be processed, and returns 304 to indicate that SDK's cache is not refreshed.
If the rev of SDK is less than SC's rev, SC returns the updated result set to SDK, SDK refreshes the cache and updates the Rev.


service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -100,17 +111,23 @@ public void pullInstances() {
...
+    MicroserviceInstances microserviceInstances = RegistryUtils.findServiceInstances(appId,
         microserviceName,
-        DefinitionConst.VERSION_RULE_ALL);
-    if (pulledInstances == null) {
+        DefinitionConst.VERSION_RULE_ALL,
+        revision);
+    if (microserviceInstances == null) {
+      return;
+    }
+    if (!microserviceInstances.isNeedRefresh()) {
```
liubao68: There are conflicts with https://github.com/apache/incubator-servicecomb-java-chassis/pull/486

service-registry/src/main/java/io/servicecomb/serviceregistry/client/LocalServiceRegistryClientImpl.java
```diff
@@ -267,6 +270,43 @@ protected Microservice findLatest(String appId, String serviceName, VersionRule
     return allInstances;
   }
 
+  @Override
+  public MicroserviceInstances findServiceInstances(String selfMicroserviceId, String appId, String serviceName,
```
wujimin: you copy old findServiceInstance, and do some modify
DO NOT do this things.

can extract a method, all invoke the new one
or one invoke another
BUT DO NOT COPY CODE

wujimin: also local registry must impl revision logic
otherwise will print too many logs in pull mode.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -549,6 +579,44 @@ public void watch(String selfMicroserviceId, AsyncResultCallback<MicroserviceIns
     return null;
   }
 
+  @Override
+  public MicroserviceInstances findServiceInstances(String consumerId, String appId, String serviceName,
```
wujimin: DO NOT COPY CODE.

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -45,6 +46,8 @@
 
   private String microserviceName;
 
+  private String revision = DefinitionConst.DEFAULT_REVISION;
```
wujimin: default must be null
if did not get revison from SC, then in request header must not exist revision key.

service-registry/src/main/java/io/servicecomb/serviceregistry/definition/DefinitionConst.java
```diff
@@ -34,4 +34,6 @@
   String VERSION_RULE_LATEST = "latest";
 
   String VERSION_RULE_ALL = "0.0.0+";
+
+  String DEFAULT_REVISION = "0";
```
wujimin: there is no "DEFAULT_REVISION" logic, no need to define this.

service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
```diff
@@ -225,6 +226,39 @@ public boolean unregisterInstance() {
     return instances;
   }
 
+  public MicroserviceInstances findServiceInstances(String appId, String serviceName,
```
wujimin: io.servicecomb.serviceregistry.registry.AbstractServiceRegistry.findServiceInstance(String, String, String)
can invoke this one, no need to keep two copy.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/LocalServiceRegistryClientImpl.java
weichao666: done

service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
weichao666: done

service-registry/src/main/java/io/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
weichao666: done

service-registry/src/main/java/io/servicecomb/serviceregistry/definition/DefinitionConst.java
weichao666: done

service-registry/src/main/java/io/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
weichao666: done

PR #484, opened 2018-01-04
[SCB-199] bug fix: consumer upload file, did not set correct boundary c…

wujimin:
consumer upload file, did not set correct boundary content type


seanyinx:
shouldn't we use apache jira to track issues?
https://issues.apache.org/jira

wujimin:
yes
some days ago, i try to use apache jira, but have some problems, and someone tell me use old jira after got new notify.

and after apache jira work normally, old issues in these days need to be copy to apache jira.

PR #485, opened 2018-01-04
[SCB-176] Generate traceId

yhs0092:

---

See details in [https://issues.apache.org/jira/browse/SCB-176](https://issues.apache.org/jira/browse/SCB-176).
- provide a traceId generator
- provide a HttpServerFilter:
  + if header contains traceId, move it into invocation context
  + if invocation context carries traceId, let it go
  + if http header and invocation context don't contain traceId, generate a traceId and set into invocation context
- use "X-B3-TraceId" as the header name of traceId, since this is the default traceId header name of zipkin.


WillemJiang:
Just a quick question, what's the relationship between this change and the old zipkin tracing code.
AFAIK,  ZipkinTracingHandler of Java Chassis can take care the creation of SpanID, and it should take care of the TraceID at the same time.  Can you double check it?

wujimin:
we always generate traceId, trace solution will based on this
please ensure zipkin no problem, then  i will  approve this  PR

WillemJiang:
@seanyinx Could you take a look at this PR?

yhs0092:
I've taken an exam on the situation when users enable our ZipkinTracingHandler to use open zipkin. 
In current solution, the traceId generated by TracingFilter will be ignored by the inner logic of Brave, and a new span will be generated. The traceId transported in the invocation chain is the one that Brave generates. And the function of zipkin tracing works normally.

PR #486, opened 2018-01-04
[SCB-165] Service registry not given proper instances

liubao68:

---

### DIFF
--- a/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/IpPort.java
+++ b/foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/IpPort.java
@@ -53,6 +53,28 @@ public class IpPort {
     this.port = port;
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    IpPort ipPort = (IpPort) o;
+
+    if (port != ipPort.port) {
+      return false;
+    }
+    return hostOrIp.equals(ipPort.hostOrIp);
+  }
+
+  @Override
+  public String toString() {
+    return hostOrIp + ":" + port;
+  }
+
   public InetSocketAddress getSocketAddress() {
     if (socketAddress == null) {
       synchronized (lock) {
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
@@ -75,9 +75,10 @@ public class IpPortManager {
   }
 
   public IpPort getNextAvailableAddress(IpPort failedIpPort) {
-    IpPort current = getAvailableAddress();
+    int currentIndex = currentAvailableIndex.get();
+    IpPort current = getAvailableAddress(currentIndex);
     if (current.equals(failedIpPort)) {
-      currentAvailableIndex.incrementAndGet();
+      currentAvailableIndex.compareAndSet(currentIndex, currentIndex + 1);
       current = getAvailableAddress();
     }
 
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
@@ -74,31 +74,32 @@ public class IpPortManager {
     }
   }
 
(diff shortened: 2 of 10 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
```diff
@@ -46,10 +47,10 @@
 
   private ArrayList<IpPort> defaultIpPort;
 
-  private InstanceCache instanceCache = null;
-
   private AtomicInteger currentAvailableIndex;
 
+  private boolean autoDiscoveryInited = false;
```
WillemJiang: The default value is false, we don't need assign the false value here.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
```diff
@@ -65,45 +66,47 @@ public IpPortManager(ServiceRegistryConfig serviceRegistryConfig, InstanceCacheM
...
-    }
-    return result;
+    return new URIEndpointObject(nextEndpoint.getEndpoint());
   }
 
   private List<CacheEndpoint> getDiscoveredIpPort() {
-    if (instanceCache == null) {
-      return new ArrayList<>(0);
+    if (!autoDiscoveryInited || !this.serviceRegistryConfig.isRegistryAutoDiscovery()) {
+      return null;
```
WillemJiang: Is it safe to run null here?  How about return the empty list?

service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -111,7 +111,8 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
               holder.value =
                   JsonUtils.readValue(bodyBuffer.getBytes(), cls);
             } catch (Exception e) {
-              LOGGER.warn(bodyBuffer.toString());
+              LOGGER.warn("read value failed and response message is {}",
```
WillemJiang: Please add the exception into the LOGGER.

liubao68: This is exception is always Json Parse exceptions and no need to print out.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
liubao68: This is private method and it is safe

service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
liubao68: I think it's better to give explicit default value here to false, because it's more readable.

service-registry/src/main/java/io/servicecomb/serviceregistry/client/IpPortManager.java
```diff
@@ -65,45 +66,47 @@ public IpPortManager(ServiceRegistryConfig serviceRegistryConfig, InstanceCacheM
...
-  public IpPort getAvailableAddress(boolean invalidate) {
-    int index;
-    if (invalidate) {
-      index = currentAvailableIndex.incrementAndGet();
-    } else {
-      index = currentAvailableIndex.get();
+  public IpPort getNextAvailableAddress(IpPort failedIpPort) {
+    IpPort current = getAvailableAddress();
+    if (current.equals(failedIpPort)) {
+      currentAvailableIndex.incrementAndGet();
```
wujimin: multiple threads all run with "current.equals(failedIpPort) return true", then cause inc multiple times

so we should record old index when enter "public IpPort getNextAvailableAddress(IpPort failedIpPort)", and then assign currentAvailableIndex = old + 1 or use compareAndSet

foundations/foundation-common/src/main/java/io/servicecomb/foundation/common/net/IpPort.java
```diff
@@ -53,6 +53,28 @@ public void setPort(int port) {
     this.port = port;
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
```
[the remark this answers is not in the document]
liubao68: This is generated code by IDE, and it's correct, so not adjust it now

PR #493, opened 2018-01-08
[SCB-195] Support swagger @Api annotation to specify default tags

yhs0092:

---

See details in [https://issues.apache.org/jira/browse/SCB-195](https://issues.apache.org/jira/browse/SCB-195).
- users can specify default tags to operations by using `@Api`, the default tags can be overwritten by `@ApiOperation` in operation level
- improve empty value filter logic of `@SwaggerDefinition`

### DIFF
--- a/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/OperationGenerator.java
+++ b/swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/OperationGenerator.java
@@ -221,6 +221,24 @@ public class OperationGenerator {
       }
       processor.process(annotation, this);
     }
+
+    setDefaultTag();
+  }
+
+  private void setDefaultTag() {
+    // if tag has been defined, do nothing
+    if (null != operation.getTags()) {
+      for (String tag : operation.getTags()) {
+        if (!StringUtils.isEmpty(tag)) {
+          return;
+        }
+      }
+    }
+
+    // if there is no tag, set default tag
+    if (!swaggerGenerator.getDefaultTags().isEmpty()) {
+      operation.setTags(new ArrayList<>(swaggerGenerator.getDefaultTags()));
+    }
   }
 
   /**
--- a/swagger/swagger-generator/generator-core/src/test/java/io/servicecomb/swagger/generator/core/processor/annotation/ApiOperationProcessorTest.java
+++ b/swagger/swagger-generator/generator-core/src/test/java/io/servicecomb/swagger/generator/core/processor/annotation/ApiOperationProcessorTest.java
@@ -17,8 +17,9 @@
 
 package io.servicecomb.swagger.generator.core.processor.annotation;
 
-import static org.junit.Assert.assertEquals;
+import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
 
 import java.lang.reflect.Method;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


swagger/swagger-generator/generator-core/src/test/java/io/servicecomb/swagger/generator/core/processor/annotation/ApiOperationProcessorTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    Method function = TestClass.class.getMethod("function");
+    SwaggerGenerator swaggerGenerator = new SwaggerGenerator(new PojoSwaggerGeneratorContext(), TestClass.class);
+    OperationGenerator operationGenerator = new OperationGenerator(swaggerGenerator, function);
+
+    apiOperationProcessor.process(function.getAnnotation(ApiOperation.class),
+        operationGenerator);
+
+    List<String> tagList = operationGenerator.getOperation().getTags();
+    assertEquals(2, tagList.size());
```
WillemJiang: You can use org.hamcrest.Matcher.containsInAnyOrder() to check the Tags just like this.

assertThat(operationGenerator.getOperation().getTags(), containsInAnyOrder("tag1", "tag2"));

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/OperationGenerator.java
```diff
@@ -221,6 +221,24 @@ protected void scanMethodAnnotation() {
       }
       processor.process(annotation, this);
     }
+
+    setDefaultTag();
+  }
+
+  private void setDefaultTag() {
+    // if tag has been defined, do nothing
+    if (null != operation.getTags()) {
```
wujimin: never can get a null value?

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/processor/annotation/SwaggerDefinitionProcessor.java
```diff
@@ -74,20 +74,27 @@ private void convertInfo(io.swagger.annotations.Info infoAnnotation, Swagger swa
   private License convertLicense(io.swagger.annotations.License licenseAnnotation) {
     License license = new License();
 
-    license.setName(licenseAnnotation.name());
-    license.setUrl(licenseAnnotation.url());
-
-    return license;
+    license.setName(getStringValue(licenseAnnotation.name()));
+    license.setUrl(getStringValue(licenseAnnotation.url()));
+    if (null == license.getName() && null == license.getUrl()) {
```
WillemJiang: How about the license name is null but the url is not null?
It's better to use StringUtils.isEmpty directly instead of getStringValue.

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/OperationGenerator.java
yhs0092: this modify has been rolled back. users determine what will be printed in these fields by themselves.

swagger/swagger-generator/generator-core/src/main/java/io/servicecomb/swagger/generator/core/processor/annotation/SwaggerDefinitionProcessor.java
yhs0092: This method is for empty value filter purpose. But this is no need to do as wujimin commented. So these have been removed.

swagger/swagger-generator/generator-core/src/test/java/io/servicecomb/swagger/generator/core/processor/annotation/ApiOperationProcessorTest.java
yhs0092: Got it. I have checked the similar situations in my unit tests and changed them into this kind of check method. Thank you.

PR #494, opened 2018-01-08
[SCB-198] fileupload in linux is too slow, use system default setting i…

wujimin:


wujimin:
default is -1, no limit

PR #499, opened 2018-01-10
[SCB-214] AbstractServiceRegistry findServiceInstance process null return value, not print too much exception

weichao666:

---
AbstractServiceRegistry findServiceInstance process null return value, not print too much exception

### DIFF
--- a/service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
+++ b/service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
@@ -177,11 +177,11 @@ public final class ServiceRegistryClientImpl implements ServiceRegistryClient {
                   mInstances.setInstancesResponse(JsonUtils.readValue(bodyBuffer.getBytes(), FindInstancesResponse.class));
                   break;
                 default:
-                  LOGGER.error(bodyBuffer.toString());
+                  LOGGER.warn(bodyBuffer.toString());
                   break;
               }
             } catch (Exception e) {
-              LOGGER.warn(bodyBuffer.toString(), e);
+              LOGGER.warn("read value failed and response message is {}", bodyBuffer.toString());
             }
             countDownLatch.countDown();
           });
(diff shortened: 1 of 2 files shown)


service-registry/src/main/java/io/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -169,10 +169,16 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
           bodyBuffer -> {
             try {
               mInstances.setRevision(response.getHeader("X-Resource-Revision"));
```
liubao68: Please analyse if syncHandler need to modified also.

weichao666: [2018-01-10 12:40:56,382/GMT+08:00][main][INFO]running microservice register task. io.servicecomb.serviceregistry.task.MicroserviceRegisterTask.doRegister(MicroserviceRegisterTask.java:58)
[2018-01-10 12:40:56,413/GMT+08:00][main][INFO]Using the default address resolver as the dns resolver could not be loaded io.vertx.core.spi.resolver.ResolverProvider.factory(ResolverProvider.java:51)
[2018-01-10 12:40:56,632/GMT+08:00][registry-vert.x-eventloop-thread-0][ERROR]{"errorCode":"400012","errorMessage":"Micro-service does not exist","detail":"Service does not exist."}
 io.servicecomb.serviceregistry.client.http.ServiceRegistryClientImpl.lambda$10(ServiceRegistryClientImpl.java:115)
[2018-01-10 12:40:56,679/GMT+08:00][main][INFO]Registry Microservice successfully. id=72ccb9b6f5c011e78b1c286ed4891d48 appId=springmvctest, name=springmvc, version=0.0.3, schemaIds=[controller, metricsEndpoint, codeFirst] io.servicecomb.serviceregistry.task.MicroserviceRegisterTask.doRegister(MicroserviceRegisterTask.java:90)
when 400, syncHandler just print a line code, 
from the result of print exception, it looks like no need to modified.
but if think it from whether read json,can it be modified like below?
              if (response.statusCode() == 200) {
                try {
                  holder.value =
                      JsonUtils.readValue(bodyBuffer.getBytes(), cls);
                } catch (Exception e) {
                  LOGGER.warn("read value failed and response message is {}",
                      bodyBuffer.toString());
                }
              } else {
                LOGGER.warn(bodyBuffer.toString());
              }

liubao68: midify like this is better. but we do not care about it now.

PR #519, opened 2018-01-18
[SCB-256] fix bug: SC 0.5 is not compatible to 0.4, change sdk to suppo…

wujimin:


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -259,7 +275,7 @@ public boolean isSchemaExist(String microserviceId, String schemaId) {
           schemaId,
           e);
     }
-    return holder.value != null;
+    return holder.value != null && schemaId.equals(holder.value.getServiceId());
```
liubao68: getSchemaId?

wujimin: done

PR #522, opened 2018-01-19
[SCB-266] Support generic rest

wujimin:


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -118,7 +118,25 @@ public void pullInstances() {
     List<MicroserviceInstance> pulledInstances = microserviceInstances.getInstancesResponse().getInstances();
     String rev = microserviceInstances.getRevision();
 
-    setInstances(pulledInstances, rev);
+    safeSetInstances(pulledInstances, rev);
+  }
+
+  protected void safeSetInstances(List<MicroserviceInstance> pulledInstances, String rev) {
+    try {
+      setInstances(pulledInstances, rev);
```
liubao68: In which situation will this fail?

wujimin: like dynamic create class failed

wujimin: or failed to request from sc 
if happened, exception eat by scheduled thread pool, we can not know there is a exception.

demo/demo-springmvc/springmvc-server/src/main/java/org/apache/servicecomb/demo/springmvc/server/CodeFirstSpringmvc.java
```diff
@@ -389,6 +390,23 @@ public Object testObject(@RequestBody Object input) {
...
+
+  @PostMapping(path = "/generic1")
+  public Generic<User> testGeneric1(@RequestBody Generic<User> input) {
+    Assert.isInstanceOf(Generic.class, input);
+    Assert.isInstanceOf(User.class, input.value);
+    return input;
+  }
+
+  @PostMapping(path = "/generic2")
+  public Generic<Generic<User>> testGeneric2(@RequestBody Generic<Generic<User>> input) {
```
liubao68: We can add more test cases such as Long, Status(enum),Date types. Because before found in highway, Long is not supported.

wujimin: done
and move generic test case to TestGeneric.java

PR #523, opened 2018-01-22
[SCB-272] Edge find service instances null problem

wujimin:


swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/converter/AbstractConverter.java
```diff
@@ -70,7 +70,10 @@ public JavaType convert(ClassLoader classLoader, String packageName, Swagger swa
 
       return result;
     } catch (IllegalArgumentException e) {
-      LOGGER.info("failed to load generic class, use {}.", result.getGenericSignature(), e);
+      LOGGER.info("failed to load generic class {}, use {}. cause: {}.",
+          rawClassName,
+          result.getGenericSignature(),
+          e.getMessage());
```
WillemJiang: We can see the stack trace when passing the exception instance into LOGGER.  My suggestion is keeping the e there instead of using e.getMessage().

wujimin: this affect in edge , and almost all generic class not exist, so will print too many stacktrace
in this time , we know what happened and why it happened, so only message is useful

PR #527, opened 2018-01-24
[SCB-297] The local cache is not update when the server reboots at th…

laijianbin:
…e situation of accross application

---


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -195,11 +196,13 @@ protected MicroserviceVersionRule createAndInitMicroserviceVersionRule(String st
 
   @Subscribe
   public void onMicroserviceInstanceChanged(MicroserviceInstanceChangedEvent changedEvent) {
-    if (!appId.equals(changedEvent.getKey().getAppId()) ||
-        !microserviceName.equals(changedEvent.getKey().getServiceName())) {
+    boolean isEventAccept = appId.equals(changedEvent.getKey().getAppId()) &&
```
wujimin: wrong logic
do you know what's this means?

wujimin: sorry, logic is right.
but the express is not good.

PR #529, opened 2018-01-25
[SCB-297] The local cache is not update when the server reboots betwe…

laijianbin:
…en different application

---


service-registry/src/test/java/org/apache/servicecomb/serviceregistry/consumer/TestMicroserviceVersions.java
```diff
@@ -292,4 +293,24 @@ void setInstances(List<MicroserviceInstance> pulledInstances, String rev) {
     
     Assert.assertEquals(1, pendingPullCount.get());
   }
+  
+  @Test
+  public void  testIsEventAccept(){
+    MicroserviceKey key = new MicroserviceKey();
```
wujimin: missed not accept case?
if you read the coverage data, you will find the judgment code is yellow, not green.

PR #534, opened 2018-01-26
SCB-293 sleep before quering in tracing test

yangbor:
The zipkin reporter is asynchronous, and there is no gurarantee when
the metrics will be reported to the server. Just sleep for 5 seconds
before quering should be enough for our simple test case scenario.

---


integration-tests/test-common/src/test/java/org/apache/servicecomb/tests/tracing/TracingTestBase.java
```diff
@@ -75,6 +75,15 @@ protected void assertThatSpansReceivedByZipkin(Collection<String> logs, String..
         .map(this::extractIds)
         .collect(Collectors.toList());
 
+    // Sleep for 5 seconds to wait the reporter finish posting to zipkin server.
```
WillemJiang: Using the wait and try look up span id could be a reasonable way for checking the async report result.

PR #551, opened 2018-02-07
[SCB-331] log optimization

laijianbin:

---


liubao68:
Need english speaker to check. I think it's not better than before.

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -151,7 +151,7 @@ private static void duplicateServiceCombConfigToCse(ConcurrentCompositeConfigura
 
   public static DynamicWatchedConfiguration createConfigFromConfigCenter(Configuration localConfiguration) {
     if (localConfiguration.getProperty(configCenterUrlKey) == null) {
-      LOGGER.info("config center URL is missing, skip to load configuration from config center");
+      LOGGER.info("config center URL is not configured");
```
wujimin: 1.maybe print config key is better
2."skip to load configuration from config center", it's better to keep this information

PR #552, opened 2018-02-07
[SCB-294]when send a request to a not exist microservice, edge will always try to find instances of it

liubao68:

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceManager.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceManager.java
@@ -26,8 +26,6 @@ import org.apache.servicecomb.serviceregistry.task.event.RecoveryEvent;
 import com.google.common.eventbus.Subscribe;
 
 public class MicroserviceManager {
-  private static final int MAX_NUM_OF_MICROSERVICES = 1000;
-
   private AppManager appManager;
 
   private String appId;
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceManager.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceManager.java
@@ -17,6 +17,10 @@
 
 package org.apache.servicecomb.serviceregistry.consumer;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
@@ -26,6 +30,8 @@ import org.apache.servicecomb.serviceregistry.task.event.RecoveryEvent;
 import com.google.common.eventbus.Subscribe;
 
 public class MicroserviceManager {
+  private static final int MAX_NUM_OF_MICROSERVICES = 1000;
+
   private AppManager appManager;
 
   private String appId;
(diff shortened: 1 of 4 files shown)


liubao68:
Test cases mentioned in issues have done by hand and can not run automatically now.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -210,9 +220,9 @@ public void onMicroserviceInstanceChanged(MicroserviceInstanceChangedEvent chang
     //   if pull 1/2/3, and then delete 3, but "delete 3" received before pull result, will have wrong 3.
     // EXPIRE::
     //   black/white config in SC changed, we must refresh all data from sc.
-    postPullInstanceEvent(TimeUnit.MILLISECONDS.toMillis(1));
+    postPullInstanceEvent(0);
```
WillemJiang: Why do you change the time to zero.

liubao68: This is an event fired by SC of instance change and we need to get the changed instance immediately. Some user applications based on this event to make calls, and if we pending to get instance, will cause their logic more complex

wujimin: even time is 0, it's a async task.
everything not changed, except time gap is smaller, but time gap is not 0.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -76,6 +77,18 @@ public MicroserviceVersions(AppManager appManager, String appId, String microser
     appManager.getEventBus().register(this);
   }
 
+  public void updateLastAccessTime() {
+    this.lastAccessedTime = System.currentTimeMillis();
+  }
+
+  public boolean isValidated() {
```
wujimin: when will a MicroserviceVersions created, but never used?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceManager.java
```diff
@@ -41,11 +47,31 @@ public MicroserviceManager(AppManager appManager, String appId) {
...
       instance.submitPull();
       return instance;
     });
+    if (!microserviceVersions.isValidated()) {
+      // remove this microservice if it does not exist or not registered in order to get it back when access it again
+      versionsByName.remove(microserviceName);
+    } else {
+      microserviceVersions.updateLastAccessTime();
+    }
+    if (versionsByName.size() >= MAX_NUM_OF_MICROSERVICES) {
```
wujimin: i'm not sure if it is right.
think about edge, if realy need route to so many targets, this logic will cause problem.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -109,8 +122,6 @@ public void pullInstances() {
         DefinitionConst.VERSION_RULE_ALL,
         revision);
     if (microserviceInstances == null) {
```
wujimin: pull failed or setInstance failed, will never try again in watch mode except notified by another event.
i remember it's a bug in old version
now, it will happened too?

liubao68: in dynamic schema loading, when server not started, pull fail due to service not exists, at this time this not notification will given by sc and we will fetch create MicroserviceVersions again when call the method. check logic in MicroserviceVersions.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
liubao68: when microserivce not exists or for the first time some unexpected sc problems happens

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
liubao68: but it's better to use 0 than 1, right?

wujimin: ok

PR #565, opened 2018-03-02
[SCB-355]If auth failed,sc/cc will not connect server to reduce server pressure

jeho0815:

---


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
```diff
@@ -56,6 +56,11 @@ public void onMicroserviceRegisterTask(MicroserviceRegisterTask task) {
 
   @Override
   protected boolean doRegister() {
+    if (microserviceInstance.getServiceId() == null) {
```
wujimin: why register microservice failed, but try to register microservice instance?

PR #567, opened 2018-03-02
[SCB-365]when configured ssl certificates not exists, startup will hang and no error messages

liubao68:

---

### DIFF
--- a/foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/client/ClientVerticle.java
+++ b/foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/client/ClientVerticle.java
@@ -17,15 +17,25 @@
 
 package org.apache.servicecomb.foundation.vertx.client;
 
+import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


wujimin:
maybe there is something not confirmed.
in my test, there is no problem.  throw exception directly or throw from future, both got the same result.
so maybe this is not the root cause.

image

liubao68:
Please see issues description. In your demo, should include HttpServer or HttpClient setUp， otherwise no CloseHook is registered and vert.x works fine.  Or you can run SpringMVCServer deom, and configure make DemoSSLScutom to return wrong path and this will happen.

wujimin:
submit a vertx issue, waiting for response.

liubao68:
@wujimin can you provide issue link? And I think this modification have no side effect and can be merged before vert.x fixed the problem.

wujimin:
https://github.com/eclipse/vert.x/issues/2331

already closed, vertx team did not know what happened for this issue, still try to discuss

liubao68:
It seems vert.x is not going to handle this, can we merge this PR to fix the problem ?

PR #569, opened 2018-03-05
SCB-210 remove extra useless config item

lijasonvip:
Signed-off-by: lijasonvip <libo75@huawei.com>

---
Use apollo.config.serverUri to note address of Apollo configuration.
Remove cse.config.client.serverUri which is useless.

### DIFF
--- a/dynamic-config/config-apollo/src/main/java/org/apache/servicecomb/config/archaius/sources/ApolloConfigurationSourceImpl.java
+++ b/dynamic-config/config-apollo/src/main/java/org/apache/servicecomb/config/archaius/sources/ApolloConfigurationSourceImpl.java
@@ -49,10 +49,20 @@ public class ApolloConfigurationSourceImpl implements ConfigCenterConfigurationS
 
   private List<WatchedUpdateListener> listeners = new CopyOnWriteArrayList<>();
 
+  private static final String APOLLO_CONFIG_URL_KEY = "apollo.config.serverUri";
+
   public ApolloConfigurationSourceImpl() {
   }
 
   private final UpdateHandler updateHandler = new UpdateHandler();
+
+  @Override
+  public boolean isValidSource(Configuration localConfiguration) {
+    if (localConfiguration.getProperty(APOLLO_CONFIG_URL_KEY) == null){
+      return false;
+    }
+    return true;
+  }
 
   @Override
   public void init(Configuration localConfiguration) {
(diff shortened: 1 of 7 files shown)


foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -150,7 +151,7 @@ private static void duplicateServiceCombConfigToCse(ConcurrentCompositeConfigura
   }
 
   public static DynamicWatchedConfiguration createConfigFromConfigCenter(Configuration localConfiguration) {
-    if (localConfiguration.getProperty(configCenterUrlKey) == null) {
+    if ((localConfiguration.getProperty(configCenterUrlKey) == null) && (localConfiguration.getProperties((configApolloUrlKey)) == null )) {
```
wujimin: each time, we add a new IMPL.
we need to modify this place.
it's very bad.

lijasonvip: OK, lets remove this check here and do the check after we get the configCenterSource.

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/spi/ConfigCenterConfigurationSource.java
```diff
@@ -22,5 +22,6 @@
 import com.netflix.config.WatchedConfigurationSource;
 
 public interface ConfigCenterConfigurationSource extends WatchedConfigurationSource {
+  boolean isValidSource(Configuration localConfiguration);
```
liubao68: For compatibility, this method should give default implementation. But it's fine now , this is not known implementations.

dynamic-config/config-apollo/src/main/java/org/apache/servicecomb/config/archaius/sources/ApolloConfigurationSourceImpl.java
```diff
@@ -49,11 +49,21 @@
...
+  private static final String APOLLO_CONFIG_URL_KEY = "apollo.config.serverUri";
+
   public ApolloConfigurationSourceImpl() {
   }
 
   private final UpdateHandler updateHandler = new UpdateHandler();
 
+  @Override
+  public boolean isValidSource(Configuration localConfiguration) {
+    if (localConfiguration.getProperty(APOLLO_CONFIG_URL_KEY) == null){
```
WillemJiang: It's better to write the Log here (to let the user know about which configure is wrong).

lijasonvip: Fixed.

PR #570, opened 2018-03-05
[SCB-360] write simpler eventBus to optimize performance

wujimin:
IDE compile warnings check:  
image

### DIFF
--- a/foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
+++ b/foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.common.eventbus;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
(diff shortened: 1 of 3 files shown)


foundations/foundation-common/pom.xml
```diff
@@ -16,65 +16,69 @@
...
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
```
liubao68: why add this dependency? Can not find it's usage from code below

foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
```diff
@@ -0,0 +1,84 @@
...
+      if (exists == null) {
+        continue;
+      }
+
+      exists.removeAll(entry.getValue());
+    }
+  }
+
+  @Override
+  public void post(Object event) {
```
liubao68: This implementation will block event senders if subscribers processed not very fast.
It's better to mention this in the class description.

foundations/foundation-common/pom.xml
wujimin: my first impl extend EventSubscriber to write a new Subscriber, if not add the dependency, will cause class not found exception
but impl changed at last, i will if need it too.

foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
wujimin: registrer  and  unregister will not block post

foundations/foundation-common/pom.xml
wujimin: no need this anymore, delete it.

foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
liubao68: OK.  This will just make events out of order when trigger events in events handling logic.

foundations/foundation-common/src/main/java/com/google/common/eventbus/SimpleEventBus.java
```diff
@@ -0,0 +1,84 @@
...
+
+// EventBus has performance problem, we create a new simpler eventBus
+//
+// extend from EventBus is not necessary
+// but EventManager has a public EventBus field, we need to make compatible
+//
+// different between this class and EventBus:
+// 1. not support cycle trigger:
+//   subscribe a, when handle a, trigger event b
+//   subscribe b, when handle b, trigger event a
```
liubao68: It's better to make this comments as Java Docs in EventManager  in case user miss use it.

wujimin: changed to javadoc in SimpleEventBus
but not in EventManager

PR #572, opened 2018-03-06
[SCB-370] Improve latency and max precision

zhengyangyong:
…ency and max latency precision of metrics

---
If operation make two request with latency 1234567 nanoseconds and 7654321 nanoseconds
Before :
latency (average) output will be 4 milliseconds,and max output will be 7 milliseconds (value after the decimal point lost)
Now :
latency (average) output will be 4.444444 milliseconds,and max output will be 7.654321 milliseconds


metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public Double getValue() {
+      return this.getValue(0);
+    }
+
+    @Override
+    public Double getValue(int pollerIndex) {
+      //we need direct div for keep value after the decimal point
+      long nanoValue = timer.getValue();
+      return (double) nanoValue / (double) 1000000;
```
wujimin: Output is ms , not second?
this is not compatible with servo and spectator

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public MonitorConfig getConfig() {
+      return timer.getConfig().withAdditionalTag(Tags.newTag(MetricsConst.TAG_STATISTIC, "latency"));
+    }
+  }
+
+  class MonitorForTimerMax implements Monitor<Double> {
+
```
wujimin: max and total time already in basicTimer, why we need so many  logic ……

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
zhengyangyong: servo default ouput is ms

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
zhengyangyong: I will not direct call getTotal() in basicTimer,this method is private for BasicTimer 
Metrics output will get all register Monitors from MonitorRegistry  and call it's getValue()
more detail can see MonitorManager.measure()

zhengyangyong: MonitorRegistry holds all registered Monitors (timer,counter,gauge etc...) , getRegistedMonitors() return base Monitor<T> interface collection;
BaiscTimer extends Monitor<Long>, getValue()  will return Long type value,so only one logic is div 1000000

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
wujimin: where is the code, what i see is second.

wujimin: it seems that you confused input unit and output unit

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
wujimin: basicTimer provide monitors, include: totalTimeFactor, count, min, max  
totalTimeFactoris FactorMonitor wrapped totalTime  
totalTime is StepCounter  
FactorMonitor .getValue delegate to StepCounter.getValue  
so output is second  

did i missed something?

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
zhengyangyong: public BasicTimer(MonitorConfig config) {
        this(config, TimeUnit.MILLISECONDS);
    }

BasicTimer(MonitorConfig config, TimeUnit unit, Clock clock) {
...
timeUnitNanosFactor = 1.0 / timeUnit.toNanos(1);
...
}

private double getTotal(int pollerIndex) {
        return totalTime.getCurrentCount(pollerIndex) * timeUnitNanosFactor;
    }

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HighPrecisionBasicTimer.java
zhengyangyong: I direct use getValue() implemented in basicTimer:
    @Override
    public Long getValue(int pollerIndex) {
        final long cnt = count.getCurrentCount(pollerIndex);
        final long value = (long) (getTotal(pollerIndex) / cnt);
        return (cnt == 0) ? 0L : value;
    }

set it's output unit to nano then div 1000000

zhengyangyong:
从内部MonitorRegistry获取值改为直接输出原始值不加工，之后用户使用MetricsUtils.convertMeasurements依据需求转换单位，现在已有场景：
1./metrics Publish 使用**秒**
2.对接Prometheus 使用毫秒
3.perf 打印使用毫秒
4.写文件的例子使用毫秒

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/MetricsUtils.java
```diff
@@ -0,0 +1,103 @@
...
+              convertFunctions.get(unit).apply(measurement.getValue()));
+        } else {
+          metrics.put(getMonitorKey(measurement.getKey()), measurement.getValue());
+        }
+      }
+      return metrics;
+    }
+    //no need support MINUTES,HOURS,DAYS because latency under this unit is unsuitable
+    throw new ServiceCombException(
+        "illegal unit : " + String.valueOf(unit) + ", only support NANOSECONDS,MICROSECONDS,MILLISECONDS and SECONDS");
```
yangbor: Metrics calculated in minutes or even hours is normal, we should support at least to hours.

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/MetricsUtils.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.apache.servicecomb.foundation.common.exceptions.ServiceCombException;
+import org.apache.servicecomb.foundation.metrics.MetricsConst;
+
+import com.netflix.servo.monitor.MonitorConfig;
+import com.netflix.servo.tag.Tag;
+import com.netflix.servo.tag.TagList;
+import com.netflix.servo.tag.Tags;
+
+public class MetricsUtils {
```
yangbor: XXXUtils suggests a bad design, later developers may put any unrelated things into this class.
Move the getMonitorKey to MonitorManager and rename this class to MeasurementConversion or something suitable.

PR #575, opened 2018-03-07
[SCB-369] Extract info from spectator measurement

wujimin:
image

### DIFF
--- a/foundations/foundation-metrics/src/test/java/org/apache/servicecomb/foundation/metrics/publish/spectator/TestMeasurementTree.java
+++ b/foundations/foundation-metrics/src/test/java/org/apache/servicecomb/foundation/metrics/publish/spectator/TestMeasurementTree.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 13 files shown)


foundations/foundation-metrics/src/test/java/org/apache/servicecomb/foundation/metrics/publish/spectator/TestMeasurementTree.java
```diff
@@ -0,0 +1,88 @@
...
+
+    clock.setWallTime(clock.wallTime() + 1000);
+
+    MeasurementGroupConfig config = new MeasurementGroupConfig("id", "g1", "g2", Statistic.count.key());
+    tree.from(registry.iterator(), config);
+
+    Assert.assertEquals(1, tree.getChildren().size());
+
+    MeasurementNode node = tree.findChild("id", "g1v", "g2v");
+    System.out.println(node.getMeasurements());
```
WillemJiang: Please don't use the System.out.println here.

wujimin: oh, sorry, forget to  delete test code.
and forget to add notice information, i will fix it.

wujimin: already add spectator information to license file

PR #583, opened 2018-03-10
[SCB-381] Fix foundation-vertx UT failure

yangbor:
Vertx uses some static fields which will cause problems if resued.
Disabled surefire fork for foundation-vertx to avoid this issue.

---

### DIFF
--- a/foundations/foundation-vertx/pom.xml
+++ b/foundations/foundation-vertx/pom.xml
@@ -53,4 +53,21 @@
       <artifactId>foundation-test-scaffolding</artifactId>
     </dependency>
   </dependencies>
+
+  <!--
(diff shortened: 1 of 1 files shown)


foundations/foundation-vertx/pom.xml
```diff
@@ -53,4 +53,21 @@
...
+   on some platforms(mainly linux).
+   -->
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <forkCount>1</forkCount>
+          <reuseForks>false</reuseForks>
```
wujimin: not a good idea
should fix the root cause.

PR #585, opened 2018-03-12
[SCB-374] Invocation publish life event

wujimin:
invocation publish start and finish event, and event subscribers can get startTime/startExecutionTime/finishTime from invocation or event.


WillemJiang:
@wujimin Could you add a brief description about what you have done in this PR?

common/common-rest/src/test/java/org/apache/servicecomb/common/rest/TestAbstractRestInvocation.java
```diff
@@ -104,13 +110,23 @@ protected OperationLocator locateOperation(ServicePathManager servicePathManager
...
   AbstractRestInvocation restInvocation = new AbstractRestInvocationForTest();
 
+  @BeforeClass
+  public static void classSetup() {
+    EventManager.eventBus = new EventBus();
+  }
+
+  @AfterClass
+  public static void classTeardown() {
+    EventManager.eventBus = new EventBus();
```
WillemJiang: Do we need to setup new EventBus again here?

wujimin: no
out eventBus extend from this one, but faster.

wujimin:
added.
but title is very clear.

wujimin:
image

PR #607, opened 2018-03-15
[SCB-406] Chassis must support standard parameter validation handler

acsukesh:

---

### DIFF
--- a/handlers/handler-validator/src/main/java/org/apache/servicecomb/validator/ParamValidateHandler.java
+++ b/handlers/handler-validator/src/main/java/org/apache/servicecomb/validator/ParamValidateHandler.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.validator;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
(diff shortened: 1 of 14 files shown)


handlers/handler-validator/src/main/java/org/apache/servicecomb/validator/ParamValidateHandler.java
```diff
@@ -0,0 +1,114 @@
...
+public class ParamValidateHandler implements Handler {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ParamValidateHandler.class);
+
+  private static Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+
+  @Override
+  public void handle(Invocation invocation, AsyncResponse asyncResp) throws Exception {
+    Object[] args = invocation.getArgs();
+    boolean invalid = false;
+    List<Object> errList = new ArrayList<>();
```
wujimin: both solution and code is too complex
i remember it's very simple to IMPL validator, i'll give a example later

jeho0815: have you showed the example to sukesh? @wujimin

handlers/handler-validator/src/main/java/org/apache/servicecomb/validator/ParamValidateHandler.java
```diff
@@ -0,0 +1,114 @@
...
+    if (null != args) {
+      for (Object arg : args) {
+        Set<ConstraintViolation<Object>> violations = validator.validate(arg);
```
wujimin: validation annotation addressed not only model field, but also method parameter
so just validate arg is useless.

validate method invoke should use this way:
javax.validation.Validator.forExecutables()
javax.validation.executable.ExecutableValidator.validateParameters(T, Method, Object[], Class<?>...)

and i'm not sure if we need to validate return value:
javax.validation.executable.ExecutableValidator.validateReturnValue(T, Method, Object, Class<?>...)

wujimin: invocation arguments in  handler, is about contract arguments, not always equals producer arguments.
so validation feature should be inject to SwaggerProducerOperation.
maybe need to create a extension point before producerMethod.invoke, do not named it to be validateXXX, this mechanism just like AOP.

wujimin: do not hard code response for validate failed, just throw some validation exception(maybe you can reference spring mechanism)
by org.apache.servicecomb.swagger.invocation.exception.ExceptionToResponseConverters,we provide a default validation exception converter
add a method: int getOrder() in ExceptionToResponseConverter, so that customers can override our default converter.

acsukesh:
validation will be added in already existing producer invoke handler

PR #609, opened 2018-03-17
[SCB-383] Metrics subscribe invocation life event and statistics

wujimin:
image

### DIFF
--- a/foundations/foundation-metrics/src/main/java/org/apache/servicecomb/foundation/metrics/MetricsBootstrap.java
+++ b/foundations/foundation-metrics/src/main/java/org/apache/servicecomb/foundation/metrics/MetricsBootstrap.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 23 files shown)


foundations/foundation-metrics/src/main/java/org/apache/servicecomb/foundation/metrics/MetricsBootstrap.java
```diff
@@ -0,0 +1,82 @@
...
+        new ThreadFactoryBuilder()
+            .setNameFormat("spectator-poller-%d")
+            .build());
+
+    loadMetricsInitializers();
+    startPoll();
+  }
+
+  public void shutdown() {
+    executorService.shutdown();
```
liubao68: Do you actually mean shutdownNow?

wujimin: only one instance, and shutdown will not interrupt task

foundations/foundation-metrics/src/main/java/org/apache/servicecomb/foundation/metrics/MetricsInitializer.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.foundation.metrics;
+
+import com.google.common.eventbus.EventBus;
+import com.netflix.spectator.api.CompositeRegistry;
+
+public interface MetricsInitializer {
+  // if create new registry, must add to globalRegistry
```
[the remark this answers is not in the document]
wujimin: done

PR #613, opened 2018-03-19
[SCB-416]For load balance rule configurations, we need provider service level configuration

liubao68:

---


handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/Configuration.java
```diff
@@ -101,10 +101,11 @@ public String getRuleStrategyName(String microservice) {
         PROP_ROOT + PROP_RULE_STRATEGY_NAME);
   }
 
-  public int getSessionTimeoutInSeconds() {
+  public int getSessionTimeoutInSeconds(String microservice) {
     final int defaultValue = 30;
     // do not support MicroService level now
```
weichao666: this annotation can be deleted.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/Configuration.java
```diff
@@ -113,10 +114,11 @@ public int getSessionTimeoutInSeconds() {
     }
   }
 
-  public int getSuccessiveFailedTimes() {
+  public int getSuccessiveFailedTimes(String microservice) {
     final int defaultValue = 5;
     // do not support MicroService level now
```
weichao666: this annotation can be deleted too.

handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalancer.java
```diff
@@ -146,4 +148,32 @@ public void testGetAllServers() {
     Mockito.when(filter.getFilteredListOfServers(servers)).thenReturn(servers);
     Assert.assertEquals(servers, loadBalancer.getAllServers());
   }
+  
+  @Test
+  public void testLaodBalanceWithSessionSticknessRule() {
```
weichao666: it's testLoad, not testLaod

handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalancer.java
```diff
@@ -146,4 +148,41 @@ public void testGetAllServers() {
...
+    lb.setServerList(servers);
+    
+    Server s = lb.chooseServer("test");
+    Assert.assertEquals(server2, s);
+    s = lb.chooseServer("test");
+    Assert.assertEquals(server2, s);
+    
+    long time =  Deencapsulation.getField(rule, "lastAccessedTime");
+    Deencapsulation.setField(rule, "lastAccessedTime", time - 1000 * 10);
+    Utils.updateProperty("cse.loadbalance.service.SessionStickinessRule.sessionTimeoutInSeconds", 9);
```
wujimin: is it possible to use: org.apache.servicecomb.foundation.test.scaffolding.config.ArchaiusUtils.setProperty(String, Object)
if can not use it, is it possible to add a new method in it.
so that we can reuse the capability in other test case.

PR #626, opened 2018-03-29
[SCB-447] Optimize SPIServiceUtils

wujimin:
1.cache SPI service instances, avoid create new instance every time
2.support find special IPML instance

image


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/SPIServiceUtils.java
```diff
@@ -76,18 +63,65 @@ private SPIServiceUtils() {
...
     });
 
-    return serviceEntries.stream()
+    List<T> services = serviceEntries.stream()
         .sorted(Comparator.comparingInt(Entry::getKey))
         .map(Entry::getValue)
         .collect(Collectors.toList());
+
+    LOGGER.info("Found SPI service {}, count={}.", serviceType.getName(), services.size());
+    for (int idx = 0; idx < services.size(); idx++) {
```
WillemJiang: Can we use the accumulator to do this job ?

PR #628, opened 2018-03-31
[SCB-385] Metrics publisher switch to new mechanism

wujimin:
image


demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
```diff
@@ -96,20 +96,14 @@ public static void run() {
...
-      TestMgr.check("true", "false");
-    }
+    @SuppressWarnings("unchecked")
+    Map<String, Double> metrics = restTemplate.getForObject(prefix + "/metrics", Map.class);
+
+//    TestMgr.check(true, metrics.get("jvm(name=heapUsed,statistic=gauge)") != 0);
+    TestMgr.check(true, metrics.size() > 0);
+    TestMgr.check(true,
+        metrics.get(
+            "servicecomb.invocation(operation=springmvc.codeFirst.saySomething,role=PRODUCER,stage=total,statistic=count,status=200,transport=highway)") >= 0);
```
WillemJiang: Can we use the lower case in the role tage?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
```diff
@@ -195,16 +195,11 @@ private boolean needSetFrameworkVersion() {
...
-      Version scVersion = VersionUtils.getOrCreate(serviceCenterInfo.getVersion());
-      Version frameworkVersion = VersionUtils.getOrCreate(Const.SERVICECENTER_FRAMEWORK_VERSION);
-
-      if (scVersion.compareTo(frameworkVersion) >= 0) {
-        return true;
-      } else {
-        return false;
-      }
     }
+
```
WillemJiang: This change doesn't relate to this JIRA, we need to avoid mixing the unrelated changes together.

demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
wujimin: role value is from InvocationType enum, i think we should follow the rule with other features that use invocationType directly

PR #630, opened 2018-03-31
[SCB-444]try to optimize autoDiscovery function

mt-monster:

---
default

This PR is intend to optimize autoDiscovery function. as we know ,now we use retry at the first time sdk fail to register sc . when deployed in multi sc ( more than one sc instance ) enviroment , one sc instance breakup, this stategy exhausts lot of time to switch ip to next one , so i think it better to retry all the loop-ips in enviroment so that sdk can meet the health sc intance.

@liubao68 @wujimin @WillemJiang

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/IpPortManager.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/IpPortManager.java
@@ -50,6 +50,12 @@ public class IpPortManager {
 
   private boolean autoDiscoveryInited = false;
 
+  private int maxRetryTimes;
+
+  public int getMaxRetryTimes() {
+    return maxRetryTimes;
+  }
+
   public IpPortManager(ServiceRegistryConfig serviceRegistryConfig, InstanceCacheManager instanceCacheManager) {
     this.serviceRegistryConfig = serviceRegistryConfig;
     this.instanceCacheManager = instanceCacheManager;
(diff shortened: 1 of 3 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/IpPortManager.java
```diff
@@ -50,6 +50,12 @@
 
   private boolean autoDiscoveryInited = false;
 
+  private int maxRetryTimes;
+
+  public int getMaxRetryTimes() {
+    return maxRetryTimes;
```
liubao68: Maybe maxRetryTimes should be bigger than 2. Return Max(maxRestryTime, 2)？

mt-monster: no need ,already retry twice

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -98,7 +98,7 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
       HttpClientResponse response = restResponse.getResponse();
```
yangbor: Duplicated code piece for retry, extract it as a method.

PR #632, opened 2018-04-03
[SCB-456]Provider a way to input configuration from a Map

jasononion:

---


foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -52,9 +52,15 @@
 
   private static final String MICROSERVICE_CONFIG_LOADER_KEY = "cse-microservice-config-loader";
 
+  private static ConfigModel model = new ConfigModel();
+
   private ConfigUtil() {
   }
 
+  public static void addConfigs(Map<String, Object> config) {
```
WillemJiang: As this method is just set the internal cofigModel,  it's better to use setConfigs instead of addConfigs.

wujimin:
what's requirements of this PR?
just provide a chance for developers to init some configuration before servicecomb init?

if we want to do this, just let developers invoke System.setProperties(Properties props), it's enough?

jasononion:
@wujimin This PR is for inputing configuration from Map instead of microservice.yaml

wujimin:
System.setProperties(Properties props) can take the same affect?

jasononion:
@wujimin Yes, these two method both can input configurations, but this PR can provider a new choice

WillemJiang:
@wujimin  I don't like the way of System.setProperties(Properties props), it could cause some side effect there are some starter building the microservice metadata in the same JVM.

foundations/foundation-config/src/test/java/org/apache/servicecomb/config/TestConfigUtil.java
```diff
@@ -80,6 +81,16 @@ public static void tearDown() throws Exception {
     ArchaiusUtils.resetConfig();
   }
 
+  @Test
+  public void testAddConfig() {
+    Map config = new HashMap<String, Object>();
+    config.put("APPLICATION_ID", "app");
+    ConfigUtil.setConfigs(config);
+    ConcurrentCompositeConfiguration configuration = ConfigUtil.createLocalConfig();
+    Assert.assertEquals(configuration.getString("APPLICATION_ID"), "app");
```
liubao68: Can you provide a test case for dynamic property change is work? (callback)

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -52,9 +52,15 @@
 
   private static final String MICROSERVICE_CONFIG_LOADER_KEY = "cse-microservice-config-loader";
 
+  private static ConfigModel model = new ConfigModel();
+
   private ConfigUtil() {
   }
 
+  public static void setConfigs(Map<String, Object> config) {
```
liubao68: I think if we can provide a more convenient method to add configs . e.g. 
public static void addConfig(String k, Object v)

WillemJiang:
@jasononion  Could you take a look at the new comments from Liubao?

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
jasononion: you mean we provide both addConfig and setConfigs? or just replace setConfigs ?

PR #635, opened 2018-04-04
[SCB-462] cloud eye publisher switch to new mechanism

wujimin:
image


samples/metrics-write-file-sample/metrics-write-file/src/main/resources/sample-log4j2.xml
```diff
@@ -0,0 +1,29 @@
+<Configuration status="INFO">
```
WillemJiang: License header please.

wujimin: done

samples/metrics-write-file-sample/metrics-write-file/src/main/java/org/apache/servicecomb/samples/mwf/CloudEyeMetricModel.java
```diff
@@ -0,0 +1,93 @@
...
+  @JsonAnyGetter
+  public Map<String, String> getDynamicValue() {
+    return dynamicValue;
+  }
+
+  public void setDynamicValue(Map<String, String> dynamicValue) {
+    this.dynamicValue = dynamicValue;
+  }
+
+  public static void main(String[] args) throws IOException {
```
liubao68: I think this test case can be removed

wujimin: stupid mistake
fixed

PR #651, opened 2018-04-17
[SCB-498] Configuration Center ip address need to config default port

weichao666:

---
for http/https, the default ports are 80 and 443, if port is not configured, the default port needs to be used. this simplifies the configuration.


yangbor:
It seems to be the wrong place to fix the problem. the http(80) and https(443) are Well Known Ports and all http servers and clients should handle them correctly. We should fix the code where this is not correctly handled instead of appending the port to the URL.

weichao666:
modified

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -145,12 +145,25 @@ public static IpPort parseIpPortFromURI(String uriAddress) {
...
   }
 
+  private static IpPort parseIpPort(String scheme, String authority) {
+    if (authority == null)
+      return null;
+    int idx = authority.indexOf(':');
+    if (idx != -1)
+      return parseIpPort(authority);
+    if (scheme.equals("http"))
+      return parseIpPort(new StringBuffer(authority).append(":80").toString());
```
liubao68: return new IpPort(authority, 80)

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -145,12 +145,25 @@ public static IpPort parseIpPortFromURI(String uriAddress) {
...
+    if (scheme.equals("https"))
+      return parseIpPort(new StringBuffer(authority).append(":443").toString());
+    return parseIpPort(authority);
```
liubao68: return null;

weichao666: only http and https was deal with here, if it's starts with others, e.g "rest://",shouldn't return null

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
weichao666: done

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
weichao666: done

sh.exe.stackdump
```diff
@@ -0,0 +1,11 @@
+Stack trace:
```
wujimin: why we need this file?

foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/net/TestNetUtils.java
```diff
@@ -49,6 +49,10 @@ public void testNetutils() {
     Assert.assertEquals(NetUtils.parseIpPortFromURI(null), null);
     Assert.assertEquals(NetUtils.parseIpPortFromURI("ss"), null);
     Assert.assertEquals(NetUtils.parseIpPortFromURI("rest://127.0.0.1:8080").getHostOrIp(), "127.0.0.1");
+    Assert.assertEquals(NetUtils.parseIpPortFromURI("http://127.0.0.1:8080").getPort(), 8080);
```
wujimin: expect value is the first parameter

weichao666: done

PR #671, opened 2018-04-23
[SCB-516] fix accessLog traceId printing problem in EdgeService

yhs0092:

---
See details in [SCB-516](https://issues.apache.org/jira/browse/SCB-516).
- add a TracePrepareHandler to set traceId
- make TraceIdItem can get traceId from request headers


yhs0092:
- I’ve modified the Getter in ZipkinProviderDelegate.java to avoid brave's abandoning the traceId generated by ServiceComb.
- make ServiceComb generate traceId in hexadecimal format as I find that brave does it like so.
- set access log printing handler in `context.response().endHandler()` instead of `context.addBodyEndHandler()` because the comment of `addBodyEndHandler()` says that it may not be invoked.

PR #690, opened 2018-05-05
[SCB-529] producer download file from byte[]

wujimin:


wujimin:
infact, maybe it's not a good practic
1.if byte[] come from file, then return file is better
2.if byte[] come from inputStream, then return InputStream is better

i don't know what's other scenes
so currently, the only purpose is just make developer's old code can run.

PR #692, opened 2018-05-07
[SCB-543]optimize registry procedure

mt-monster:

---
@liubao68 @laijianbin @wujimin @WillemJiang 
In one case of csp upgrade scene 1: successfully registry microsevice at first , 2: then delete the microservice static information but instance is still registrying , the result is registry instance in hanging on because sc has no microservice static information.so in order to optimize this registry halt , i put registry instance procedure to registry microservice again so let the registry can pass


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/event/InstanceRegistryFailedEvent.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.servicecomb.serviceregistry.task.event;
+
+/**
```
wujimin: template is not correct?

mt-monster: ok.fix it right now

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -52,6 +53,11 @@ public void onMicroserviceInstanceHeartbeatTask(MicroserviceInstanceHeartbeatTas
     }
   }
 
+  @Subscribe
```
wujimin: it's better to change task controller
when failed to register instance, then change status, next time, task will run from register microservice

mt-monster: how to change task controller ?

liubao68: Just set 
this.taskStatus = TaskStatus.READY;
this.registered = false;

like onMicroserviceInstanceHeartbeatTask

mt-monster: get it

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.servicecomb.serviceregistry.task;
 
+import org.apache.servicecomb.foundation.common.event.EventManager;
```
wujimin: import bu not used

PR #698, opened 2018-05-10
[SCB-546] bug fix

jeho0815:
in production enviroment, first time register microservice, schemas's summary is null, registerSchemas will throw exception.

if the summary is null, will re-gister the schema.

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
@@ -155,7 +155,7 @@ public class MicroserviceRegisterTask extends AbstractRegisterTask {
         String curSchemaSumary = existSchema.getSummary();
         String schemaSummary = Hashing.sha256().newHasher().putString(content, Charsets.UTF_8).hash().toString();
         if (!schemaSummary.equals(curSchemaSumary)) {
-          if (microservice.getInstance().getEnvironment().equalsIgnoreCase("development")) {
+          if (curSchemaSumary == null || microservice.getInstance().getEnvironment().equalsIgnoreCase("development")) {
             LOGGER.info(
                 "schemaId [{}]'s content changes and the current enviroment is development, so re-register it!",
                 schemaId);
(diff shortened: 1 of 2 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -155,7 +155,7 @@ private boolean registerSchemas() {
         String curSchemaSumary = existSchema.getSummary();
         String schemaSummary = Hashing.sha256().newHasher().putString(content, Charsets.UTF_8).hash().toString();
         if (!schemaSummary.equals(curSchemaSumary)) {
-          if (microservice.getInstance().getEnvironment().equalsIgnoreCase("development")) {
+          if (curSchemaSumary == null || microservice.getInstance().getEnvironment().equalsIgnoreCase("development")) {
```
wujimin: boolean exists = (existSchema != null) && (existSchema .getSummary() != null);

wujimin: if schema content is null, then local registry impl works well?

wujimin: why CI passed for preview version?

jeho0815: 1、has changed, but old version' summary content is null too, may has some potential problem
2、local impl work well, because it generate one list on startup
3、i don't know why CI can't check this error out. will check go on

wujimin: 1.must compatible with old version
   treat summary is null as schema not registered, register again with summary, SC allow this action.
2.currently getSchemas of local impl is not equals to SC, infact it's a bug
3.ok, remember to check it.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -145,7 +145,7 @@ private boolean registerSchemas() {
       String schemaId = entry.getKey();
       String content = entry.getValue();
       GetSchemaResponse existSchema = extractSchema(schemaId, existSchemas);
-      boolean exists = existSchema != null;
+      boolean exists = existSchema != null && existSchema.getSummary() != null;
```
wujimin: SC support batch register schemas
we should swtich to new api

be careful that maybe will send too big request, need to split to multi request by max size of limit.

PR #699, opened 2018-05-11
[SCB-532] Support recursive dependency

wujimin:
support create recursive dependency class
```
1. class A {
  A a;
}

2. class A {
  B b;
}
class B {
  A a;
}
```

### DIFF
--- a/common/common-javassist/src/main/java/org/apache/servicecomb/common/javassist/JavassistUtils.java
+++ b/common/common-javassist/src/main/java/org/apache/servicecomb/common/javassist/JavassistUtils.java
@@ -139,7 +139,7 @@ public final class JavassistUtils {
     return createClass(null, config);
   }
 
-  public static Class<?> createClass(ClassLoader classLoader, ClassConfig config) {
+  public static CtClass createCtClass(ClassLoader classLoader, ClassConfig config) {
     if (classLoader == null) {
       classLoader = Thread.currentThread().getContextClassLoader();
     }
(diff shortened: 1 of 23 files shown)


common/common-javassist/src/main/java/org/apache/servicecomb/common/javassist/JavassistUtils.java
```diff
@@ -185,10 +185,47 @@ private static void addEnumValuesMethod(CtClass ctClass, List<String> values) th
...
+      return ctClass;
+    } catch (Throwable e) {
+      throw new IllegalStateException(
+          String.format("Failed to create CtClass %s in classLoader %s.", config.getClassName(),
+              classLoader), e);
+    }
+  }
+
+  public static Class<?> createClass(ClassLoader classLoader, ClassConfig config) {
+    if (classLoader == null) {
```
WillemJiang: It's better to add a log here for tracking the classloader issue.

common/common-javassist/src/main/java/org/apache/servicecomb/common/javassist/JavassistUtils.java
```diff
@@ -185,10 +185,47 @@ private static void addEnumValuesMethod(CtClass ctClass, List<String> values) th
...
+      classLoader = Thread.currentThread().getContextClassLoader();
+    }
+
+    CtClass ctClass = createCtClass(classLoader, config);
+    return createClass(classLoader, ctClass);
+  }
+
+  public static Class<?> createClass(ClassLoader classLoader, CtClass ctClass) {
+    if (classLoader == null) {
+      classLoader = Thread.currentThread().getContextClassLoader();
```
WillemJiang: Same issue here.

common/common-javassist/src/main/java/org/apache/servicecomb/common/javassist/JavassistUtils.java
wujimin: i have test for this
because already throw exception with error context, invoker already print the exception
if print in this place, then will print multi times.

so i think it's better not log here.
what's your suggestion?

WillemJiang: oh, my suggestion is not print the exception here , it just write some log message for switching  the classloader to  thread context classloader.

wujimin: Whether success or failure， will print the class belongs to which classLoader
so that no need to print this log.

PR #702, opened 2018-05-14
[SCB-580] fix response of the upload file too large

yhs0092:

---
see details in [SCB-580](https://issues.apache.org/jira/browse/SCB-580)

### DIFF
--- a/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
+++ b/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
@@ -17,6 +17,11 @@
 
 package org.apache.servicecomb.transport.rest.vertx;
 
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.core.Response.Status.Family;
+
 import org.apache.servicecomb.common.rest.AbstractRestInvocation;
 import org.apache.servicecomb.common.rest.RestConst;
 import org.apache.servicecomb.common.rest.RestProducerInvocation;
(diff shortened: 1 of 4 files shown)


transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
```diff
@@ -64,9 +69,71 @@ private void failureHandler(RoutingContext context) {
...
+    // the connection should be closed.
+    handleFailureAndClose(context, restProducerInvocation, e);
+  }
+
+  /**
+   * Try to find out the failure information and send it in response.
+   */
+  private void handleFailureAndClose(RoutingContext context, AbstractRestInvocation restProducerInvocation,
+      Throwable e) {
+    if (null != restProducerInvocation) {
```
wujimin: 1.flat the branchs
2.most branchs did not close?

yhs0092: These two problems have been fixed. Please review again.

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
```diff
@@ -64,9 +69,85 @@ private void failureHandler(RoutingContext context) {
...
+
+  /**
+   * Consumer will treat the response body as json by default, so it's necessary to wrap response body with ""
+   * to avoid deserialization error.
+   *
+   * @param message response body
+   * @return response body wrapped with ""
+   */
+  private String wrapResponseBody(String message) {
+    return "\"" + message + "\"";
```
liubao68: Is it possible to use Json encode API?

yhs0092: Maybe it's a little bit complex to cover all situations, because I'm not sure what is contained in the message.
But I'll try to make it and reply later.

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
```diff
@@ -64,9 +69,85 @@ private void failureHandler(RoutingContext context) {
...
+    context.response().close();
+  }
+
+  /**
+   * Use routingContext to send failure information in throwable.
+   */
+  private void sendExceptionByRoutingContext(RoutingContext context, Throwable e) {
+    if (InvocationException.class.isInstance(e)) {
+      InvocationException invocationException = (InvocationException) e;
+      context.response().putHeader(HttpHeaders.CONTENT_TYPE, MediaType.WILDCARD)
```
liubao68: wrapResponseBody uses json, is it here APPLICATION/JSON?

yhs0092: If user set provider's operation produce "text/plain", and we set the Content-Type header as "application/json", it will cause consumer failed to decode response body, and the consumer will response like below:
```json
{
    "message": "method POST, path /upload/multipartFileUpload/, statusCode 413, reasonPhrase Request Entity Too Large, response content-type application/json is not supported"
}
```

But "*/*" will let the producer use the default ProduceProcessor to decode the body. No matter it's json processor or text processor, it will handle the body without exception thrown.

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
wujimin: if there is some special char in message(eg: "), will output a invalid json string
so must use json encode API

yhs0092: I've done it. Thanks to the powerful special char escape ability of Vertx's JsonObject, it's not so complex as I expected. 
In the UT testWrapResponseBody, you can see that JsonObject can handle special char well.

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
```diff
@@ -64,9 +70,106 @@ private void failureHandler(RoutingContext context) {
...
+   */
+  String wrapResponseBody(String message) {
+    if (isValidJson(message)) {
+      return message;
+    }
+
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.put("message", message);
+
+    return jsonObject.toString();
```
wujimin: this will produce:
{
  "message": ......
}

is this we expected?

yhs0092: Actually I have no idea about what kind of Json response should be written. But I find out that currently we wrap exception and write response like below, so I copy this pattern.

wujimin: ok
because your last commit just return a string, now return a CommonExceptionData structure, so i have this question

PR #704, opened 2018-05-15
[SCB-582]Provide a way to protect instance removal

liubao68:

---

### DIFF
--- a/demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
+++ b/demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
@@ -27,7 +27,7 @@ servicecomb:
         healthCheck:
           interval: 2
         watch: false
-        remove:
+        empty:
           protection: true
   handler:
     chain:
--- a/demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
+++ b/demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
@@ -23,6 +23,12 @@ servicecomb:
   service:
     registry:
       address: http://127.0.0.1:30100
+      instance:
+        healthCheck:
+          interval: 2
+        watch: false
+        remove:
+          protection: true
   handler:
     chain:
       Consumer:
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
@@ -23,12 +23,15 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
+import org.apache.servicecomb.foundation.common.utils.SPIServiceUtils;
 import org.apache.servicecomb.serviceregistry.RegistryUtils;
 import org.apache.servicecomb.serviceregistry.api.Const;
 import org.apache.servicecomb.serviceregistry.api.registry.MicroserviceInstance;
 import org.apache.servicecomb.serviceregistry.api.registry.MicroserviceInstanceStatus;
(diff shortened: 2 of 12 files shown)


liubao68:
Change the behavior to only protect empty instances and make this behavior on default.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/SimpleMicroserviceInstancePing.java
```diff
@@ -0,0 +1,47 @@
...
+  @Override
+  public int getOrder() {
+    return 100;
+  }
+
+  @Override
+  public boolean ping(MicroserviceInstance instance) {
+    if (instance.getEndpoints() != null && instance.getEndpoints().size() > 0) {
+      IpPort ipPort = NetUtils.parseIpPortFromURI(instance.getEndpoints().get(0));
+      try (Socket s = new Socket(ipPort.getHostOrIp(), ipPort.getPort())) {
```
wujimin: lost heartbeat means tens of seconds passed
if deployment managered by docker, in this time, maybe already start another microservce instance for this ip and port

liubao68: I have think this problem very carefully. This mechanism only work when client get empty instances list, and we face this problem due to service center bugs, unexpected start/stop of microservices and service center, and other network failures not properly handled. In theory, these problems should be properly handled and SDK do not need to care about it, but It happens very frequently. 
So we add this mechanism to make SDK itself more stable in bad circumstances. Since SDK will query instances periodically, after there are instances, no problem for your mentioned scenario. And if even your scenario happens(rarely), this time, to consumer both of them are failures(no instance vs send to wrong destination). This time is very short and acceptable.

demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
```diff
@@ -23,6 +23,12 @@ servicecomb:
   service:
     registry:
       address: http://127.0.0.1:30100
+      instance:
+        healthCheck:
+          interval: 2
+        watch: false
+        empty:
```
WillemJiang: Please update document if there is new option introduced. 
Is there any other empty related options?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -169,6 +166,28 @@ private void setInstances(List<MicroserviceInstance> pulledInstances, String rev
...
+        .filter(instance -> {
+          return MicroserviceInstanceStatus.UP.equals(instance.getStatus());
+        })
+        .collect(Collectors.toList());
+    if (upInstances.isEmpty() && inUseInstances != null && ServiceRegistryConfig.INSTANCE.isEmptyInstanceProtectionEnabled()) {
+      MicroserviceInstancePing ping = SPIServiceUtils.getPriorityHighestService(MicroserviceInstancePing.class);
+      inUseInstances.stream()
+          .forEach(instance -> {
+            if (!upInstances.contains(instance)) {
+              if (ping.ping(instance)) {
```
WillemJiang: As the ping method could take some time. It could take a while if there are lots of upInstances need to be merged.

liubao68: This scenario is rarely as I explained above, mostly because of service center bugs. And this health check run a separate background thread, will not influence normal performance.

demo/demo-pojo/pojo-client/src/main/resources/microservice.yaml
liubao68: Currently there are a lot of configurations not documented. I create a new issue to tracker this and will add docs after we have merged documents as [VOTE] done in email. 
https://issues.apache.org/jira/browse/SCB-594

PR #712, opened 2018-05-17
[SCB-590] Update the validator version

acsukesh:

---

### DIFF
--- a/java-chassis-dependencies/pom.xml
+++ b/java-chassis-dependencies/pom.xml
@@ -169,12 +169,12 @@
       <dependency>
         <groupId>javax.validation</groupId>
         <artifactId>validation-api</artifactId>
-        <version>1.1.0.Final</version>
+        <version>2.0.0.Final</version>
       </dependency>
       <dependency>
         <groupId>org.hibernate</groupId>
         <artifactId>hibernate-validator</artifactId>
-        <version>5.2.4.Final</version>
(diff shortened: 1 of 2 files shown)


java-chassis-dependencies/pom.xml
```diff
@@ -324,7 +324,7 @@
       <dependency>
         <groupId>org.hibernate</groupId>
         <artifactId>hibernate-validator-annotation-processor</artifactId>
-        <version>5.2.2.Final</version>
+        <version>6.0.2.Final</version>
```
wujimin: license file need to be updated too.

WillemJiang: Please update the version in the https://github.com/apache/incubator-servicecomb-java-chassis/blob/master/java-chassis-distribution/src/release/LICENSE#L349

acsukesh: Fixed

WillemJiang: It looks like the annotation processor is option one,  it's just for the IDE validation.  Do we need to ship it?

yangbor:
@WillemJiang The hibernate-validator-annotation-processor is not shipped with the binary release(I chechked that the jar is not packed in to the binary release). hibernate-validator is shipped and mentioned in the license file.

As for other jars missing in the license file, I'll check and add them after the feature freeze.

PR #714, opened 2018-05-17
[SCB-579] fix NullPointerException when consumer upload null file

yhs0092:

---
See details in [SCB-579](https://issues.apache.org/jira/browse/SCB-579)

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/RestClientRequestImpl.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/RestClientRequestImpl.java
@@ -83,6 +83,10 @@ public class RestClientRequestImpl implements RestClientRequest {
 
   @Override
   public void attach(String name, Part part) {
+    if (null == part) {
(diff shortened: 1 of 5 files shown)


WillemJiang:
Not sure if the file name is null is common use case. If it is not, we'd better to use Info to tell the system operator there is something is abnormal.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/RestClientRequestImpl.java
```diff
@@ -169,6 +173,10 @@ private void attachFile(String boundary, Iterator<Entry<String, Part>> uploadsIt
     // because pojo consumer not easy to set name to part
     String name = entry.getKey();
     Part part = entry.getValue();
+    if (null == part) {
```
liubao68: Since uploads do not have null entries, this is redundant null check.

yhs0092: Redundant null check has been removed. Please review again.

PR #727, opened 2018-05-24
[SCB-607] Accesslog support invocation context item

yhs0092:

---
See details in [SCB-607](https://issues.apache.org/jira/browse/SCB-607)

- set Invocation into RoutingContext
- add an AccessLogItem to get content from invocation context


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -249,4 +265,21 @@ protected void onExecuteHttpServerFiltersFinish(Response response, Throwable e)
...
+  }
+
+  /**
+   * After the {@link AbstractRestInvocation#invocation} is instantiated, the {@link AbstractRestInvocation#invocation}
+   * will be passed into {@link AfterCreateInvocationHandler#handle(Invocation)} so that users can do something with it.
+   */
+  public static interface AfterCreateInvocationHandler {
+    /**
+     * @see AfterCreateInvocationHandler
+     * @param invocation {@link AbstractRestInvocation#invocation}
```
liubao68: This document refers itself, and can be deleted

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -249,4 +265,21 @@ protected void onExecuteHttpServerFiltersFinish(Response response, Throwable e)
       invocation.onFinish(response);
     }
   }
+
+  public void setAfterCreateInvocationHandler(
```
liubao68: Maybe create a new protected method is better.
protected void afterCreateInvocation()
protected void beforeCreateInvocation()
protected void createInvocation()

 And we can implement it in RestProducerInvocation, e.g
RestProducerInvocation invo = new RestProducerInvocation.java(RoutingContext context)
..

edge/edge-core/src/main/java/org/apache/servicecomb/edge/core/EdgeInvocation.java
```diff
@@ -59,6 +59,7 @@ public void init(String microserviceName, RoutingContext context, String path,
     this.responseEx = new VertxServerResponseToHttpServletResponse(context.response());
     this.httpServerFilters = httpServerFilters;
     requestEx.setAttribute(RestConst.REST_REQUEST, requestEx);
+    setAfterCreateInvocationHandler(invocation -> context.put(RestConst.REST_INVOCATION_CONTEXT, invocation));
```
wujimin: if just need to do something after createInvocation
just override and call super first is enough?

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
```diff
@@ -182,6 +182,8 @@ private void onRequest(RoutingContext context) {
 
     RestProducerInvocation restProducerInvocation = new RestProducerInvocation();
     context.put(RestConst.REST_PRODUCER_INVOCATION, restProducerInvocation);
+    restProducerInvocation.setAfterCreateInvocationHandler(
```
wujimin: both edge and normal need to set this, so set it to be default action?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
yhs0092: I guess you mean to replace the `AfterCreateInvocationHandler` with the method `protected void afterCreateInvocation()` ?
But currently we cannot get RoutingContext directly in `AbstractRestInvocation`, unless we cast the type of  `AbstractRestInvocation.requestEx` and add a getter into `VertxServerRequestToHttpServletRequest` like below:
```java
protected void afterCreateInvocation() {
  if (!VertxServerRequestToHttpServletRequest.class.isAssignableFrom(this.requestEx.getClass())) {
    return;
  }
  VertxServerRequestToHttpServletRequest request = (VertxServerRequestToHttpServletRequest) this.requestEx;
  RoutingContext routingContext = request.getContext();
  routingContext.put(RestConst.REST_INVOCATION_CONTEXT, this.invocation);
}
```

Am I right?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
yhs0092: `AfterCreateInvocationHandler` has been removed.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
yhs0092: The code has been refactored, please review the new commit

edge/edge-core/src/main/java/org/apache/servicecomb/edge/core/EdgeInvocation.java
yhs0092: New setting logic is in the overridden `createInvocation()` method, please review the new commit

transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestDispatcher.java
yhs0092: The code has been refactored, please review the new commit

PR #728, opened 2018-05-25
 [SCB-589] Consumer producer upgrade unordered

wujimin:
1.normal consumer use PrivateMicroserviceClassLoaderFactory, just like edge
2.consumer microserviceMeta will not save in microserviceMetaManager any more, just like edge, use appManager; and change consumer default versionRule from latest to all
3.allow consumer schema set or method set bigger than contract
4.protobuf dynamic class created in microservice classloader

integration test case:
1.start old producer
2.start new consumer, which have new schems or new method belongs to old schema
  features about old producer no problem
3.start new producer
  not restart consumer, features about new producer become no problem
  
------------
All commits are logically related, difficult to split to subtasks
please review commits one by one.

there is no automatic integration test case, because need a controller to do this,  our integration test not support this now.


common/common-protobuf/src/main/java/org/apache/servicecomb/codec/protobuf/definition/ProtobufManager.java
```diff
@@ -45,13 +50,28 @@
...
+   * only for app classloader
+   * @return
+   */
+  public static ScopedProtobufSchemaManager getDefaultScopedProtobufSchemaManager() {
+    return defaultScopedProtobufSchemaManager;
+  }
+
   public static OperationProtobuf getOrCreateOperation(OperationMeta operationMeta) throws Exception {
-    OperationProtobuf operationProtobuf = operationMeta.getExtData(ProtobufManager.EXT_ID);
+    OperationProtobuf operationProtobuf = operationMeta.getExtData(EXT_ID);
```
WillemJiang: if the ExtData is not volatile,  and there are two thread a modify it, you may face some issue here.

core/src/main/java/org/apache/servicecomb/core/provider/consumer/ConsumerProviderManager.java
```diff
@@ -54,38 +51,32 @@ public void init() throws Exception {
...
+        .get();
+    String transport = DynamicPropertyFactory.getInstance()
+        .getStringProperty(key + ".transport", defaultTransport)
+        .get();
+
+    return new ReferenceConfig(appManager, microserviceName, versionRule, transport);
+  }
+
+  public ReferenceConfig getReferenceConfig(String microserviceName) {
+    return referenceConfigMap.computeIfAbsent(microserviceName, this::createReferenceConfig);
```
WillemJiang: Not sure if the ConcurrentHashMapEx.computeIfAbsent is atomic,  but if we use the JDK's  one, it's atomic.
https://stackoverflow.com/questions/26481796/concurrenthashmap-computeifabsent

core/src/main/java/org/apache/servicecomb/core/provider/consumer/InvokerUtils.java
```diff
@@ -35,31 +33,37 @@
...
     if (response.isSuccessed()) {
       return response.getResult();
     }
     throw ExceptionFactory.convertConsumerException(response.getResult());
   }
 
+  /**
+   * it's a internal API, caller make sure already invoked SCBEngine.ensureStatusUp
+   * @param invocation
+   * @return
```
WillemJiang: It's better to specify the return response.

common/common-protobuf/src/main/java/org/apache/servicecomb/codec/protobuf/definition/ProtobufManager.java
wujimin: extData is a ConcurrentHashMap, no problem.

core/src/main/java/org/apache/servicecomb/core/provider/consumer/ConsumerProviderManager.java
wujimin: ConcurrentHashMapEx extends from ConcurrentHashMap
ConcurrentHashMap.computeIfAbsent even lock when key exists, so there is performance problem

```
  // ConcurrentHashMap.computeIfAbsent always do "synchronized" operation
  // so we wrap it to improve performance
  @Override
  public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
    V value = get(key);
    if (value != null) {
      return value;
    }

    return super.computeIfAbsent(key, mappingFunction);
  }
```

core/src/main/java/org/apache/servicecomb/core/provider/consumer/InvokerUtils.java
wujimin: done

common/common-protobuf/src/main/java/org/apache/servicecomb/codec/protobuf/definition/ProtobufManager.java
```diff
@@ -37,6 +39,9 @@
 
   private static final Object LOCK = new Object();
 
+  private static ScopedProtobufSchemaManager defaultScopedProtobufSchemaManager = new ScopedProtobufSchemaManager(
```
WillemJiang: It could cause some trouble as the schema manager's class loader cannot be changed after the static initial time.

wujimin: this is for app classloader, will never changed.

PR #729, opened 2018-05-28
[SCB-562] fix NPE when had empty interface (not have any method)

zhengyangyong:

---
do not init operations if there are no path (empty interface) in swagger(prevent NPE)

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
@@ -83,42 +83,44 @@ public class SchemaMeta extends CommonService<OperationMeta> {
   }
 
   private void initOperations() {
-    if (swagger.getPaths() != null) {
-      for (Entry<String, Path> entry : swagger.getPaths().entrySet()) {
-        String strPath = entry.getKey();
-        Path path = entry.getValue();
-        for (Entry<HttpMethod, Operation> operationEntry : path.getOperationMap().entrySet()) {
-          Operation operation = operationEntry.getValue();
-          if (operation.getOperationId() == null) {
-            throw ExceptionUtils.operationIdInvalid(getSchemaId(), strPath);
-          }
+    if (swagger.getPaths() == null) {
+      return;
+    }
 
-          // org.apache.servicecomb.swagger.engine.SwaggerEnvironment.createConsumer(Class<?>, Class<?>)
-          // org.apache.servicecomb.swagger.engine.SwaggerEnvironment.createProducer(Object, Swagger)
-          // had make sure that consumer/swagger or producer/swagger can work
-          //
-          // in this place, do not throw exception when method not exists
-          // eg:
-          //   swagger interface is a.b.c, and consumer interface is a.b.c too.
-          //   version 1, they are the same
-          //   version 2, producer add a new operation, that means swagger have more operation than consumer interface a.b.c
-          //              interface a.b.c in consumer process is the old interface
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
```diff
@@ -83,40 +83,42 @@ public String getPackageName() {
...
 
   private void initOperations() {
-    for (Entry<String, Path> entry : swagger.getPaths().entrySet()) {
-      String strPath = entry.getKey();
-      Path path = entry.getValue();
-      for (Entry<HttpMethod, Operation> operationEntry : path.getOperationMap().entrySet()) {
-        Operation operation = operationEntry.getValue();
-        if (operation.getOperationId() == null) {
-          throw ExceptionUtils.operationIdInvalid(getSchemaId(), strPath);
+    if (swagger.getPaths() != null) {
```
WillemJiang: I think we to write a logger for it, so use can know what's wrong here.

zhengyangyong: I think no necessary, user want fill this empty interface later, so it's not a `wrong scenario`

wujimin: so big block
just return when null

zhengyangyong: good idea, done

zhengyangyong: emmm, ok , add an error output in order to notice user like
`swagger definition for org.apache.servicecomb.samples.uselog4j2.HelloImpl with path /hello is an empty interface, please delete it or fill with one method!`

core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
```diff
@@ -84,6 +84,8 @@ public String getPackageName() {
 
   private void initOperations() {
```
WillemJiang: It just does the check, maybe we can name it as checkOperations.

core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
```diff
@@ -83,6 +83,12 @@ public String getPackageName() {
   }
 
   private void initOperations() {
+    if (swagger.getPaths() == null) {
+      LOGGER.error(swagger.getInfo().getTitle() + " with path " + swagger.getBasePath()
```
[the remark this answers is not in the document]
zhengyangyong: Had considered as warn before, but may mark as error to let user know it is not an appropriate practice and fix as quick as possible ?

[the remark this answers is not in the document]
liubao68: I think we can throw an exception if possible.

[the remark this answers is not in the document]
zhengyangyong: so, we can process this issue two way :
1. throw a ServiceCombException exception, not NPE
2. Give a error log to let user know it is not an appropriate practice and fix as quick as possible, and let service go

This issue come from user report, he want fill interface later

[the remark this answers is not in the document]
liubao68: OK. In this way, we need to change the message to warning.

PR #731, opened 2018-05-28
[SCB-617] Graceful shutdown with standalone tomcat

wujimin:


core/src/main/java/org/apache/servicecomb/core/CseApplicationListener.java
```diff
@@ -71,6 +72,8 @@ public void onApplicationEvent(ApplicationEvent event) {
       }
 
       SCBEngine.getInstance().init();
+    } else if (event instanceof ContextClosedEvent) {
+      SCBEngine.getInstance().uninit();
```
liubao68: I think it's better to use destroy, any ideas？

core/src/main/java/org/apache/servicecomb/core/provider/producer/ProducerProviderManager.java
```diff
@@ -17,39 +17,71 @@
...
+        ((ExecutorService) operationMeta.getExecutor()).shutdown();
+        continue;
+      }
+
+      if (Closeable.class.isInstance(operationMeta.getExecutor())) {
+        IOUtils.closeQuietly((Closeable) operationMeta.getExecutor());
+        continue;
+      }
+
+      LOGGER.warn("do not how to shutdown executor: {}", operationMeta.getExecutor().getClass().getName());
```
liubao68: Executor {} do not support close or shutdown, it may block service shutdown.

core/src/main/java/org/apache/servicecomb/core/executor/FixedThreadExecutor.java
```diff
@@ -74,4 +76,11 @@ private Executor chooseExecutor(long threadId) {
     int idx = index.getAndIncrement() % executorList.size();
     return executorList.get(idx);
   }
+
+  @Override
+  public void close() {
+    for (ExecutorService executorService : executorList) {
+      executorService.shutdown();
+    }
+  }
```
WillemJiang: Do we need to clear the excutorList?

core/src/main/java/org/apache/servicecomb/core/CseApplicationListener.java
wujimin: done
changed all uninit to destroy

core/src/main/java/org/apache/servicecomb/core/provider/producer/ProducerProviderManager.java
wujimin: done

core/src/main/java/org/apache/servicecomb/core/executor/FixedThreadExecutor.java
wujimin: done

core/src/main/java/org/apache/servicecomb/core/provider/producer/ProducerProviderManager.java
```diff
@@ -17,39 +17,72 @@
...
   }
+
+  @Override
+  public void onBootEvent(BootEvent event) {
+    if (!EventType.AFTER_CLOSE.equals(event.getEventType())) {
+      return;
+    }
+
+    for (OperationMeta operationMeta : microserviceMeta.getOperations()) {
+      if (ExecutorService.class.isInstance(operationMeta.getExecutor())) {
```
WillemJiang: What if the excecutor is the instand of ExecutorService and Closeable at the same time?

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
```diff
@@ -109,13 +109,15 @@
...
-  private static ClientPoolManager<HttpClientWithContext> clientMgr;
+  private ClientPoolManager<HttpClientWithContext> clientMgr;
 
   private boolean isWatching = false;
 
-  private static final ServiceLoader<AuthHeaderProvider> authHeaderProviders =
+  private final ServiceLoader<AuthHeaderProvider> authHeaderProviders =
       ServiceLoader.load(AuthHeaderProvider.class);
 
+  private URIConst uriConst = new URIConst();
```
WillemJiang: Do we need to create a instance of Constant object here?

wujimin: not necessary
only use once here

core/src/main/java/org/apache/servicecomb/core/provider/producer/ProducerProviderManager.java
wujimin: 1.ExecutorService is not closeable
2.ExecutorService support shutdown multi times

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
liubao68: After change, it's more convenient for testing different scenarios. We can benefit more :-)

PR #732, opened 2018-05-28
SCB-333 Update to support the date time with JSR-310

zhfeng:

---

### DIFF
--- a/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/TestJsonUtils.java
+++ b/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/TestJsonUtils.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.foundation.common.utils;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.time.Duration;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.Period;
+import java.time.ZoneOffset;
+
+public class TestJsonUtils {
+  @Test
+  public void testJSR310() throws Exception {
+    String period = "\"P1Y1M2D\"";
+    String duration = "\"P1DT1H2M3S\"";
+    String offsetDateTime = "\"2018-05-28T11:00:00+08:00\"";
+    String localDate = "\"2018-05-28\"";
+    String localDateTime = "\"2018-05-28T11:00:00\"";
+
+    Assert.assertEquals(
+        Period.of(1, 1, 2),
+        JsonUtils.OBJ_MAPPER.readValue(period, Period.class)
+    );
+    Assert.assertEquals(
+        Duration.ofDays(1).plusHours(1).plusMinutes(2).plusSeconds(3),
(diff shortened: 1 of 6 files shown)


foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/TestJsonUtils.java
```diff
@@ -0,0 +1,42 @@
...
+public class TestJsonUtils {
+    @Test
+    public void testJSR310() throws Exception {
+        String period = "\"P1Y1M2D\"";
+        String duration = "\"P1DT1H2M3S\"";
+        String offsetDateTime = "\"2018-05-28T11:00:00+08:00\"";
+        String localDate = "\"2018-05-28\"";
+        String localDateTime = "\"2018-05-28T11:00:00\"";
+
+        JsonUtils.OBJ_MAPPER.readValue(period, Period.class);
```
WillemJiang: I think we need to verify the value of object here.

zhfeng: Ok, it makes sense.

wujimin:
1.RestObjectMapper is for RESTful codec, it's better to add UT for it.
2.this is just for codec, i'm not sure if we can generate swagger from JSR-310 dataTypes
  so it's better to add UT and IT for new dataTypes.

zhfeng:
@wujimin I'm not very sure what the UT and the IT means ?

wujimin:
Unit Test
Integration Test
:)

liubao68:
Integrations tests mean we need to add an end-to-end use case in a RPC consumer & provider. You can add examples in demos/SpringMVCclient and SpringMVCserver. 
Because we will use service definition to generate swagger, and you only changed Json mapper, this PR may not work as we expected.

zhfeng:
So I also need to take a look at the swagger generator codes to make sure it can handle the localDate with the JSR310 ?

liubao68:
Yes, I think so. Or users how to use this feature you provided? I can't figure out what is changed after this PR to developers using java-chassis.

imlidian:
I think I have some pojo codes  for this localDate Type testing. Now I give it a try.
https://github.com/microDevOps/microservice-test

WillemJiang:
@imlidian  Please updated the verification that you did.

imlidian:
The Error and my test case implement way as follows. Did I miss something about adding one test case?

Error:
2018-06-05 10:24:19,605 [WARN] consumer method org.apache.servicecomb.demo.springmvc.client.CodeFirstSpringmvcIntf:testObject not exist in contract. org.apache.servicecomb.swagger.engine.SwaggerEnvironment.createConsumer(SwaggerEnvironment.java:98)

Exception in thread "main" java.lang.IllegalStateException: Consumer method org.apache.servicecomb.demo.springmvc.client.CodeFirstSpringmvcIntf:testObject not exist in contract, microserviceName=springmvc, schemaId=codeFirst; new producer not running or not deployed.

Interface Side:
public interface CodeFirstSpringmvcIntf {
 
   Object testObject(Object input);
 
  LocalDate testLocalDate(LocalDate input);

  //LocalDateTime testLocalDateTime(LocalDateTime input);


}
Server Side:
@RestSchema(schemaId = "codeFirst")
 @RequestMapping(path = "/codeFirstSpringmvc", produces = MediaType.APPLICATION_JSON_VALUE)
 public class CodeFirstSpringmvc {
     return input;
   }
  @PostMapping(path = "/testLocalDate")
  public LocalDate testLocalDate(@RequestBody LocalDate input) {
     return input;
   }

Client Side:
b/demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/TestObject.java
 public class TestObject {
     TestMgr.check("str", result);
     TestMgr.check(String.class, result.getClass());
 
    // LocalDate
    result = intf.testLocalDate(LocalDate.of(1968, 12, 8));
    TestMgr.check(LocalDate.of(1968, 12, 8), result);
    TestMgr.check(LocalDate.class, result.getClass());

    result = restTemplate.postForObject(prefix  "/object", "str", String.class);
    TestMgr.check("str", result);
    TestMgr.check(String.class, result.getClass());
}

liubao68:
You can check your provider logs for detail messages. It usually maybe you add some new methods and microservice version not changed and the consumer cannot see the newly added method from service center.  See here for more details: https://huaweicse.github.io/servicecomb-java-chassis-doc/zh_CN/question-and-answer/interface-compatibility.html

liubao68:
We have opened an API to let users to customize RestObjectMapper https://github.com/apache/incubator-servicecomb-java-chassis/pull/874. If users want to use the feature provided by this PR, they can write their own RestObjectMapper. 
I am going to close this PR since it's not a complete feature. Please feel free to add more comments when you have another solutions.

PR #743, opened 2018-05-31
[SCB-601] when java.lang.Error: not support def type: class io.swagger.models.ComposedMode we don't know which swagger

liubao68:

---

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
@@ -57,21 +57,26 @@ public class SchemaMeta extends CommonService<OperationMeta> {
   private SwaggerToClassGenerator swaggerToClassGenerator;
 
   public SchemaMeta(Swagger swagger, MicroserviceMeta microserviceMeta, String schemaId) {
-    this.packageName = SchemaUtils.generatePackageName(microserviceMeta, schemaId);
+    try {
+      this.packageName = SchemaUtils.generatePackageName(microserviceMeta, schemaId);
 
-    this.swagger = swagger;
-    this.name = schemaId;
+      this.swagger = swagger;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/servicecomb/core/definition/SchemaMeta.java
```diff
@@ -57,21 +57,26 @@
...
+      swaggerIntf = swaggerToClassGenerator.convert();
 
-    createOperationMgr("schemaMeta " + schemaId + " operation mgr");
-    operationMgr.setRegisterErrorFmt("Operation name repeat, schema=%s, operation=%s");
+      createOperationMgr("schemaMeta " + schemaId + " operation mgr");
+      operationMgr.setRegisterErrorFmt("Operation name repeat, schema=%s, operation=%s");
 
-    initOperations();
+      initOperations();
+    } catch (Throwable e) {
```
WillemJiang: It's better to log the error with the exception, so we can get the stack trace of it.

liubao68: This will be logger in main thread or by user code. Here do not need, or will print twice

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/SwaggerUtils.java
```diff
@@ -67,6 +71,27 @@ public static Swagger parseSwagger(String swaggerContent) {
     }
   }
 
+  /**
+   * Provide a method to validate swagger. This method is now implemented to check common errors, and the logic
+   * will be changed when necessary. For internal use only.
+   */
+  public static void invalidateSwagger(Swagger swagger) {
```
WillemJiang: the method name should be validateSwagger instead of invalidateSwagger.

PR #744, opened 2018-05-31
[SCB-627] Client Request Timeout support for operation/schema/service level

maheshrajus:

---

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
+++ b/core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
@@ -23,16 +23,13 @@ import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.http.message.BasicNameValuePair;
 import org.apache.servicecomb.core.Const;
 import org.apache.servicecomb.core.Endpoint;
-import org.apache.servicecomb.core.Invocation;
 import org.apache.servicecomb.core.Transport;
-import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
 import org.apache.servicecomb.foundation.common.exceptions.ServiceCombException;
 import org.apache.servicecomb.foundation.common.net.NetUtils;
 import org.apache.servicecomb.foundation.common.net.URIEndpointObject;
@@ -41,7 +38,6 @@ import org.apache.servicecomb.serviceregistry.RegistryUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.netflix.config.DynamicLongProperty;
 import com.netflix.config.DynamicPropertyFactory;
 
 import io.vertx.core.Vertx;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
```diff
@@ -160,4 +161,83 @@ public Object parseAddress(String address) {
...
+    }
+    return DEFAULT_TIMEOUT_MILLIS;
+  }
+
+  /**
+   * Get the configuration value
+   * @param config config parameter
+   * @return long value
+   */
+  private static long getConfigParam(String config, long defaultValue) {
```
liubao68: Maybe you can refer to https://github.com/apache/incubator-servicecomb-java-chassis/blob/master/handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/Configuration.java to make this method simpler. 

And you do not need to cache configure center values, since DynamicPropertyFactory.getInstance().getLongProperty will get the latest value for you.

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
```diff
@@ -160,4 +161,83 @@ public Object parseAddress(String address) {
...
+    DynamicLongProperty dynamicLongProperty = DynamicPropertyFactory.getInstance().getLongProperty(config,
+        defaultValue);
+
+    cfgCallback.computeIfAbsent(config, key -> {
+      dynamicLongProperty.addCallback(() -> {
+        long newValue = dynamicLongProperty.get();
+        String cfgName = dynamicLongProperty.getName();
+
+        //store the value in config center map and check for next requests.
+        configCenterValue.put(cfgName, new AtomicLong(newValue));
```
wujimin: why need Atomic logic?

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
```diff
@@ -160,4 +161,83 @@ public Object parseAddress(String address) {
...
+  }
+
+  /**
+   * Get the configuration value
+   * @param config config parameter
+   * @return long value
+   */
+  private static long getConfigValue(String config) {
+    //first need to check in config center map which has high priority.
+    if (configCenterValue.containsKey(config)) {
```
wujimin: why named "configCenterValue", it's just current value, right?

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
```diff
@@ -160,4 +161,83 @@ public Object parseAddress(String address) {
...
+  public static long getReqTimeout(Invocation invocation) {
+    long value = 0;
+    String config;
+
+    // get the config base on priority. operationName-->schema-->service-->global
+    String operationName = invocation.getOperationName();
+    String schema = invocation.getSchemaId();
+    String serviceName = invocation.getMicroserviceName();
+
+    config = CONSUMER_REQUEST_TIMEOUT + "." + serviceName + "." + schema + "." + operationName;
```
wujimin: try logic for every invocation is not a good idea
it's not better than liubao's suggestion, but more complex.

demo/demo-jaxrs/jaxrs-client/src/main/resources/microservice.yaml
```diff
@@ -32,3 +32,10 @@ servicecomb:
       name: mycustomrule
     retryEnabled: true
     retryHandler: mycustomhandler
+cse:
+  request:
+    timeout:
+      jaxrs:
+        clientreqtimeout:
```
wujimin: how to express jaxrs timeout is x, but jaxrs.clientreqtimeout is y, and jaxrs.clientreqtimeout.add is z

below is not a valid yaml:
```
a: 10
  b: 20
```

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
maheshrajus: ok. we will name accordingly

demo/demo-jaxrs/jaxrs-client/src/main/resources/microservice.yaml
maheshrajus: ok. we will modify the key structure.

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
maheshrajus: we will modify to long. multiple threads will not operate on this map so atomic not required in this case.

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
maheshrajus: done

core/src/main/java/org/apache/servicecomb/core/transport/AbstractTransport.java
maheshrajus: i just read and passed operation name/schema id/service name to the method now. any suggestion on how to get operation name/schema/service name without invocation object ?

PR #746, opened 2018-06-03
[SCB-637] Enhance HttpServletRequestEx impl

wujimin:


foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/StandardHttpServletRequestEx.java
```diff
@@ -79,4 +99,74 @@ public Buffer getBodyBuffer() {
...
+
+  private Map<String, String[]> parseParameterMap() {
+    // 1.post method already parsed by servlet
+    // 2.not APPLICATION_FORM_URLENCODED, no need to enhance
+    if (getMethod().equalsIgnoreCase(HttpMethod.POST)
+        || !StringUtils.startsWithIgnoreCase(getContentType(), MediaType.APPLICATION_FORM_URLENCODED)) {
+      return super.getParameterMap();
+    }
+
+    Map<String, List<String>> listMap = parseUrlEncodedBody();
```
WillemJiang: If the content type is not url encoded, we will consume the input stream, and we cannot read the body content again.
I think the need to check the content type before parse the body.

wujimin: already checked between line 104~107

PR #750, opened 2018-06-05
[SCB-484] Servlet rest support download

wujimin:
did not provide download integration test, but already test locally
because upload did not support for servlet yet,
but it's not good to delete upload test case.


foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/stream/OutputStreamToWriteStream.java
```diff
@@ -0,0 +1,200 @@
...
+      try {
+        outputStream.write(buffer.getBytes());
+
+        synchronized (OutputStreamToWriteStream.this) {
+          currentBufferCount--;
+          Runnable action = (currentBufferCount == 0 && closedDeferred != null) ? closedDeferred : this::checkDrained;
+          action.run();
+        }
+      } catch (IOException e) {
+        currentBufferCount--;
```
WillemJiang: Here is no sychronized.   If we just need to make sure currentBufferCount is automatic, we don't need to use the synchroized to do this job.

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/stream/OutputStreamToWriteStream.java
```diff
@@ -0,0 +1,200 @@
...
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.streams.WriteStream;
+
+/**
+ * for pump from a readStream
+ */
+public class OutputStreamToWriteStream implements WriteStream<Buffer>, AsyncCloseable<Void> {
```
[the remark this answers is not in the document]
wujimin: if readStream and writeStream both are vertx eventloop stream, then it's only one thread
but if one of readStream and writeStream is wrapped from a blocked stream(inputStream/outputStream), then it's multiple thread
so we must use synchronized logic just like vertx asyncFile do.

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/stream/OutputStreamToWriteStream.java
wujimin: not just protect currentBufferCount, we must protect close/drain logic.

PR #755, opened 2018-06-06
[SCB-649] when port can't been listened,log warn and show more detail message

weichao666:

---

### DIFF
--- a/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestTransport.java
+++ b/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestTransport.java
@@ -35,7 +35,7 @@ import io.vertx.core.DeploymentOptions;
 
 @Component
 public class VertxRestTransport extends AbstractTransport {
-  private static final Logger log = LoggerFactory.getLogger(VertxRestTransport.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(VertxRestTransport.class);
 
   private RestTransportClient restClient;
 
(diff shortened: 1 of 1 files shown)


transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/VertxRestTransport.java
```diff
@@ -59,7 +59,7 @@ public boolean canInit() {
     }
 
     if (!NetUtils.canTcpListen(ep.getSocketAddress().getAddress(), ep.getPort())) {
-      log.info("can not listen {}, skip {}.", ep.getSocketAddress(), this.getClass().getName());
+      log.warn("can not listen {}, port may have been occupied, this may cause client report error when being accessed. skip {}.", ep.getSocketAddress(), this.getClass().getName());
```
wujimin: so strange......
canTcpListen is to detect, if this transport can init
when customers deployed vertx and servlet rest transport, we must detect which one will be used
only both of them refused to init is a problem.
we have already print that......

liubao68: We can't detect if the started port is tomcat, so in this scenario(port started by another service, and the dependency contains both transport-rest-servlet and transport-rest-vertx) will cause problem. May the warning message can be changed as: 
WARNING: Can not start VertxRestTransport, the port may have been occupied. You can ignore this message if you are using a web container like tomcat.

weichao666: done

PR #759, opened 2018-06-07
[SCB-653]Transfer-Encoding and Content-Length can not be set to response

liubao68:

---

### DIFF
--- a/demo/demo-jaxrs/jaxrs-client/src/main/resources/microservice.yaml
+++ b/demo/demo-jaxrs/jaxrs-client/src/main/resources/microservice.yaml
@@ -34,12 +34,12 @@ servicecomb:
     retryHandler: mycustomhandler
 cse:
   request:
-    timeout: 5000
+    timeout: 30000
     jaxrs:
-      timeout: 1000
+      timeout: 30000
       clientreqtimeout:
-        timeout: 150
+        timeout: 30000
         sayHello:
-          timeout: 100
+          timeout: 30000
         add:
-          timeout: 100
+          timeout: 2000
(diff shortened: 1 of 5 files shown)


demo/demo-jaxrs/jaxrs-server/src/main/java/org/apache/servicecomb/demo/jaxrs/server/RequestClientTimeOut.java
```diff
@@ -35,7 +35,7 @@
   @Path("/add")
   @POST
   public int add(@FormParam("a") int a, @FormParam("b") int b) throws InterruptedException {
-    Thread.sleep(1000);
+    Thread.sleep(3000);
```
WillemJiang: Do we need to sleep for 3 second?

liubao68: more than 2 seconds. I found time based test cases may not work property if we using time smaller than 1 seconds. So used 1 second as a unit. Integration tests takes more several seconds and I think is acceptable.

demo/demo-jaxrs/jaxrs-client/src/main/resources/microservice.yaml
```diff
@@ -34,12 +34,12 @@ servicecomb:
     retryHandler: mycustomhandler
 cse:
```
WillemJiang: It's better to use servicecomb instead.

PR #761, opened 2018-06-08
[SCB-658] MicroserviceVersions eventbus leak

wujimin:


service-registry/src/test/java/org/apache/servicecomb/serviceregistry/consumer/TestMicroserviceManager.java
```diff
@@ -98,11 +100,23 @@ public void testCreateRuleServiceNotExists() {
...
+
     MicroserviceVersionRule microserviceVersionRule =
         microserviceManager.getOrCreateMicroserviceVersionRule(serviceName, versionRule);
     Assert.assertEquals("0.0.0+", microserviceVersionRule.getVersionRule().getVersionRule());
     Assert.assertNull(microserviceVersionRule.getLatestMicroserviceVersion());
     Assert.assertEquals(0, cachedVersions.size());
+
+    MicroserviceInstanceChangedEvent changedEvent = new MicroserviceInstanceChangedEvent();
+    eventBus.post(changedEvent);
+    Assert.assertEquals(0, triggerCount.get());
```
WillemJiang: Does the changedEvnet trigger the onMicroserviveInstranceChanged method?

wujimin: it can verify the unregister really
if delete the unregister code, then this case will failed.

when invoke microserviceManager.getOrCreateMicroserviceVersionRule
if related MicroserviceVersions is not exist, will create a new instance, the instance will register to eventBus
if some error happend, the created MicroserviceVersions will be removed from microserviceManager automaticlly 
if not unregister from eventBus, then triggerCount  will be 1 not 0

WillemJiang: OK. It makes sense.

PR #767, opened 2018-06-13
[SCB-662]Using cse/servicecomb duplicate key will get list of values

liubao68:

---

### DIFF
--- a/demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
+++ b/demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
@@ -36,6 +36,8 @@ import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.client.RestTemplate;
 
+import com.netflix.config.DynamicPropertyFactory;
+
 public class SpringmvcClient {
   private static RestTemplate templateUrlWithServiceName = new CseRestTemplate();
 
@@ -53,6 +55,8 @@ public class SpringmvcClient {
   }
 
   public static void run() {
+    testConfigurationDuplicate();
+
     templateUrlWithServiceName.setRequestFactory(new UrlWithServiceNameClientHttpRequestFactory());
     restTemplate = RestTemplateBuilder.create();
     controller = BeanUtils.getBean("controller");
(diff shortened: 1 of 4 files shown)


foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -171,7 +171,14 @@ private static void duplicateServiceCombConfigToCse(AbstractConfiguration source
       }
 
       String cseKey = CONFIG_CSE_PREFIX + key.substring(key.indexOf(".") + 1);
-      source.addProperty(cseKey, source.getProperty(key));
+      if (!source.containsKey(cseKey)) {
+        source.addProperty(cseKey, source.getProperty(key));
+      } else {
+        LOGGER
+            .warn(
```
WillemJiang: Not sure if the warning message is enough for this time, if the configuration is wrong, java-chassis may need to stop working instead of sending out the warning message to let the admin know about it.
BTW,  we may need a default override policy here.

liubao68: As shown in test case, this can work. Just when merge, it will not take the cse-config-order(which is implemented by config-cc). We can not cover all orders for ConfigurationsImpl, so give a warning  message. 
If we cover this, we need to change any implementations. This is not a good idea.

liubao68: "stop working" can not be applied because we may configure some default values using cse or servicecomb, if doing this users code have to change to work properly. We can't do this now.

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -171,7 +171,14 @@ private static void duplicateServiceCombConfigToCse(AbstractConfiguration source
       }
 
       String cseKey = CONFIG_CSE_PREFIX + key.substring(key.indexOf(".") + 1);
-      source.addProperty(cseKey, source.getProperty(key));
+      if (!source.containsKey(cseKey)) {
+        source.addProperty(cseKey, source.getProperty(key));
```
wujimin: source.setProperty is better?

liubao68: source.setProperty will clear and add. Here we checked existence and logged a warning message.

wujimin: ok

demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
```diff
@@ -36,6 +36,8 @@
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.client.RestTemplate;
 
+import com.netflix.config.DynamicPropertyFactory;
```
WillemJiang: This import can be removed.

PR #769, opened 2018-06-14
[SCB-667] fix gracefully shutdown is not work in some case

zhengyangyong:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
@@ -108,7 +108,13 @@ public abstract class AbstractRestInvocation {
   }
 
   protected void scheduleInvocation() {
-    createInvocation();
+    try {
+      createInvocation();
+    } catch (IllegalStateException e) {
+      sendFailResponse(e);
+      return;
+    }
+
     invocation.onStart();
     OperationMeta operationMeta = restOperationMeta.getOperationMeta();
 
@@ -161,7 +167,6 @@ public abstract class AbstractRestInvocation {
 
   protected Response prepareInvoke() throws Throwable {
     this.initProduceProcessor();
-
     this.setContext();
     invocation.getHandlerContext().put(RestConst.REST_REQUEST, requestEx);
 
--- a/providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/Invoker.java
+++ b/providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/Invoker.java
@@ -57,6 +57,7 @@ public class Invoker implements InvocationHandler {
 
     // initialized and meta not changed
     public boolean isValid() {
+      SCBEngine.getInstance().ensureStatusUp();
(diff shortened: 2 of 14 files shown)


providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/Invoker.java
```diff
@@ -57,7 +58,10 @@ public InvokerMeta(ReferenceConfig referenceConfig,
 
     // initialized and meta not changed
     public boolean isValid() {
-      return swaggerConsumer != null && microserviceMeta == referenceConfig.getMicroserviceMeta();
+      SCBStatus currentStatus = SCBEngine.getInstance().getStatus();
+      return SCBStatus.UP.equals(currentStatus)
```
wujimin: when shutdown, this will cause createInvokerMeta again, not throw exception

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -160,16 +163,24 @@ public void invoke() {
   }
 
   protected Response prepareInvoke() throws Throwable {
-    this.initProduceProcessor();
+    SCBStatus currentStatus = SCBEngine.getInstance().getStatus();
```
wujimin: why not just invoke ensureStatusUp, or refactor ensureStatusUp, so that no need to write the error message everywhere

wujimin: is it possible check in InvocationFactory.forProvider?
if need to send response, then catch exception for "forProvider"

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -160,16 +163,24 @@ public void invoke() {
...
-        return response;
+      for (HttpServerFilter filter : httpServerFilters) {
+        Response response = filter.afterReceiveRequest(invocation, requestEx);
+        if (response != null) {
+          return response;
+        }
       }
+    } else {
+      LOGGER.warn(
+          "System is starting and not ready for remote calls or shutting down in progress, STATUS = " + currentStatus);
```
WillemJiang: The Request is rejected, as the service provider cannot  process the request due to STATUS =  (you can add the status information here).

providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/Invoker.java
zhengyangyong: Done

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
zhengyangyong: sounds good,Done

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
zhengyangyong: OK,it seems better,Done

PR #770, opened 2018-06-14
[SCB-651] Fixes bug of qps rate limit value

limbojinne:
Added the logic for the first cycle of qps controller so that the number of requests allowed per second equals to its limit value.

---

### DIFF
--- a/handlers/handler-flowcontrol-qps/src/main/java/org/apache/servicecomb/qps/QpsController.java
+++ b/handlers/handler-flowcontrol-qps/src/main/java/org/apache/servicecomb/qps/QpsController.java
@@ -31,12 +31,9 @@ public class QpsController {
   private AtomicLong requestCount = new AtomicLong();
 
   // 本周期之前的请求数
-  private volatile long lastRequestCount = 0;
+  private volatile long lastRequestCount = 1;
 
   private static final int CYCLE_LENGTH = 1000;
-
-  // 是否为第一周期
-  private boolean firstCycle = true;
 
   public QpsController(String key, Integer qpsLimit) {
     this.key = key;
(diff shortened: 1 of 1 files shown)


handlers/handler-flowcontrol-qps/src/main/java/org/apache/servicecomb/qps/QpsController.java
```diff
@@ -57,6 +60,11 @@ public void setQpsLimit(Integer qpsLimit) {
   public boolean isLimitNewRequest() {
     long newCount = requestCount.incrementAndGet();
     long msNow = System.currentTimeMillis();
+    if (firstCycle) {
```
liubao68: Is this solution the same as set lastRequestCount = 1?
In concurrent scenario, flow control is not work very precisely as the configured rate. In practice, we can not assume a concrete rate and will set the limit a bit higher than the expected value. 
So I think this modification is not quit necessary.

limbojinne: Yes, it's same as setting lastRequestCount = 1. I separate the assignment clause since this issue only occurs at the beginning (i.e. in the first cycle), but the following cycles operate just as expected. 
I committed this modification because I was expecting the number of requests allowed per second should be equal to the qps limit value. For example, 5 successful requests when setting limit value as 5.

yangbor: I agree with liubao's opinion. It's impractical to make the actual qps exactly equal to the configured limit. Thus the fix is not needed.

limbojinne: But why only set the limit in the first cycle higher than the expected value at the beginning? After the first cycle, the number of successful requests in the following new cycles will still be equal to the limit value. The modification is supposed to keep the logic consistent both in the first cycle and the following new cycles.

liubao68: If you change lastRequestCount to 1, the algorithm is consistent. Because for next cycles it's initial value is the newCount, not newCount - 1. This can be a minor fix.

limbojinne: It is. So do you want me to keep the code this way or change lastRequestCount to 1?

liubao68: Change lastRequestCount to 1 is preferred. Thanks

PR #774, opened 2018-06-22
[SCB-685] Service comb chassis must support default values in query parameters

acsukesh:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
@@ -33,18 +33,8 @@ public class QueryProcessorCreator implements ParamValueProcessorCreator {
   public static final String PARAMTYPE = "query";
 
   public static class QueryProcessor extends AbstractParamProcessor {
-    Object defaultValue;
-
-    public Object getDefaultValue() {
-      return defaultValue;
-    }
-
-    public void setDefaultValue(Object defaultValue) {
-      this.defaultValue = defaultValue;
-    }
-
-    public QueryProcessor(String paramPath, JavaType targetType) {
-      super(paramPath, targetType);
+    public QueryProcessor(String paramPath, JavaType targetType, Object defaultValue) {
+      super(paramPath, targetType, defaultValue);
     }
 
     @Override
@@ -83,10 +73,6 @@ public class QueryProcessorCreator implements ParamValueProcessorCreator {
   @Override
   public ParamValueProcessor create(Parameter parameter, Type genericParamType) {
     JavaType targetType = TypeFactory.defaultInstance().constructType(genericParamType);
-    QueryProcessor queryProcessor = new QueryProcessor(parameter.getName(), targetType);
-    if (parameter instanceof QueryParameter) {
-      queryProcessor.setDefaultValue(((QueryParameter) parameter).getDefaultValue());
-    }
-    return queryProcessor;
+    return new QueryProcessor(parameter.getName(), targetType, ((QueryParameter) parameter).getDefaultValue());
   }
 }
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
@@ -27,11 +27,22 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 
 import io.swagger.models.parameters.Parameter;
+import io.swagger.models.parameters.QueryParameter;
 
 public class QueryProcessorCreator implements ParamValueProcessorCreator {
   public static final String PARAMTYPE = "query";
 
   public static class QueryProcessor extends AbstractParamProcessor {
+    Object defaultValue;
+
+    public Object getDefaultValue() {
+      return defaultValue;
+    }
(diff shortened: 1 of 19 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
```diff
@@ -43,6 +54,12 @@ public Object getValue(HttpServletRequest request) throws Exception {
         value = request.getParameterValues(paramPath);
       } else {
         value = request.getParameter(paramPath);
+        if (value == null || value.equals("")) {
```
wujimin: 1."" should not change to default value?
2.only process query parameter? what about header/cookie and so on?

samples/springmvc-sample/springmvc-consumer/src/main/java/org/apache/servicecomb/samples/springmvc/consumer/SpringmvcConsumerMain.java
```diff
@@ -51,9 +51,12 @@ public static void main(String[] args) throws Exception {
     // RestTemplate Consumer
     String sayHiResult =
         restTemplate.postForObject("cse://springmvc/springmvchello/sayhi?name=Java Chassis", null, String.class);
+    String sayHiDefaultResult =
```
liubao68: It's better to add test cases in demo projects. Sample are short examples, and demos folder are integration tests, you can add as many test cases you like. 

And maybe you need to test some primitive types default value, and from the following code, they seems not properly handled.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/processor/parameter/AbstractParameterProcessor.java
```diff
@@ -55,6 +57,18 @@ protected void setParameterName(Object annotation, OperationGenerator operationG
...
 
+  protected void setParameterDefaultValue(Object annotation, T parameter) {
+    String defaultValue = getAnnotationParameterDefaultValue(annotation);
+    if (StringUtils.isNotEmpty(defaultValue)) {
+      parameter.setDefaultValue(defaultValue);
+    }
+
+  }
+
+  protected String getAnnotationParameterDefaultValue(Object annotation) {
```
liubao68: primitive and other types default value not properly handled?

swagger/swagger-generator/generator-springmvc/src/main/java/org/apache/servicecomb/swagger/generator/springmvc/processor/annotation/RequestParamAnnotationProcessor.java
```diff
@@ -32,4 +33,13 @@ protected QueryParameter createParameter() {
   protected String getAnnotationParameterName(Object annotation) {
     return ((RequestParam) annotation).name();
   }
+
+  @Override
+  protected String getAnnotationParameterDefaultValue(Object annotation) {
+    String defaultValue = ((RequestParam) annotation).defaultValue();
+    if (defaultValue.equals(ValueConstants.DEFAULT_NONE)) {
+      return "";
```
liubao68: maybe null is better?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
liubao68: Maybe we can add defaultValue to base interface

acsukesh: "" comes only when user inputs the same and is considered as valid input parameter.

swagger/swagger-generator/generator-springmvc/src/main/java/org/apache/servicecomb/swagger/generator/springmvc/processor/annotation/RequestParamAnnotationProcessor.java
acsukesh: might cause NullPointerException, also some message might get added with null which may not be correct , so returning empty

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/QueryProcessorCreator.java
liubao68: OK. For  http://10.57.65.159:8080/codeFirstSpringmvc/reduce?a=
give default Value is better.

swagger/swagger-generator/generator-springmvc/src/main/java/org/apache/servicecomb/swagger/generator/springmvc/processor/annotation/RequestParamAnnotationProcessor.java
liubao68: Maybe it's better to return ValueConstants.DEFAULT_NONE. When no default value specified and users do not give the parameter value, we can throw an BadRequest exception.

liubao68:
I'll merge this PR. Following work need to be done:
1. Support JAX-RS
2. Adding more test cases, to cover default values not specified and user do not parse any value. See review comments: " Maybe it's better to return ValueConstants.DEFAULT_NONE. When no default value specified and users do not give the parameter value, we can throw an BadRequest exception or give default value based on parameter type, e.g. int is 0 and String is null."

PR #781, opened 2018-06-27
[SCB-701] RequestBody(required = false) should invoke method when bod…

acsukesh:
…y is null

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
@@ -45,8 +45,11 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
   public static class BodyProcessor implements ParamValueProcessor {
     protected JavaType targetType;
 
-    public BodyProcessor(JavaType targetType) {
+    protected boolean isRequired;
+
+    public BodyProcessor(JavaType targetType, boolean isRequired) {
       this.targetType = targetType;
+      this.isRequired = isRequired;
     }
 
     @Override
@@ -64,6 +67,9 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
         return null;
       }
 
+      if (isRequired == false && inputStream.available() == 0) {
+        return null;
+      }
       String contentType = request.getContentType();
       if (contentType != null && !contentType.toLowerCase(Locale.US).startsWith(MediaType.APPLICATION_JSON)) {
         // TODO: we should consider body encoding
@@ -77,7 +83,10 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
       try (BufferOutputStream output = new BufferOutputStream()) {
         clientRequest.putHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
         RestObjectMapper.INSTANCE.writeValue(output, arg);
-        clientRequest.write(output.getBuffer());
+        if (arg != null) {
+          clientRequest.write(output.getBuffer());
+        }
+
       }
     }
 
@@ -94,8 +103,8 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
 
   public static class RawJsonBodyProcessor extends BodyProcessor {
 
-    public RawJsonBodyProcessor(JavaType targetType) {
-      super(targetType);
+    public RawJsonBodyProcessor(JavaType targetType, boolean isRequired) {
+      super(targetType, isRequired);
     }
 
     @Override
@@ -130,9 +139,9 @@ public class BodyProcessorCreator implements ParamValueProcessorCreator {
     JavaType targetType = TypeFactory.defaultInstance().constructType(genericParamType);
     boolean rawJson = ClassUtils.isRawJsonType(parameter);
     if (genericParamType.getTypeName().equals(String.class.getTypeName()) && rawJson) {
-      return new RawJsonBodyProcessor(targetType);
+      return new RawJsonBodyProcessor(targetType, parameter.getRequired());
     }
 
-    return new BodyProcessor(targetType);
+    return new BodyProcessor(targetType, parameter.getRequired());
   }
 }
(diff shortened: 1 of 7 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
```diff
@@ -77,7 +83,10 @@ public void setValue(RestClientRequest clientRequest, Object arg) throws Excepti
       try (BufferOutputStream output = new BufferOutputStream()) {
         clientRequest.putHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
         RestObjectMapper.INSTANCE.writeValue(output, arg);
-        clientRequest.write(output.getBuffer());
+        if (arg != null) {
```
liubao68: Can you add a test case to show return body is required while users pass a null value? Json will encoding null to "null", so I think here may got some problem.

common/common-rest/src/test/java/org/apache/servicecomb/common/rest/codec/param/TestBodyProcessor.java
```diff
@@ -61,11 +61,11 @@
   Buffer outputBodyBuffer;
 
   private void createProcessor(Class<?> type) {
-    processor = new BodyProcessor(TypeFactory.defaultInstance().constructType(type));
+    processor = new BodyProcessor(TypeFactory.defaultInstance().constructType(type), false);
```
liubao68: We must not change a test case if it is required. And here should equivalent to true

demo/demo-schema/src/main/resources/microservices/pojo/server.yaml
```diff
@@ -150,7 +150,7 @@ paths:
       parameters:
         - name: TestRequest
           in: body
-          required: true
+          required: false
```
liubao68: We must not change a test case if it is required.  Please add a new one if possible.

demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
```diff
@@ -221,17 +222,33 @@ private static void testConfigurationDuplicate() {
...
+    user.setName("world");
+    TestMgr.check("ha world",
+        template.postForObject(prefix + "/annotations/saysomething?prefix={prefix}",
+            user,
+            String.class,
+            "ha"));
+
+    TestMgr.check("No user data found",
+        template.postForObject(prefix + "/annotations/saysomething?prefix={prefix}",
+            null,
```
liubao68: As mentioned in above comments, "null" may not indicates an empty body. Need to distinguish null or empty.

swagger/swagger-generator/generator-springmvc/src/test/resources/schemas/mixupAnnotations.yaml
```diff
@@ -39,7 +39,7 @@ paths:
       parameters:
       - in: "body"
         name: "srcUser"
-        required: false
+        required: true
```
liubao68: Why change this test case? Can not figure out the related code change.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
acsukesh: its already added. private static void testRequiredBody(RestTemplate template, String microserviceName)   , server code @ AnnotationsTest.java,
TestMgr.check("No user data found",
+        template.postForObject(prefix + "/annotations/saysomething?prefix={prefix}",
+            null,

demo/demo-schema/src/main/resources/microservices/pojo/server.yaml
acsukesh: here the test case is to check null when no body , but if we configure required as true it will always throw exception in client itself as null body not allowed as per swagger. so changed this to false as that is what is expected out of this test case.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
liubao68: I think there are still something missed. Let us view the consumer as a browser or postman.

Assume we have a service:
   void post(Person user)
Before this feature:
post man body: 
1. empty: provider will report invalidate parameter
2. null: (string null): provider will get a null Person instance

After this change:
1. empty: provider will get a null Person
2. null: provider will get a null Person(I am not tested)

And here you add if(arg !=null) statement, 
Before:
consumer send body: null string

Now: consumer send: empty 

This is a change, maybe you need to a manual test using postman. 

For null objects, consumer send empty is better than null, I think, so I can accept this change. But there are one consequence, that when required=true, users can send null string to provider but after change, user can never pass null to Person. 

Can you test and confirm? If my guess is correct, this code can not change, that's consumer always send null to provider for objects. required only work for 3rdparty clients.

demo/demo-springmvc/springmvc-client/src/main/java/org/apache/servicecomb/demo/springmvc/client/SpringmvcClient.java
liubao68: As mentioned in above latest comments, if here can pass "" to user parameter which indicate an empty body? null to indicate a null instance?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
liubao68: Maybe can pass "" as empty body using RestTemplate, but it's a bit strange send "" can not cast to Person.

liubao68: when required = true
1. Resttempalte or RPC always send empty body to provider and provider throws Exception
2. postman can send empty or null to provider, and provider throws Exception
when required = false
1. Resttempalte or RPC always send empty body to provider and provider get null
2. postman can send empty or null to provider, and provider get null

After change this is the expected value. I think is acceptable. When users not specify required = true, we can default to false.

liubao68:
After some considerations about following scenario:
1. generated schema change: before required = false, and now required = true. 
          For this: users can increment microservice version and solve it
2. consume encoding null change: consumer sends null as a parameter
         For this: provider must upgrade SDK
It's very hard to make compatible solutions to this, and the compatible solutions will produce more things to hard to understand. As this two scenario have clear solution and rare, I accept this PR change and notice this change in RN.

liubao68:
I created a new issue to track JAX-RS implementations
https://issues.apache.org/jira/browse/SCB-709

PR #785, opened 2018-06-29
[SCB-705] consumer don't cache ReferenceConfig of an unregistered provider

yhs0092:
So that consumer can re-query provider from the service-center. And when the provider is registered, the invocation can be executed successfully.

---
See details in [SCB-705](https://issues.apache.org/jira/browse/SCB-705)

### DIFF
--- a/core/src/test/java/org/apache/servicecomb/core/provider/consumer/TestConsumerProviderManager.java
+++ b/core/src/test/java/org/apache/servicecomb/core/provider/consumer/TestConsumerProviderManager.java
@@ -17,6 +17,8 @@
 
 package org.apache.servicecomb.core.provider.consumer;
 
+import static org.junit.Assert.fail;
+
 import java.util.Collections;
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/servicecomb/core/provider/consumer/TestConsumerProviderManager.java
```diff
@@ -96,4 +108,42 @@ public void createReferenceConfig_config() {
...
+      }
+
+      @Mock
+      String getMicroserviceName() {
+        return "ms";
+      }
+    };
+
+    try {
+      ReferenceConfig referenceConfig = consumerProviderManager.createReferenceConfig("app:ms");
```
liubao68: unused variable. and if this method not throw exception, test should be failed.

yhs0092: fixed, please review again.

PR #797, opened 2018-07-07
samples 示例中存留的显示契约大部分没用，甚至有可能造成误解。这里对这些契约进行了改进，对无效契约进行提示，并且添加有效契约

heyile:
Signed-off-by: heyile<heyile1@huawei.com>

### DIFF
--- a/samples/jaxrs-sample/jaxrs-provider/src/main/resources/microservices/hello/hello.yaml
+++ b/samples/jaxrs-sample/jaxrs-provider/src/main/resources/microservices/hello/hello.yaml
@@ -14,7 +14,7 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
-
+#这个契约在本示例中没有任何作用
 swagger: '2.0'
 info:
   title: hello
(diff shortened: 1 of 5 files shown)


WillemJiang:
It doesn't help lot if we just add comments on the contract file, for the code first we don't need to specify the contract file.

liubao68:
I think this is a misunderstanding about code-first & contract-first. This is PR is not good to be merged. Many users use code-first style in their development process.

samples/jaxrs-sample/jaxrs-provider/src/main/resources/microservices/hello/hello.yaml
```diff
@@ -14,7 +14,7 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
-
+#这个契约在本示例中没有任何作用
```
liubao68: This file actually effects how the schema registered in service center. If it is exists, will register this, otherwise will generate a new one based on code. Handy yaml may be different than generated one.

wujimin:
any update?

PR #799, opened 2018-07-10
[SCB-725] support get main class package when run with "java -jar xxx.jar

wujimin:


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
```diff
@@ -36,21 +42,50 @@ private JvmUtils() {
...
     if (command == null || command.isEmpty()) {
       return null;
     }
 
-    // command is main class and args
-    String mainClass = command.trim().split(" ")[0];
+    String mainClassOrJar = command.trim().split(" ")[0];
+    String mainClass = readFromJar(mainClassOrJar);
+    if (mainClass == null || mainClass.isEmpty()) {
+      return null;
```
WillemJiang: We may need add a log message here also.

wujimin: already print in "readFromJar"

WillemJiang: OK

PR #800, opened 2018-07-10
[SCB-726] edge support convert from form-data or x-www-form-urlencoded to json automatically

wujimin:


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/BodyProcessorCreator.java
```diff
@@ -70,8 +78,8 @@ public Object getValue(HttpServletRequest request) throws Exception {
       if (isRequired == false && inputStream.available() == 0) {
         return null;
       }
-      String contentType = request.getContentType();
-      if (contentType != null && !contentType.toLowerCase(Locale.US).startsWith(MediaType.APPLICATION_JSON)) {
+
+      if (!contentType.isEmpty() && !contentType.startsWith(MediaType.APPLICATION_JSON)) {
         // TODO: we should consider body encoding
         return IOUtils.toString(inputStream, "UTF-8");
```
WillemJiang: we may need to check the charset in the content-type  for the encoding.

wujimin: this is a old issue, already traced.

PR #803, opened 2018-07-11
 [SCB-706]change filters to discovery filters and support invocation based filters

liubao68:

---


handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServer.java
```diff
@@ -79,29 +69,16 @@ public String getHost() {
...
-
-  public int getCountinuousFailureCount() {
-    return continuousFailureCount.get();
-  }
-
+  // take endpoints that belongs to same instance as same server
   public boolean equals(Object o) {
-    if (o instanceof CseServer) {
-      return this.getHost().equals(((CseServer) o).getHost());
+    if (o instanceof ServiceCombServer) {
```
wujimin: rest endpoint equals highway endpoint?
why?

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/filter/SimpleTransactionControlFilter.java
```diff
@@ -34,23 +36,36 @@
...
+
+  @Override
+  public int getOrder() {
+    return 200;
+  }
+
+  @Override
+  public List<Server> getFilteredListOfServers(List<Server> servers, Invocation invocation) {
     List<Server> filteredServers = new ArrayList<>();
     Map<String, String> filterOptions =
```
wujimin: seems this is about configuration and instances
better to move to DiscoveryFilter?

so that we can cache ability?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/discovery/AbstractDiscoveryFilter.java
```diff
@@ -30,6 +30,9 @@ public DiscoveryTreeNode discovery(DiscoveryContext context, DiscoveryTreeNode p
     }
 
     String childName = findChildName(context, parent);
+    if (childName == null) {
+      return null;
```
wujimin: when will this happened?
after return null, DiscoveryTree will throw exception:

```
// impossible, throw exception to help fix bug
throw new ServiceCombException(filter.getClass().getName() + " discovery return null.");
```

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServer.java
liubao68: DiscoveryFilter sees MicroserviceInstance, not Endpoint. And the health check is also to MicroserviceInstance, not Endpoint.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/filter/SimpleTransactionControlFilter.java
liubao68: This filter leaves for compatible, maybe we need provide a new one in future. So I leave it here but not remove it. We can think about your idea in future when needed.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/Configuration.java
```diff
@@ -173,7 +173,7 @@ public int getRetryOnSame(String microservice) {
   }
 
   public boolean isIsolationFilterOpen(String microservice) {
-    String p = getStringProperty("false",
+    String p = getStringProperty("true",
```
WillemJiang: There are lots of default value changed, not sure if it break the old behavior.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -121,67 +114,40 @@ public void handle(Invocation invocation, AsyncResponse asyncResp) throws Except
...
-      transactionControlFilter.setLoadBalancerStats(lb.getLoadBalancerStats());
-      transactionControlFilter.setMicroserviceName(microserviceName);
-      lb.putFilter(filterName, transactionControlFilter);
-    } catch (Throwable e) {
-      String errMsg = "Fail to create instance of class: " + policyClsName;
-      LOGGER.error(errMsg);
-      throw new Error(errMsg, e);
+    if (!policyClsName.isEmpty()) {
+      LOGGER.error(Configuration.TRANSACTIONCONTROL_POLICY_KEY_PATTERN + " is not supported anymore." +
+          "You can change this class to SPI, and filters will be loaded by SPI.");
```
WillemJiang: We need to provide a document for the user to upgrade the code the new SPI.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/Configuration.java
liubao68: Isolation is not enabled by old version, now we enable it by default and change it's default values. Documents here: https://huaweicse.github.io/servicecomb-java-chassis-doc/zh_CN/references-handlers/loadbalance.html

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
liubao68: Document added : https://huaweicse.github.io/servicecomb-java-chassis-doc/zh_CN/references-handlers/loadbalance.html

PR #804, opened 2018-07-13
[SCB-738] Delete microservice from manager after not exist

wujimin:


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -203,8 +203,18 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
                   mInstances
                       .setInstancesResponse(JsonUtils.readValue(bodyBuffer.getBytes(), FindInstancesResponse.class));
                   break;
+                case 400: {
+                  @SuppressWarnings("unchecked")
+                  Map<String, Object> error = JsonUtils.readValue(bodyBuffer.getBytes(), Map.class);
+                  if ("400012".equals(error.get("errorCode"))) {
+                    mInstances.setMicroserviceNotExist(true);
+                    mInstances.setNeedRefresh(false);
```
WillemJiang: Why do we just set the microservice not exist with specific version number?

wujimin: currently, we only care for this.

PR #809, opened 2018-07-16
[SCB-729] Instance cache check

wujimin:
1.instance cache check task can be trigger periodically or manually
2.after check will post "InstanceCacheSummary" event, subscriber should subscrbe the event, and run their special logic


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/ServiceRegistryTaskInitializer.java
```diff
@@ -1,4 +1,4 @@
-/*
+package org.apache.servicecomb.serviceregistry.registry;/*
```
WillemJiang: The package should add below the License header.

wujimin: oh, this happened randomly
do you know why......

wujimin: done

PR #810, opened 2018-07-16
[SCB-728] Decrypt/signature and adjust form in edge

wujimin:
1.add "decrypt“ logic, decrypt field named "body" in form,  generate new multiple fields to form by "body"
2.move "signature" to edge, and only signature for one operation
3.request with form-data or x-www-form-urlencoded, edge convert to json automatically
4.query userId by serviceToken, and put userId back to form
5.calc signature from response buffer,  and add "signature" field to response body, eg:
  before: {"k":"v"}
  after:    {"k":"v","signature":"xxxxx"}

### DIFF
--- a/demo/demo-edge/consumer/src/main/java/org/apache/servicecomb/demo/edge/consumer/Consumer.java
+++ b/demo/demo-edge/consumer/src/main/java/org/apache/servicecomb/demo/edge/consumer/Consumer.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -43,6 +44,8 @@ import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.Assert;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
 import org.springframework.web.client.RestTemplate;
 
 public class Consumer {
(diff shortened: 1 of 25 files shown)


demo/demo-edge/consumer/src/main/java/org/apache/servicecomb/demo/edge/consumer/Consumer.java
```diff
@@ -111,6 +114,33 @@ public void run(String prefix) {
...
+
+    HttpEntity<MultiValueMap<String, String>> entity = new HttpEntity<>(form, headers);
+
+    @SuppressWarnings("unchecked")
+    Map<String, Object> result = (Map<String, Object>) template.postForObject(url, entity, Map.class);
+    Assert.isTrue(result.containsKey("signature"), "must exist signature");
+    result.remove("signature");
+
+    String expected = "{name=userName, age=10, userId=serviceTokenTest-userId, body1=b1, body2=b2, body3=b3}";
+    System.out.println("encrypt result: " + result.toString());
```
WillemJiang: We are not supposed to use System.out.println in the test.

demo/demo-edge/edge-service/src/main/java/org/apache/servicecomb/demo/edge/service/encrypt/filter/EdgeSignatureResponseFilter.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void beforeSendResponse(Invocation invocation, HttpServletResponseEx responseEx) {
+    EncryptContext encryptContext = (EncryptContext) invocation.getHandlerContext().get(EdgeConst.ENCRYPT_CONTEXT);
+    if (encryptContext == null) {
+      return;
+    }
+    Hcr hcr = encryptContext.getHcr();
+
+    // bad practice: it's better to set signature in response header
```
WillemJiang: Is it a stand request or just some specific user requirement.

demo/demo-edge/edge-service/src/main/resources/META-INF/services/org.apache.servicecomb.common.rest.filter.HttpServerFilter
```diff
@@ -0,0 +1,4 @@
+org.apache.servicecomb.demo.edge.service.encrypt.filter.EdgeSignatureRequestFilter
```
WillemJiang: License header please.

demo/demo-edge/edge-service/src/main/java/org/apache/servicecomb/demo/edge/service/encrypt/filter/EdgeSignatureResponseFilter.java
wujimin: it's a demo for one specific user
it's a demo, not product.

demo/demo-edge/consumer/src/main/java/org/apache/servicecomb/demo/edge/consumer/Consumer.java
wujimin: done

demo/demo-edge/edge-service/src/main/resources/META-INF/services/org.apache.servicecomb.common.rest.filter.HttpServerFilter
wujimin: done

PR #818, opened 2018-07-18
[SCB-755] 755 Duplicate copy cse config to Servicecomb config

heyile:

---

### DIFF
--- a/foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
+++ b/foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
@@ -53,6 +53,7 @@ public final class ConfigUtil {
   private static final Logger LOGGER = LoggerFactory.getLogger(ConfigUtil.class);
 
   private static final String MICROSERVICE_CONFIG_LOADER_KEY = "cse-microservice-config-loader";
+  private static String DEFAULT_SOURCE_NAME = "CSE-DEFAULT-NAME";
 
   private static Map<String, Object> localConfig = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -161,31 +162,42 @@ public static AbstractConfiguration convertEnvVariable(AbstractConfiguration sou
     return source;
   }
 
-  //inject a copy of servicecomb.xxx for cse.xxx
+  //主要是为了兼容老的版本
```
liubao68: Use en comments when possible

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
```diff
@@ -161,31 +162,42 @@ public static AbstractConfiguration convertEnvVariable(AbstractConfiguration sou
...
-      String servicecombKey = CONFIG_SERVICECOMB_PREFIX + key.substring(key.indexOf(".") + 1);
-      if (!source.containsKey(servicecombKey)) {
-        source.addProperty(servicecombKey, source.getProperty(key));
-      } else {
-        LOGGER
-            .warn(
-                "Key {} with an ambiguous item {} exists, it's recommended to use only one of them.",
-                key, servicecombKey);
+  //inject a copy of servicecomb.xxx for cse.xxx
+  private static void duplicateCseConfigToServicecombWithSourceName(AbstractConfiguration source,String sourceName) {
```
liubao68: This parameter is better to use boolean to indicate if need to duplicate (yaml file has already duplicated, so no need to duplicate)

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/ConfigUtil.java
liubao68: If we change this line is better？ 

     duplicateCseConfigToServicecomb(config,
         new DynamicConfiguration(	         new DynamicConfiguration(
             new MicroserviceConfigurationSource(configModelList), new NeverStartPollingScheduler()),	             new MicroserviceConfigurationSource(configModelList), new NeverStartPollingScheduler()),
         "configFromYamlFile");	         "configFromYamlFile");

here do not call duplicateCseConfigToServicecomb and other code remains not changed.

PR #823, opened 2018-07-20
[SCB-762] resolve eclipse compile warnings

wujimin:


wujimin:
CI passed, but show error here, what happened?

WillemJiang:
The CI build just started,  code coverage is updated.

wujimin:
that's because i triggered it again......

PR #851, opened 2018-08-01
 [SCB-782]support revision check when use pull mode with config center

heyile:

---


dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ParseConfigUtils.java
```diff
@@ -40,32 +42,56 @@
 
   public static final Map<String, Object> flatItems = new HashMap<>();
 
+  public static String CURRENT_VERSION_INFO = "default";
```
liubao68: 1. Do not use public static field. 
2. Non final field should be lower case. Only final field be upper case.

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ParseConfigUtils.java
```diff
@@ -37,35 +39,63 @@
...
+    and if we use pull mode and push mode at the same time , we must share a common lock with all methods which would
+    change the config setting
+   */
   public void refreshConfigItems(Map<String, Map<String, Object>> remoteItems) {
-    multiDimensionItems.clear();
-    multiDimensionItems.putAll(remoteItems);
-    doRefreshItems();
-    LOGGER.debug("refresh config success");
+    try {
+      configLock.lock();
```
liubao68: This concurrent control may not work as expected. Because data holding in this class is static, and lock is instance level. 
Anyway, this class only have one instance in runtime in this module, it will cause no problem. It's better we can modify this.

PR #854, opened 2018-08-03
 [SCB-799] Fix ut in different environment

wujimin:


handlers/handler-flowcontrol-qps/src/main/java/org/apache/servicecomb/qps/Config.java
```diff
@@ -37,15 +37,15 @@
...
+  public static Config INSTANCE = new Config();
 
   private final DynamicBooleanProperty consumerEanbled =
       DynamicPropertyFactory.getInstance().getBooleanProperty(CONSUMER_ENABLED, true);
 
   private final DynamicBooleanProperty providerEanbled =
       DynamicPropertyFactory.getInstance().getBooleanProperty(PROVIDER_ENABLED, true);
 
-  private Config() {
+  public Config() {
```
liubao68: Maybe we'd better to make it visible to package and used for test only.

PR #857, opened 2018-08-03
[SCB-802]in rollback scenario, old versions meta is used and invocation is fail

liubao68:
…

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
@@ -18,8 +18,10 @@
 package org.apache.servicecomb.serviceregistry.consumer;
 
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -191,9 +193,30 @@ private void setInstances(List<MicroserviceInstance> pulledInstances, String rev
...
+    Set<String> downSet = new HashSet<>(instances.size());
+    versions.keySet().forEach(item -> {
+      if (!microservices.contains(item)) {
+        downSet.add(item);
+      }
+    });
+    downSet.forEach(item -> {
+      MicroserviceVersion version = versions.remove(item);
+      for (MicroserviceVersionRule microserviceVersionRule : versionRules.values()) {
+        microserviceVersionRule.deleteMicroserviceVersion(version);
```
wujimin: maybe we just need to update latestVersion?

liubao68: In order to update latestVersion, we also have to check which versions are removed. And I think it is good to remove unused version, because some of them may never used again.

wujimin: yes, but that's not delete from versionRule, but need to delete from MicroserviceVersions

PR #862, opened 2018-08-08
[SCB-690] add check if connection is still connected before execute invocation

zhengyangyong:

---

If connection closed before execute invocation, skip invoke it
**Mechanism :**
1. Add an `ConcurrentHashSet` named *connectedAddresses*  in each `Transport`, when consumer connection connected, add its address into it, and after it had close, remove its address from it.
2. before producer execute invocation, it will check whether the consumer address is still in *connectedAddresses*, if no, skip and return

### DIFF
--- a/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/RestServerVerticle.java
+++ b/transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/RestServerVerticle.java
@@ -63,12 +63,16 @@ public class RestServerVerticle extends AbstractVerticle {
 
   private final AtomicInteger connectedCounter;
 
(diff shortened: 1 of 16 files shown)


transports/transport-rest/transport-rest-vertx/src/main/java/org/apache/servicecomb/transport/rest/vertx/RestServerVerticle.java
```diff
@@ -95,17 +99,21 @@ public void start(Future<Void> startFuture) throws Exception {
       HttpServer httpServer = createHttpServer();
       httpServer.requestHandler(mainRouter::accept);
       httpServer.connectionHandler(connection -> {
+        String address = connection.remoteAddress().toString();
```
liubao68: How about client established multi connections to this server?

PR #874, opened 2018-08-16
[SCB-847]Provide a way to decode user's custom error data

liubao68:

---


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestObjectMapperFactory.java
```diff
@@ -17,17 +17,40 @@
...
 public class RestObjectMapperFactory {
+  public static final String KEY_PROVIDER_READ_OR_CONSUMER_READ = "provider.consumer.read";
+
+  public static final String KEY_PROVIDER_WRITE = "provider.write";
+
+  public static final String KEY_CONSUMER_WRITE = "consumer.write";
+
   private static AbstractRestObjectMapper defaultMapper = new RestObjectMapper();
 
+  private static Map<String, AbstractRestObjectMapper> allMappers = new HashMap<>();
```
wujimin: no need to wrap as a map, just two method is enough:
1.getDefaultMapper
2.getConsumerWriteMapper

PR #878, opened 2018-08-20
[SCB-854] BootListener support load by not only spring bean, but also SPI

wujimin:


WillemJiang:
I just checked the code, it provides the SPI loading mechanism. 
Please update the document with some example to show how to use APM with java-chassis.

wujimin:
i already have a APM agent project to show how to trace ServiceComb invocation
i will put it to our samples directory.

we will have ability to get more invocation detail by the enchanced mechanism
image

PR #888, opened 2018-09-03
[SCB-886] Fix path param encode decode problem

yhs0092:

---
See details in [SCB-886](https://issues.apache.org/jira/browse/SCB-886)

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java
@@ -18,13 +18,13 @@
 package org.apache.servicecomb.common.rest.codec.param;
 
 import java.lang.reflect.Type;
-import java.net.URLDecoder;
 import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/base/TestDataTypeJaxrs.java
```diff
@@ -255,4 +252,19 @@ public void testDefault_rt() {
     Integer result = restTemplate.getForObject(urlPrefix, Integer.class);
     assertEquals((int) result, 100);
   }
+
+  @Test
+  public void spaceCharEncode_intf() {
```
wujimin: this file will be deleted in next PR
and seems your new test case is not about datatype, it's about codec of path and query, create a new schema for it?

yhs0092: The tests have been moved into `ParamCodecSchema` and `TestParamCodec`. Please review

PR #889, opened 2018-09-04
[SCB-888] avoid switch SCBEngine status to up in the wrong time

wujimin:


core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
```diff
@@ -192,16 +198,17 @@ public synchronized void init() {
     if (SCBStatus.DOWN.equals(status)) {
       try {
         doInit();
-        status = SCBStatus.UP;
-      } catch (Exception e) {
+        waitStatusUp();
```
liubao68: When will status changed to UP?

wujimin: when got AFTER_REGISTRY event.

PR #890, opened 2018-09-04
[SCB-889] add SCBEngine reference to BootEvent

wujimin:


core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
```diff
@@ -90,6 +90,10 @@ public static SCBEngine getInstance() {
     return INSTANCE;
   }
 
+  public EventBus getEventBus() {
```
liubao68: getEventBus has no reference.

wujimin: add SCBEngine reference to BootEvent
and then customes can get eventBus from scbEngine, no need to use the global variable directly

PR #891, opened 2018-09-04
[SCB-887]aysnc servlet timeout is too short and may block container pool when tasks are timeout

liubao68:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
@@ -124,6 +124,9 @@ public abstract class AbstractRestInvocation {
     operationMeta.getExecutor().execute(() -> {
       synchronized (this.requestEx) {
         try {
+          if (isInQueueTimeout()) {
+            throw new InvocationException(Status.INTERNAL_SERVER_ERROR, "Timeout when processing the request.");
+          }
           if (requestEx.getAttribute(RestConst.REST_REQUEST) != requestEx) {
             // already timeout
             // in this time, request maybe recycled and reused by web container, do not use requestEx
@@ -140,6 +143,18 @@ public abstract class AbstractRestInvocation {
         }
       }
     });
+  }
+
+  private boolean isInQueueTimeout() {
+    Object timeout = requestEx.getAttribute(RestConst.REST_REQUEST_IN_QUEUE_TIME);
+    if (timeout != null) {
+      if (System.currentTimeMillis() - (Long) timeout
+          > CommonRestConfig
+          .getRequestWaitInPoolTimeout()) {
+        return true;
+      }
+    }
+    return false;
   }
 
   protected void runOnExecutor() {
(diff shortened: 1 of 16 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -142,6 +145,18 @@ protected void scheduleInvocation() {
     });
   }
 
+  private boolean isInQueueTimeout() {
+    Object timeout = requestEx.getAttribute(RestConst.REST_REQUEST_IN_QUEUE_TIME);
```
wujimin: 1.direct use invocation.getStartTime()?
is's a nanotime.

2.when invoke from highway, need this logic also

transports/transport-highway/src/main/java/org/apache/servicecomb/transport/highway/HighwayConfig.java
```diff
@@ -18,13 +18,26 @@
...
 import com.netflix.config.DynamicIntProperty;
+import com.netflix.config.DynamicLongProperty;
 import com.netflix.config.DynamicPropertyFactory;
 import com.netflix.config.DynamicStringProperty;
 
 public final class HighwayConfig {
   private HighwayConfig() {
   }
 
+  public static final String KEY_SERVICECOMB_REQUEST_WAIT_IN_POOL_TIMEOUT = "servicecomb.highway.server.requestWaitInPoolTimeout";
```
wujimin: maybe no need to 2 different config item for this?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -142,6 +145,13 @@ protected void scheduleInvocation() {
     });
   }
 
+  private boolean isInQueueTimeout() {
+    if (System.nanoTime() - invocation.getStartTime() > CommonRestConfig.getRequestWaitInPoolTimeout() * 1_000_000) {
```
wujimin: just return directly

transports/transport-rest/transport-rest-servlet/src/main/java/org/apache/servicecomb/transport/rest/servlet/RestAsyncListener.java
```diff
@@ -63,24 +64,28 @@ public void onTimeout(AsyncEvent event) throws IOException {
     // to avoid concurrent, must lock request
     ServletRequest request = event.getSuppliedRequest();
     HttpServletRequestEx requestEx = (HttpServletRequestEx) request.getAttribute(RestConst.REST_REQUEST);
+    LOGGER.error("Rest request timeout, method {}, path {}.", requestEx.getMethod(), requestEx.getRequestURI());
```
wujimin: maybe we can disable async servlet timeout, just control timeout by threadpool timeout?

transports/transport-highway/src/main/java/org/apache/servicecomb/transport/highway/HighwayConfig.java
liubao68: 2 item is more flexible. I test it for highway this configuration works much better than rest.

transports/transport-rest/transport-rest-servlet/src/main/java/org/apache/servicecomb/transport/rest/servlet/RestAsyncListener.java
liubao68: I disabled it with -1 by default. And I think it's better to leave timeout setting to users if they really need it in some scenario.

liubao68:
trigger CI

PR #897, opened 2018-09-07
[SCB-898] Governance function is not effective when configuration Start with servicecomb

weichao666:

---


handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/HystrixCommandPropertiesExt.java
```diff
@@ -178,11 +178,11 @@
   private final HystrixProperty<Boolean> requestCacheEnabled;
 
   protected HystrixCommandPropertiesExt(HystrixCommandKey key) {
-    this(key, HystrixCommandProperties.Setter(), "cse");
+    this(key, HystrixCommandProperties.Setter(), "servicecomb");
```
liubao68: we need to add a test to test this. And we need to test if configure cse.xxx work.

weichao666: i'll add some test later

PR #898, opened 2018-09-08
[SCB-900]Allow number parameter to be quoted in query string

liubao68:

---

### DIFF
--- a/integration-tests/it-consumer/src/main/resources/microservice.yaml
+++ b/integration-tests/it-consumer/src/main/resources/microservice.yaml
@@ -16,4 +16,17 @@
 ## ---------------------------------------------------------------------------
 
 service_description:
+  environment: development
   name: it-consumer
+
+servicecomb:
+  service:
+    registry:
+      instance:
+        healthCheck:
+          interval: 5
+          times: 3
+
+  rest:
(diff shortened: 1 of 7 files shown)


integration-tests/it-consumer/src/main/resources/microservice.yaml
```diff
@@ -16,4 +16,17 @@
 ## ---------------------------------------------------------------------------
 
 service_description:
+  environment: development
```
wujimin: why need this parameter?

integration-tests/it-consumer/src/main/resources/microservice.yaml
```diff
@@ -16,4 +16,17 @@
...
   name: it-consumer
+
```
wujimin: move shared config to it-common?

integration-tests/it-consumer/src/main/resources/microservice.yaml
liubao68: Make local test simpler. I added some test cases, and need to delete sc's data. Add this make local development easier.

liubao68:
Special requirement, not merge it into master.

PR #902, opened 2018-09-11
[SCB-904] SpringMVC annotation do not support value,only support name

weichao666:

---

### DIFF
--- a/integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestAnnotatedAttribute.java
+++ b/integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestAnnotatedAttribute.java
@@ -1,0 +1,139 @@
+package org.apache.servicecomb.it.testcase;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.servicecomb.it.Consumers;
+import org.apache.servicecomb.it.junit.ITJUnitUtils;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+
+public class TestAnnotatedAttribute {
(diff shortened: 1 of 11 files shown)


swagger/swagger-generator/generator-springmvc/src/main/java/org/apache/servicecomb/swagger/generator/springmvc/processor/annotation/RequestHeaderAnnotationProcessor.java
```diff
@@ -32,7 +32,11 @@ public HeaderParameter createParameter() {
 
   @Override
   public String getAnnotationParameterName(Object annotation) {
-    return ((RequestHeader) annotation).name();
+    String name = ((RequestHeader) annotation).name();
+    if (name == null || name.isEmpty()) {
+      name = ((RequestHeader) annotation).value();
```
liubao68: Should consider about RequestParam, CookieValue, etc. And please add integration tests in integration-tests.

wujimin: name will never be null

i'm not sure if it is good to use name() directly 
in very early version, we add similar logic and then delete it
because in that time, some customer use old spring version, that not support name() property.

wujimin: if we decide to add all these new properties, must look at all annotation processors

weichao666: done, springmvc's other annotation has been modified, and read from value first, if it's value is "", read from name

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestAnnotatedAttribute.java
```diff
@@ -0,0 +1,139 @@
+package org.apache.servicecomb.it.testcase;
```
liubao68: license header is missing

weichao666: done

PR #903, opened 2018-09-12
[SCB-906] add sample for invocation apm by java agent

wujimin:


samples/apm-agent/pom.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
liubao68: Is it possible to make this feature a module together with metrics to let users dynamically print logs when needed? e.g. use dynamic configuration.

liubao68: missed license

samples/apm-agent/src/main/resources/META-INF/services/org.apache.servicecomb.core.BootListener
```diff
@@ -0,0 +1 @@
+org.apache.servicecomb.samples.apm.impl.ApmBootListener
```
liubao68: missed license

samples/apm-agent/pom.xml
wujimin: yes, will be done in other PR
old metrics only handled two stage of invocation, we will enhance it.

PR #907, opened 2018-09-14
[SCB-911]Timeout scenario print too many logs

liubao68:

---

### DIFF
--- a/swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/invocation/exception/ExceptionToResponseConverters.java
+++ b/swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/invocation/exception/ExceptionToResponseConverters.java
@@ -33,7 +33,9 @@ public class ExceptionToResponseConverters {
   public ExceptionToResponseConverters() {
     SPIServiceUtils.getSortedService(ExceptionToResponseConverter.class).forEach(converter -> {
       if (converter.getExceptionClass() == null) {
-        defaultConverter = converter;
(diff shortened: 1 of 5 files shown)


swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/invocation/exception/ExceptionToResponseConverters.java
```diff
@@ -42,8 +44,18 @@ public ExceptionToResponseConverters() {
   }
 
   public Response convertExceptionToResponse(SwaggerInvocation swaggerInvocation, Throwable e) {
-    ExceptionToResponseConverter<Throwable> converter =
-        exceptionToResponseConverters.getOrDefault(e.getClass(), defaultConverter);
+    ExceptionToResponseConverter<Throwable> converter = null;
```
wujimin: we can cache the result: Map<Class<?>, ExceptionToResponseConverter<Throwable> >
because the same exception, will always get the same converter

liubao68: It will not improve a lot of performance for doing that. Because class hierarchy will not be that high, 3 is most of the case. Caching too many keys may get worse performance. And error processing is not that performance critical.

wujimin: ok

PR #911, opened 2018-09-17
 [SCB-215] Process @ApiParam

wujimin:
just only process parameter  level, it's enough now.


liubao68:
This PR implements @ApiParam to specify parameter description and collectionFormat will be implemented later.

PR #913, opened 2018-09-17
[SCB-917] parse proto file to model

wujimin:
add protoStuff-parser dependency
upgrade guava/guice version


yangbor:
Please also update the LICENSE/NOTICE files.

wujimin:
sorry, done.

PR #915, opened 2018-09-19
[SCB-926] support invoking 3rd party service

yhs0092:

---
Support invoking 3rd party service just like invoking ServiceComb microservice registered in service-center.

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/CseContext.java
+++ b/core/src/main/java/org/apache/servicecomb/core/CseContext.java
@@ -22,6 +22,7 @@ import javax.inject.Inject;
 import org.apache.servicecomb.core.definition.loader.SchemaListenerManager;
 import org.apache.servicecomb.core.definition.loader.SchemaLoader;
 import org.apache.servicecomb.core.definition.schema.ConsumerSchemaFactory;
+import org.apache.servicecomb.core.definition.schema.StaticSchemaFactory;
 import org.apache.servicecomb.core.provider.consumer.ConsumerProviderManager;
 import org.apache.servicecomb.core.provider.producer.ProducerProviderManager;
 import org.apache.servicecomb.core.transport.TransportManager;
@@ -39,6 +40,8 @@ public class CseContext {
   private SchemaLoader schemaLoader;
 
   private ConsumerSchemaFactory consumerSchemaFactory;
+
+  private StaticSchemaFactory staticSchemaFactory;
 
   private ConsumerProviderManager consumerProviderManager;
 
@@ -76,6 +79,10 @@ public class CseContext {
     return swaggerEnvironment;
   }
 
+  public StaticSchemaFactory getStaticSchemaFactory() {
+    return staticSchemaFactory;
+  }
+
   @Inject
   public void setSwaggerEnvironment(SwaggerEnvironment swaggerEnvironment) {
     this.swaggerEnvironment = swaggerEnvironment;
@@ -109,5 +116,10 @@ public class CseContext {
   @Inject
   public void setTransportManager(TransportManager transportManager) {
     this.transportManager = transportManager;
+  }
+
+  @Inject
+  public void setStaticSchemaFactory(StaticSchemaFactory staticSchemaFactory) {
+    this.staticSchemaFactory = staticSchemaFactory;
   }
 }
--- a/core/src/main/java/org/apache/servicecomb/core/CseContext.java
+++ b/core/src/main/java/org/apache/servicecomb/core/CseContext.java
@@ -22,7 +22,6 @@ import javax.inject.Inject;
 import org.apache.servicecomb.core.definition.loader.SchemaListenerManager;
 import org.apache.servicecomb.core.definition.loader.SchemaLoader;
 import org.apache.servicecomb.core.definition.schema.ConsumerSchemaFactory;
-import org.apache.servicecomb.core.definition.schema.StaticSchemaFactory;
 import org.apache.servicecomb.core.provider.consumer.ConsumerProviderManager;
 import org.apache.servicecomb.core.provider.producer.ProducerProviderManager;
 import org.apache.servicecomb.core.transport.TransportManager;
@@ -40,8 +39,6 @@ public class CseContext {
   private SchemaLoader schemaLoader;
 
   private ConsumerSchemaFactory consumerSchemaFactory;
-
-  private StaticSchemaFactory staticSchemaFactory;
 
   private ConsumerProviderManager consumerProviderManager;
 
@@ -79,10 +76,6 @@ public class CseContext {
     return swaggerEnvironment;
   }
 
-  public StaticSchemaFactory getStaticSchemaFactory() {
-    return staticSchemaFactory;
-  }
-
   @Inject
   public void setSwaggerEnvironment(SwaggerEnvironment swaggerEnvironment) {
     this.swaggerEnvironment = swaggerEnvironment;
@@ -116,10 +109,5 @@ public class CseContext {
   @Inject
   public void setTransportManager(TransportManager transportManager) {
     this.transportManager = transportManager;
-  }
-
-  @Inject
-  public void setStaticSchemaFactory(StaticSchemaFactory staticSchemaFactory) {
-    this.staticSchemaFactory = staticSchemaFactory;
   }
 }
--- a/core/src/main/java/org/apache/servicecomb/core/definition/MicroserviceVersionMeta.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/MicroserviceVersionMeta.java
@@ -20,10 +20,15 @@ package org.apache.servicecomb.core.definition;
 import org.apache.servicecomb.core.CseContext;
 import org.apache.servicecomb.core.definition.classloader.MicroserviceClassLoaderFactory;
 import org.apache.servicecomb.serviceregistry.api.Const;
+import org.apache.servicecomb.serviceregistry.api.registry.Microservice;
 import org.apache.servicecomb.serviceregistry.consumer.MicroserviceVersion;
 
 public class MicroserviceVersionMeta extends MicroserviceVersion {
-  private MicroserviceMeta microserviceMeta;
+  MicroserviceMeta microserviceMeta;
+
+  MicroserviceVersionMeta(Microservice microservice) {
+    super(microservice);
+  }
 
   public MicroserviceVersionMeta(String microserviceName, String microserviceId,
       MicroserviceClassLoaderFactory classLoaderFactory) {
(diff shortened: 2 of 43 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/ServiceRegistry.java
```diff
@@ -59,4 +59,31 @@ MicroserviceInstances findServiceInstances(String appId, String microserviceName
...
+   * can also be used for the proxy interface of RPC style invocation.
+   * @param instances the instances of this 3rd party service. Users only need to specify the endpoint information, other
+   * necessary information will be generate and set in the implementation of this method.
+   */
+  void registryMicroserviceMapping(String microserviceName, String version, Class<?> schemaIntfCls,
+      List<MicroserviceInstance> instances);
+
+  /**
+   * @see #registryMicroserviceMapping(String, String, Class, List)
+   * @param endpoints the endpoints of 3rd party service. Each of endpoints will be treated as a separated instance.
```
wujimin: add example for endpoint?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
```diff
@@ -303,4 +307,30 @@ public void destroy() {
...
+  public void registryMicroserviceMapping(String microserviceName, String version, Class<?> schemaIntfCls,
+      List<MicroserviceInstance> instances) {
+    String app = RegistryUtils.getAppId();
+    MicroserviceManager microserviceManager = appManager.getOrCreateMicroserviceManager(app);
+    StaticMicroserviceVersions microserviceVersions =
+        (StaticMicroserviceVersions) microserviceManager.getVersionsByName()
+            .computeIfAbsent(microserviceName,
+                svcName -> new StaticMicroserviceVersions(this.appManager, app, microserviceName, schemaIntfCls));
+
+    microserviceVersions.addInstances(version, instances);
```
wujimin: addInstances should put inside computeIfAbsent?

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/thirdparty/Test3rdPartyInvocation.java
```diff
@@ -0,0 +1,244 @@
...
+    ListenableFuture<ResponseEntity<String>> stringBodyFuture = cseAsyncRestTemplate
+        .exchange("cse://" + ASYNC_THIRD_PARTY_MICROSERVICE_NAME + "/rest/it-producer/v1/dataTypeJaxrs/stringBody",
+            HttpMethod.POST,
+            new HttpEntity<>("abc"), String.class);
+    ResponseEntity<String> stringBodyResponse = stringBodyFuture.get();
+    Assert.assertEquals(200, stringBodyResponse.getStatusCodeValue());
+    Assert.assertEquals("abc", stringBodyResponse.getBody());
+  }
+
+  @Path("/rest/it-producer/v1/dataTypeJaxrs")
```
wujimin: for different deploy mode, target microservice name is different.
so "it-producer" can not hard code
based on zhengyangyong's PR, if that one merged first, then this one will failed.

core/src/main/java/org/apache/servicecomb/core/CseContext.java
```diff
@@ -40,6 +41,8 @@ public static CseContext getInstance() {
 
   private ConsumerSchemaFactory consumerSchemaFactory;
 
+  private StaticSchemaFactory staticSchemaFactory;
```
wujimin: CseContext will de Deprecated, use ScbEngine

core/src/main/java/org/apache/servicecomb/core/definition/schema/AbstractSchemaFactory.java
```diff
@@ -103,4 +109,12 @@ protected SwaggerGenerator generateSwagger(CONTEXT context) {
 
     return generator;
   }
+
+  protected String getSwaggerContent(Swagger swagger) {
```
wujimin: can we just use SwaggerUtils?

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/thirdparty/Test3rdPartyInvocation.java
```diff
@@ -0,0 +1,244 @@
...
+    int endIndex = urlPrefix.indexOf("/", beginIndex + 3);
+    String endpoint = "rest:" + urlPrefix.substring(beginIndex, endIndex);
+    RegistryUtils.getServiceRegistry()
+        .registryMicroserviceMappingByEndpoints(
+            THIRD_PARTY_MICROSERVICE_NAME, "1.2.1", DataTypeJaxrsSchemaIntf.class,
+            Collections.singletonList(endpoint));
+
+    MicroserviceInstance instance = new MicroserviceInstance();
+    instance.setEndpoints(Collections.singletonList(endpoint));
+    RegistryUtils.getServiceRegistry()
```
liubao68: There are two scenarios need to consider:
1. consumers register producers endpoints and schemas
2. consumers register only schemas,  this is quite common in mesher mode, and used very widely

Can we provide scenario 2 support?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/ServiceRegistry.java
yhs0092: Yes, the example is added in the comment of method `registerMicroserviceMappingByEndpoints`

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/AbstractServiceRegistry.java
yhs0092: `addInstances` is put inside `computeIfAbsent`. Currently only initial instance method is provided. Users cannot re-set instance info of a 3rd party service.

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/thirdparty/Test3rdPartyInvocation.java
yhs0092: I've checked zhengyangyong's commits and test it by rebasing those commits into my branch. There is no problem with the producer's service name because other deploy mode will not trigger this test.
The only problem is the initialization method of `GateRestTemplate` is changed. And I've changed to use new initialization method in new commit.

core/src/main/java/org/apache/servicecomb/core/CseContext.java
yhs0092: Fixed.

core/src/main/java/org/apache/servicecomb/core/definition/schema/AbstractSchemaFactory.java
yhs0092: Fixed.

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/thirdparty/Test3rdPartyInvocation.java
yhs0092: I've asked wujimin's opinion and maybe we can have a new discussion on the scenario that there are  microservices registered in sc with only instance info but no schema.
This issue is planned to provide the basic ability to invoke 3rd party service.

core/src/main/java/org/apache/servicecomb/core/definition/schema/AbstractSchemaFactory.java
```diff
@@ -45,6 +48,8 @@
   @Inject
   protected CompositeSwaggerGeneratorContext compositeSwaggerGeneratorContext;
 
+  private ObjectWriter writer = Yaml.pretty();
```
liubao68: unused code?

core/src/main/java/org/apache/servicecomb/core/definition/MicroserviceVersionMeta.java
```diff
@@ -20,10 +20,15 @@
...
 import org.apache.servicecomb.core.definition.classloader.MicroserviceClassLoaderFactory;
 import org.apache.servicecomb.serviceregistry.api.Const;
+import org.apache.servicecomb.serviceregistry.api.registry.Microservice;
 import org.apache.servicecomb.serviceregistry.consumer.MicroserviceVersion;
 
 public class MicroserviceVersionMeta extends MicroserviceVersion {
-  private MicroserviceMeta microserviceMeta;
+  MicroserviceMeta microserviceMeta;
+
+  MicroserviceVersionMeta(Microservice microservice) {
```
liubao68: microserviceMeta is not initialized, and getMicroserviceMeta may return null and cause NPE.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/ServiceRegistry.java
```diff
@@ -59,4 +59,38 @@ MicroserviceInstances findServiceInstances(String appId, String microserviceName
...
+   * </em>
+   *
+   * @param microserviceName name of the 3rd party service, and this param also specifies the schemaId
+   * @param version version of this 3rd party service
+   * @param schemaIntfCls the producer interface of the service. This interface is used to generate swagger schema and
+   * can also be used for the proxy interface of RPC style invocation.
+   * @param instances the instances of this 3rd party service. Users only need to specify the endpoint information, other
+   * necessary information will be generate and set in the implementation of this method.
+   */
+  void registerMicroserviceMapping(String microserviceName, String version, Class<?> schemaIntfCls,
```
liubao68: It's better to put schemaIntfCls as last parameter, since we can extend it to schemaIntfCls... to support several schemas. I think this is quite normal in many scenarios like register a gateway service as a third party service.

core/src/main/java/org/apache/servicecomb/core/definition/schema/AbstractSchemaFactory.java
yhs0092: Yes, has been fixed.

core/src/main/java/org/apache/servicecomb/core/definition/MicroserviceVersionMeta.java
yhs0092: `microserviceMeta` in `MicroserviceVersionMeta` is initialized in constructors and has package access modifier which ensures that the classes out of its package can only get access to it by getter method. So I think NPE can be avoid since users can only get `microserviceMeta` after it is initialized.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/ServiceRegistry.java
yhs0092: Fixed.

PR #936, opened 2018-09-30
[SCB-737]Avoid null point exception

laijianbin:

---

### DIFF
--- a/providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/schema/PojoProducers.java
+++ b/providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/schema/PojoProducers.java
@@ -63,6 +63,9 @@ public class PojoProducers implements BeanPostProcessor {
     // aop后，新的实例的父类可能是原class，也可能只是个proxy，父类不是原class
     // 所以，需要先取出原class，再取标注
     Class<?> beanCls = BeanUtils.getImplClassFromBean(bean);
+    if(beanCls == null) {
+    	return;
+    }
     RpcSchema rpcSchema = beanCls.getAnnotation(RpcSchema.class);
     if (rpcSchema == null) {
       return;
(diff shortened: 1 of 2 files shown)


providers/provider-pojo/src/main/java/org/apache/servicecomb/provider/pojo/schema/PojoProducers.java
```diff
@@ -63,6 +63,9 @@ protected void processProvider(String beanName, Object bean) {
     // aop后，新的实例的父类可能是原class，也可能只是个proxy，父类不是原class
     // 所以，需要先取出原class，再取标注
     Class<?> beanCls = BeanUtils.getImplClassFromBean(bean);
+    if(beanCls == null) {
```
wujimin: when will this happened?
if happened, just ignore it Silently?

laijianbin: org.springframework.aop.TargetClassAware.getTargetClass(),this method return the target Class, or null if not known. more  detail, reference https://issues.apache.org/jira/browse/SCB-737?filter=-1

wujimin: if got null, maybe will cause to lost schema definition
so need to print a warn log at least.

liubao68: Can you add a example or test case showing how this will happen?

laijianbin: @liubao68 ，you can reference this blog http://3ms.huawei.com/km/groups/3618383/blogs/details/5589591 
To resolve this problem, I just need to add a warn log will be ok?

liubao68: OK, fine. When users defined bean using EnableBatchProcessing, they will not set bean implementation class. So beans annotated with @RestSchema should not do this.

PR #949, opened 2018-10-11
[SCB-957] delete too old protobuf map compatible flag

wujimin:


transports/transport-highway/src/main/java/org/apache/servicecomb/transport/highway/message/LoginRequest.java
```diff
@@ -42,11 +42,12 @@ public static LoginRequest readObject(Buffer bodyBuffer) throws Exception {
...
   private String zipName;
 
+  // no need this flag any more, but tag(3) should be reserved
   // 历史版本中的protoStuff实现的protobuf的map编码与标准的protobuf不兼容
   // 为保持highway的兼容，旧的不兼容编码也要保留
   // 只有LoginRequest/LoginResponse同时为true时，才使用标准protobuf编码
-  @Tag(3)
-  private boolean useProtobufMapCodec;
+  //@Tag(3)
+  //private boolean useProtobufMapCodec;
```
liubao68: Can old version of highway consumer access new version of provider after this change?

wujimin: impossible, maybe one years ago  version used the old format

PR #952, opened 2018-10-13
[SCB-928] support "collection-format" feature on query param

yhs0092:

---
Let users can transport query array param according to "collection-format" specified in swagger schema.


integration-tests/it-producer/src/main/java/org/apache/servicecomb/it/schema/DownloadSchema.java
```diff
@@ -183,7 +185,7 @@ public String getFilename() {
       @ApiResponse(code = 200, response = File.class, message = ""),
   })
   public ResponseEntity<InputStream> netInputStream(String content) throws IOException {
-    URL url = new URL("http://localhost:9000/download/netInputStream?content="
+    URL url = new URL("http://localhost:" + netStreamPort + "/download/netInputStream?content="
```
wujimin: conflict with https://github.com/apache/incubator-servicecomb-java-chassis/pull/946/commits/31b9925a3c45c340489096ccf7d045e8cde755e5

copy that content to this commit or remote this commit, :)

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/converter/property/SwaggerParamCollectionFormat.java
```diff
@@ -57,6 +68,41 @@ public String getSeparator() {
...
+  }
+
+  /**
+   * Join params with {@link #separator}.
+   * Null element will be ignored since {@code null} cannot be described in query array param.
+   *
+   * @return joined params, or return {@code null} if {@code params} is null or all elements of {@code params} are null.
+   */
+  public String joinParam(Collection<?> params) {
+    if (null == params) {
```
wujimin: try org.apache.commons.lang3.StringUtils#join(java.lang.Iterable<?>, char)?

integration-tests/it-producer/src/main/java/org/apache/servicecomb/it/schema/DownloadSchema.java
yhs0092: This commit has been removed. :)

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/converter/property/SwaggerParamCollectionFormat.java
yhs0092: New implementation is committed.

PR #970, opened 2018-10-26
[SCB-982] Compile with show warnings

wujimin:


core/pom.xml
```diff
@@ -69,5 +69,10 @@
       <groupId>org.apache.servicecomb</groupId>
       <artifactId>foundation-test-scaffolding</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-codegen</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
WillemJiang: Why do we need to add the vertx-codegen provide dependence here.

wujimin: vertx-web depend vertx-codegen, but set to optional
we did not use  vertx-codegen, but javac warning that can not find codegen annotation classes

liubao68:
How do we turn on show warnings options?

wujimin:
the  first commit, pom.xml

PR #980, opened 2018-11-05
[SCB-1007] not support CustomGeneric<Map<String,String>>

heyile:

---

### DIFF
--- a/integration-tests/it-common/src/main/java/org/apache/servicecomb/it/schema/User.java
+++ b/integration-tests/it-common/src/main/java/org/apache/servicecomb/it/schema/User.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


integration-tests/it-common/src/main/java/org/apache/servicecomb/it/schema/User.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.it.schema;
+
+import org.apache.servicecomb.foundation.common.utils.JsonUtils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+public class User {
```
wujimin: is it possible to use org.apache.servicecomb.foundation.test.scaffolding.model.User?

integration-tests/it-producer/src/main/java/org/apache/servicecomb/it/schema/GenericSchema.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.apache.servicecomb.provider.rest.common.RestSchema;
+import org.springframework.http.HttpStatus;
+import org.springframework.util.Assert;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RestSchema(schemaId = "generic")
+@RequestMapping(path = "/v1/generic")
```
wujimin: remove test cases from demo?

PR #999, opened 2018-11-21
[SCB-1031] resolve @ApiOperator @Response override wrong, if the prev…

jeho0815:
…ious Response's property has no-null value, latest property will not overrite

---


swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/processor/annotation/AnnotationUtils.java
```diff
@@ -127,7 +127,35 @@ public static void addResponse(Swagger swagger, ApiResponse apiResponse) {
...
+  }
+
+  private static void mergeResponse(Operation operation, ResponseConfig responseConfig) {
+    if (operation.getResponses() == null) {
+      operation.response(responseConfig.getCode(), responseConfig.getResponse());
+      return;
+    }
+    Response response = operation.getResponses().get(String.valueOf(responseConfig.getCode()));
+    if (response != null) {
+      Response targetResp = responseConfig.getResponse();
```
wujimin: 1.targetResp is source, right?
2.logic for (response == null) is simpler, better to process first and then return, and no need write "else"

PR #1013, opened 2018-11-27
[SCB-1047]microservice.yaml  service_description.version support form…

heyile:
…at xxx.xx.xxx.xxx

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
@@ -34,7 +34,7 @@ import org.apache.commons.configuration.Configuration;
 import org.apache.servicecomb.serviceregistry.config.ConfigurePropertyUtils;
 import org.apache.servicecomb.serviceregistry.config.MicroservicePropertiesLoader;
 import org.apache.servicecomb.serviceregistry.definition.MicroserviceDefinition;
-import org.apache.servicecomb.serviceregistry.version.VersionRuleUtils;
+import org.apache.servicecomb.serviceregistry.version.Version;
 
 public class MicroserviceFactory {
   public Microservice create(String appId, String microserviceName) {
@@ -57,7 +57,9 @@ public class MicroserviceFactory {
     String version = configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
         DEFAULT_MICROSERVICE_VERSION);
     //check version format
-    VersionRuleUtils.checkVersionFormat(version);
+    if (!Version.checkVersion(version)) {
+      throw new IllegalStateException(String.format("config service_description.version %s is invalid", version));
+    }
     microservice.setVersion(version);
     setDescription(configuration, microservice);
     microservice.setLevel(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_ROLE_KEY, "FRONT"));
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
@@ -56,10 +56,8 @@ public class MicroserviceFactory {
     microservice.setAppId(configuration.getString(CONFIG_APPLICATION_ID_KEY, DEFAULT_APPLICATION_ID));
     String version = configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
         DEFAULT_MICROSERVICE_VERSION);
-    //check version format
-    if (!Version.checkVersion(version)) {
-      throw new IllegalStateException(String.format("config service_description.version %s is invalid", version));
-    }
+    // just check version format
+    new Version(version);
     microservice.setVersion(version);
     setDescription(configuration, microservice);
     microservice.setLevel(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_ROLE_KEY, "FRONT"));
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
@@ -34,6 +34,7 @@ import org.apache.commons.configuration.Configuration;
 import org.apache.servicecomb.serviceregistry.config.ConfigurePropertyUtils;
 import org.apache.servicecomb.serviceregistry.config.MicroservicePropertiesLoader;
 import org.apache.servicecomb.serviceregistry.definition.MicroserviceDefinition;
+import org.apache.servicecomb.serviceregistry.version.VersionRuleUtils;
 
 public class MicroserviceFactory {
   public Microservice create(String appId, String microserviceName) {
@@ -53,8 +54,11 @@ public class MicroserviceFactory {
     microservice.setServiceName(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_NAME_KEY,
         DEFAULT_MICROSERVICE_NAME));
     microservice.setAppId(configuration.getString(CONFIG_APPLICATION_ID_KEY, DEFAULT_APPLICATION_ID));
-    microservice.setVersion(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
-        DEFAULT_MICROSERVICE_VERSION));
+    String version = configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
+        DEFAULT_MICROSERVICE_VERSION);
+    //check version format
+    VersionRuleUtils.checkVersionFormat(version);
+    microservice.setVersion(version);
     setDescription(configuration, microservice);
     microservice.setLevel(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_ROLE_KEY, "FRONT"));
     microservice.setPaths(ConfigurePropertyUtils.getMicroservicePaths(configuration));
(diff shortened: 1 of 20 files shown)


core/src/test/java/org/apache/servicecomb/core/TestConfig.java
```diff
@@ -53,7 +53,7 @@ public void testConstants() {
     Assert.assertEquals("rest", Const.RESTFUL);
     Assert.assertEquals("", Const.ANY_TRANSPORT);
     Assert.assertEquals("latest", Const.VERSION_RULE_LATEST);
-    Assert.assertEquals("0.0.0+", Const.DEFAULT_VERSION_RULE);
+    Assert.assertEquals("0.0.0.0+", Const.DEFAULT_VERSION_RULE);
```
liubao68: Implement this feature, we must make sure the old version rule works. Can you add new test cases for 4 bit version with change any old test case?

heyile: As the const DefinitionConst#VERSION_RULE_ALL  is changed from "0.0.0+" to "0.0.0.0+",
I change the value of Const.DEFAULT_VERSION_RULE. you means that I shouldn't change the const 
DefinitionConst#VERSION_RULE_ALL ?

liubao68: I think we do not need to change the VERSION_RULE_ALL constant..

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
```diff
@@ -23,35 +23,44 @@
...
 
   private final short major;
 
   private final short minor;
 
   private final short patch;
 
+  private final short build;
+
+  private final boolean versionFour;
```
wujimin: why need a flag versionFour?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
```diff
@@ -53,8 +54,11 @@ private Microservice createMicroserviceFromDefinition(Configuration configuratio
     microservice.setServiceName(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_NAME_KEY,
         DEFAULT_MICROSERVICE_NAME));
     microservice.setAppId(configuration.getString(CONFIG_APPLICATION_ID_KEY, DEFAULT_APPLICATION_ID));
-    microservice.setVersion(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
-        DEFAULT_MICROSERVICE_VERSION));
+    String version = configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
+        DEFAULT_MICROSERVICE_VERSION);
+    //check version format
+    VersionRuleUtils.checkVersionFormat(version);
```
wujimin: versionRule to check version?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/VersionRuleUtils.java
```diff
@@ -54,7 +54,17 @@ public static VersionRule create(String strVersionRule) {
         return versionRule;
       }
     }
+    throw new IllegalStateException("config service_description.version is invalid ");
+  }
 
-    throw new IllegalStateException("never run to here");
+  public static void checkVersionFormat(String strVersionRule) {
+    strVersionRule = strVersionRule.trim();
```
wujimin: named check version, but checking versionRule?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
```diff
@@ -21,37 +21,41 @@
...
     String[] versions = version.split("\\.", -1);
-    if (versions.length > 3) {
+    if (versions.length > 4) {
       throw new IllegalStateException(String.format("Invalid version \"%s\".", version));
     }
 
-    if (versions.length < 3) {
-      versions = (String[]) ArrayUtils.addAll(versions, ZERO);
-    }
+    versions = (String[]) ArrayUtils.addAll(versions, ZERO);
```
wujimin: less then 4, need to concat array

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
```diff
@@ -74,34 +78,52 @@ private short parseNumber(String name, String allVersion, String version) {
...
     this.minor = minor;
     this.patch = patch;
+    this.build = build;
     this.version = combineStringVersion();
     this.numberVersion = combineVersion();
   }
 
   private String combineStringVersion() {
-    return major + "." + minor + "." + patch;
+    StringBuilder stringBuilder = new StringBuilder();
```
wujimin: performance same to add directly

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
```diff
@@ -74,34 +78,52 @@ private short parseNumber(String name, String allVersion, String version) {
...
+    return stringBuilder.toString();
   }
 
+  // 1.0.0 equals 1.0.0.0
   private long combineVersion() {
-    return (long) major << 32 | (long) minor << 16 | (long) patch;
+    return (long) major << 48 | (long) minor << 32 | (long) patch << 16 | (long) build;
   }
 
+  @VisibleForTesting
```
wujimin: why VisibleForTesting?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
```diff
@@ -53,8 +54,13 @@ private Microservice createMicroserviceFromDefinition(Configuration configuratio
     microservice.setServiceName(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_NAME_KEY,
         DEFAULT_MICROSERVICE_NAME));
     microservice.setAppId(configuration.getString(CONFIG_APPLICATION_ID_KEY, DEFAULT_APPLICATION_ID));
-    microservice.setVersion(configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
-        DEFAULT_MICROSERVICE_VERSION));
+    String version = configuration.getString(CONFIG_QUALIFIED_MICROSERVICE_VERSION_KEY,
+        DEFAULT_MICROSERVICE_VERSION);
+    //check version format
+    if (!Version.checkVersion(version)) {
```
wujimin: why not just new a Version instance

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/VersionRuleUtils.java
```diff
@@ -54,7 +54,6 @@ public static VersionRule create(String strVersionRule) {
         return versionRule;
       }
     }
-
-    throw new IllegalStateException("never run to here");
+    throw new IllegalStateException("config service_description.version is invalid ");
```
wujimin: why versionRule relate to "service_description.version"?

service-registry/src/test/java/org/apache/servicecomb/serviceregistry/registry/TestRemoteServiceRegistry.java
```diff
@@ -141,6 +142,13 @@ public void onPullMicroserviceVersionsInstancesEvent(@Injectable ServiceRegistry
       }
     }.getMockInstance();
 
+    new MockUp<Version>() {
+      @Mock
+      boolean checkVersion(String version){
```
wujimin: why need to mock a valid version?

heyile: as now when version is invalid, we can not registe provider

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
heyile: ok

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
heyile: it's only used in Junit

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/VersionRuleUtils.java
heyile: I mistake version and versionRule at first. And forget to delete it

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
wujimin: it's a public class, and have normal getter/setter?
why only can be used in junit?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
heyile: Maybe just check format is better ? the new Version instance will never used

wujimin: but constrcut a version will got the detail of the wrong format message

wujimin: do you want to express version format check logic by two different way?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/version/Version.java
heyile: ok

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/MicroserviceFactory.java
heyile: ok, I see.

PR #1015, opened 2018-11-27
[SCB-1051]when interface set produces=text/plain;charset=utf-8. and c…

heyile:
…onsumers set accept = text/plain,will cause error

---


common/common-rest/src/test/java/org/apache/servicecomb/common/rest/definition/TestRestOperationMeta.java
```diff
@@ -147,6 +147,18 @@ public void testCreateProduceProcessorsTextAndWildcard() {
             MediaType.APPLICATION_JSON + "," + MediaType.APPLICATION_XML + "," + MediaType.WILDCARD));
   }
 
+  @Test
+  public void testCreateProduceProcessorsWithSemicolon() {
```
liubao68: Can you add an integration tests for this? This is an user case, add an integration tests is better.

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestAcceptType.java
```diff
@@ -0,0 +1,110 @@
...
+    String result = textHeader_rt(consumers, MediaType.TEXT_PLAIN_VALUE);
+    Assert.assertEquals("cse", result);
+
+    try {
+      textHeader_rt(consumers, MediaType.APPLICATION_JSON_VALUE);
+      Assert.fail("should throw exception");
+    } catch (InvocationException e) {
+      Assert.assertEquals(406, e.getStatusCode());
+      Assert.assertTrue(e.getMessage().contains("Accept application/json is not supported"));
+    } catch (Exception e) {
```
wujimin: change Exception to Throwable 
and delete below useless empty lines.

heyile: ok

PR #1016, opened 2018-11-28
[SCB-1054]when download file, we should ignore consumer acceptType

heyile:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import javax.servlet.http.Part;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 
@@ -35,11 +36,13 @@ import org.apache.servicecomb.common.rest.definition.path.PathRegExp;
 import org.apache.servicecomb.common.rest.definition.path.URLPathBuilder;
 import org.apache.servicecomb.core.definition.OperationMeta;
 import org.apache.servicecomb.foundation.vertx.http.HttpServletRequestEx;
+import org.apache.servicecomb.swagger.invocation.response.ResponseMeta;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.databind.JavaType;
+
 import io.swagger.models.Operation;
-import io.swagger.models.Response;
 import io.swagger.models.Swagger;
 import io.swagger.models.parameters.Parameter;
 import io.vertx.ext.web.impl.MimeTypesUtils;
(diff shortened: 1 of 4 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
```diff
@@ -105,6 +110,16 @@ public void init(OperationMeta operationMeta) {
     setAbsolutePath(concatPath(swagger.getBasePath(), operationMeta.getOperationPath()));
   }
 
+  private void checkDownloadFile(Operation operation) {
+    try {
+      Response response = operation.getResponses().get("200");
+      downloadFile = response.getSchema().getType().toLowerCase().equals("file");
+    } catch (Exception e) {
```
liubao68: Do not use exception to do flow control. Check null is better.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
```diff
@@ -217,9 +232,14 @@ public ProduceProcessor ensureFindProduceProcessor(String acceptType) {
     if (StringUtils.isEmpty(acceptType)) {
       return defaultProcessor;
     }
-
-    List<String> mimeTyps = MimeTypesUtils.getSortedAcceptableMimeTypes(acceptType.toLowerCase(Locale.US));
-    for (String mime : mimeTyps) {
+    if (downloadFile) {
```
liubao68: Is it better to add it before StringUtils.isEmpty(acceptType)?

heyile: ok, I will change it. actually, it doesn't matter where it's .
As   **produceProcessorMap.putIfAbsent(MediaType.WILDCARD, defaultProcessor);**
so **defaultProcessor**  equals  **produceProcessorMap.get(MediaType.WILDCARD);**

heyile: ok

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
heyile: ok

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/definition/RestOperationMeta.java
```diff
@@ -105,6 +111,24 @@ public void init(OperationMeta operationMeta) {
     setAbsolutePath(concatPath(swagger.getBasePath(), operationMeta.getOperationPath()));
   }
 
+  private void checkDownloadFile(Operation operation) {
```
wujimin: 1.change name to checkDownloadFileFlag
2.OperationMeta already collect ResponseMetas, just invoke org.apache.servicecomb.swagger.invocation.response.ResponsesMeta#findResponseMeta, and determined by org.apache.servicecomb.swagger.invocation.response.ResponseMeta#javaType

PR #1022, opened 2018-12-04
[SCB-1065] when request not contain traceId,should use provider's invocation's traceId

weichao666:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
@@ -103,7 +103,7 @@ public abstract class AbstractRestInvocation {
     @SuppressWarnings("unchecked")
     Map<String, String> cseContext =
         JsonUtils.readValue(strCseContext.getBytes(StandardCharsets.UTF_8), Map.class);
-    invocation.addContext(cseContext);
+    invocation.mergeContext(cseContext);
   }
 
   public String getContext(String key) {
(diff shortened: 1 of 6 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -103,7 +103,7 @@ protected void setContext() throws Exception {
     @SuppressWarnings("unchecked")
     Map<String, String> cseContext =
         JsonUtils.readValue(strCseContext.getBytes(StandardCharsets.UTF_8), Map.class);
-    invocation.setContext(cseContext);
+    invocation.addContext(cseContext);
```
wujimin: highway have the same problem

we can add a new method in invocation: mergeContext
1.if new context have more items, then addAll to new context, and replace old context
2.if new context have less items, then allAll to old context directly.

swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/invocation/context/InvocationContext.java
```diff
@@ -68,6 +68,22 @@ public void addContext(Map<String, String> otherContext) {
...
+  public void mergeContext(InvocationContext otherContext) {
+    mergeContext(otherContext.getContext());
+  }
+
+  public void mergeContext(Map<String, String> otherContext) {
+    if (otherContext == null) {
+      return;
+    }
+    if (otherContext.size() > context.size()) {
+      otherContext.putAll(context);
```
wujimin: 1.loop context and otherContext.putIfAbsent
2.still did not process highway transport?

jeho0815: > 1.loop context and otherContext.putIfAbsent
> 2.still did not process highway transport?

new traceid created in invocation.onStart, then merge context. if only merge absent keys, the traceid is the new one.
can we put the merge context before invocation.onStart?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
weichao666: done

PR #1036, opened 2018-12-13
[SCB-1079]allow consumer-id to be empty when query instance

liubao68:

---


yhs0092:
Hi @liubao68 
As far as I know, we treat service-id of the consumer as the consumer-id. So there seems no possibility that the consumer-id is null, as long as the consumer service instance is registered in service-center properly.
Or the service-id mechanism will also be changed?

liubao68:
@yhs0092 consumer-id is null when the startup procedure is in progress (BEFORE_REGISTRY), the service-id is not registered to service-center or we have not quried service-id from service-center. This feature is for scenarios that users want to do some extra control before register himself. e.g. Registry before querie service center that some dependeded service exists.

yhs0092:
@liubao68   OK, get it.

PR #1049, opened 2019-01-02
[SCB-1101] support IPv6 address format

yhs0092:

---
Currently our IP address parsing logic cannot work properly in IPv6 environment

### DIFF
--- a/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
+++ b/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
@@ -70,6 +70,7 @@ public final class NetUtils {
         hostAddress = availabelAddress.getHostAddress();
         LOGGER.warn("cannot find a proper host address, choose {}, may not be correct.", hostAddress);
       } else {
+        LOGGER.info("get localhost address: {}", localHost.getHostAddress());
         hostAddress = localHost.getHostAddress();
       }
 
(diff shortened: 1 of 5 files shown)


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -58,6 +60,8 @@
     doGetHostNameAndHostAddress();
   }
 
+  static Pattern ipv6Pattern = Pattern.compile("[\\[][0-9a-fA-F:]*:[0-9a-fA-F:]*:.*[]]");
```
wujimin: why not just let system to check?

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/net/NetUtils.java
```diff
@@ -58,6 +60,8 @@
...
+
   private static void doGetHostNameAndHostAddress() {
     try {
       doGetIpv4AddressFromNetworkInterface();
```
wujimin: after support ipv6, we still ignore ipv6 addresses?

yhs0092: Should we take ipv6 address into consideration in common situation? Or we may add a switch to let users to decide?

yhs0092: @liubao68  any idea?

yhs0092:
There seems another problem. I need to check the instance ping mechanism.

PR #1061, opened 2019-01-14
Add "import package" message in README.

bettermanzzy:
In README, we should describe the steps of use in more detailed. I'd prefer to add "import package" message in README, because it may be helpful for new hand.


zenlinTechnofreak:
LGTM.

PR #1078, opened 2019-01-29
 [SCB-1140]容器场景，默认配置服务监听地址0.0.0.0会有安全问题

NevilleYeung:

---


liubao68:
I think this is not a valid issue. You can configure servicecomb.rest.address to the target IP address if you do not want to listen all interfaces. 
e.g. 
  172.16.0.200:8080

NevilleYeung:
Yep,that's right. But if the application runs in a container, IP address is unknownable. Developers have to configure servicecomb.rest.address to 0.0.0.0:8080.

mt-monster:
The program can only choose the one indicated or do it randomly.-----right,so i think better to bind one indicated which is eth0 rather than radom one which is unsafe. so fix this issue is necessary.

NevilleYeung:
Well, I think Mr. Liubao is right, will cancel this pr.

PR #1083, opened 2019-01-30
[SCB-1144]change samples use log4j2

heyile:

---


samples/auth-sample/auth-consumer/pom.xml
```diff
@@ -29,6 +29,12 @@
     <dependency>
       <groupId>org.apache.servicecomb</groupId>
       <artifactId>transport-highway</artifactId>
+      <exclusions>
```
wujimin: is this necessary?

samples/auth-sample/auth-consumer/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,45 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!--this is sample configuration, please modify as your wish-->
+
+<configuration>
+  <Properties>
+    <Property name="log_path">./samples/auth-sample/auth-consumer/target/log/</Property>
```
wujimin: current dir is not auth-consumer?

samples/auth-sample/auth-consumer/pom.xml
heyile: of course. If do not  have this ...
the log is as bellow.
```
log4j:WARN No appenders could be found for logger (org.springframework.core.env.StandardEnvironment).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

samples/auth-sample/auth-consumer/src/main/resources/log4j2.xml
heyile: should I change the log dir ?

liubao68:
Why we need to do this change? Is it OK we add only one sample to show how to use log4j2?

PR #1084, opened 2019-02-01
[SCB-1139] update vertx version from 3.5.3 to 3.6.2

jeho0815:
1、update netty to 4.1.28 ncnative to 2.0.14.Final jackson 2.9.8
2、vertximpl and ConnectionBase(just for test) does not have an public construtor, so add the VertxImpl to resolve first, later will be removed
3、DefaultVertxMetrics does not have the vertx parameter, so add an setter method

---


transports/transport-rest/transport-rest-client/src/test/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -0,0 +1,398 @@
+/*
```
wujimin: must use vertx LICENSE
and why copy this file?

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
```diff
@@ -163,6 +163,7 @@ private void refreshMembers(MemberDiscovery memberDiscovery) {
       String configCenter = memberDiscovery.getConfigServer();
       IpPort ipPort = NetUtils.parseIpPortFromURI(configCenter);
       clientMgr.findThreadBindClientPool().runOnContext(client -> {
+        @SuppressWarnings("deprecation")
```
liubao68: Do not use deprecad api if necessary.

jeho0815: vertx 3.6.3 statement the interface is deprecated, but has not provider a new interface to replace, so we have to add the annotation to paas the CI buiding

transports/transport-rest/transport-rest-client/src/test/java/io/vertx/core/net/impl/ConnectionBase.java
jeho0815: we mock some methods of ConnectionBase have changed to final, but mockito's the old version can't mock the final method, so copy this file.

LICENSE
```diff
@@ -217,7 +217,10 @@ This product bundles files from vertx which is licensed under the Apache License
 For details, see https://github.com/vert-x3/vertx-web
 
 ================================================================
-For  foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/stream/PumpImplEx.java
+For transports/transport-rest/transport-rest-client/src/test/java/io/vertx/core/net/impl/ConnectionBase.java
```
wujimin: no need to declare ConnectionBase.java any more

pom.xml
```diff
@@ -195,6 +195,8 @@
             <exclude>**/target/**</exclude>
             <!-- Skip the source files which are forked from vertx -->
             <exclude>**/io/vertx/ext/web/impl/MimeTypesUtils.java</exclude>
+            <exclude>**/io/vertx/core/net/impl/ConnectionBase.java</exclude>
```
wujimin: no need to declare ConnectionBase.java any more

PR #1087, opened 2019-02-03
[SCB-638] Config model mechanism

wujimin:
support get config value by priority or inject priority value to model
```
  @InjectProperties(prefix = "root")
  public static class ConfigWithAnnotation {
    @InjectProperty(prefix = "override", keys = {"high", "low"})
    public String strValue;

    @InjectProperty(keys = "${key}.value")
    public int intValue;

    @InjectProperty(keys = "${list-1}.a.${list-2}.b")
    public long longValue;

    @InjectProperty(keys = "${full-list}")
    public float floatValue;
}
```

### DIFF
--- a/foundations/foundation-config/src/main/java/org/apache/servicecomb/config/priority/impl/PropertyGetter.java
+++ b/foundations/foundation-config/src/main/java/org/apache/servicecomb/config/priority/impl/PropertyGetter.java
@@ -1,0 +1,125 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.servicecomb.config.priority.impl;
+
+import com.netflix.config.DynamicStringProperty;
+import com.netflix.config.Property;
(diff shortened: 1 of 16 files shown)


foundations/foundation-config/src/main/java/org/apache/servicecomb/config/priority/impl/PropertyGetter.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public Long getValue() {
+      return get();
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  static <T> Property<T> getLongProperty(String propName, T defaultValue) {
+    return (Property<T>) new DynamicLongProperty(propName, (Long) defaultValue);
```
yhs0092: Maybe we can remove the annotation `@SuppressWarnings("unchecked")` like below?
```java
  static Property<Long> getLongProperty(String propName, Long defaultValue) {
    return new DynamicLongProperty(propName, defaultValue);
  }
```

foundations/foundation-config/src/test/java/org/apache/servicecomb/config/inject/TestConfigObjectFactory.java
```diff
@@ -0,0 +1,437 @@
...
+
+  @Test
+  public void placeholder_multi_list() {
+    ConfigWithAnnotation config = new ConfigObjectFactory().create(ConfigWithAnnotation.class,
+        "list-1", Arrays.asList("l1-1", "l1-2"),
+        "list-2", Arrays.asList("l2-1", "l2-2"));
+    // l1-1.a.l2-1.b
+    // l1-2.a.l2-1.b
+    // l1-1.a.l2-2.b
+    // l1-2.a.l2-2.b
```
yhs0092: It seems that the backward placeholders have a higher priority. Does this meets your expectation?

wujimin: it's my expectation, but it's not correct
i'll fix it.

wujimin: done

foundations/foundation-config/src/main/java/org/apache/servicecomb/config/priority/impl/PropertyGetter.java
wujimin: done

PR #1093, opened 2019-02-13
[SCB-1147] Log slow invocation stage trace information

wujimin:
- support global/service/schema/operation level configuration
- support enabled and time configuration
- enabled default to be false
- time default to be 1000 millisecond

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.core.definition;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public static final List<String> PRODUCER_OP_PRIORITY = Arrays.asList(
+      ".${schema}.${operation}",
+      ".${schema}",
+      "");
+
+  @InjectProperty(keys = "${consumer-producer}.invocation.slow.enabled${op-priority}", defaultValue = "false")
+  private boolean slowInvocationEnabled;
+
+  @InjectProperty(keys = {"${consumer-producer}.invocation.slow.msTime${op-priority}"}, defaultValue = "1000")
```
yhs0092: `msTime${op-priority}`
Should there be a dot between `msTime` and `${op-priority}`?
The same question to line 48.

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
```diff
@@ -0,0 +1,76 @@
...
+  public static final List<String> CONSUMER_OP_ANY_RIORITY = Arrays.asList(
+      "${service}.${schema}.${operation}",
+      "${service}.${schema}",
+      "${service}");
+
+  public static final List<String> PRODUCER_OP_ANY_PRIORITY = Arrays.asList(
+      "${schema}.${operation}",
+      "${schema}");
+
+  public static final List<String> CONSUMER_OP_RIORITY = Arrays.asList(
```
yhs0092: There seems a typo. Maybe it should be `CONSUMER_OP_PRIORITY`?
The same problem to line 28

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
wujimin: op-priority reference to CONSUMER_OP_RIORITY or PRODUCER_OP_PRIORITY
already have prefix dot

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
wujimin: done

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
yhs0092: OK, got it

PR #1096, opened 2019-02-15
[SCB-1148] Read transport configuration from model

wujimin:


heyile:
looks like that all things are fine. but do not know the details about annotation `@InjectProperty` and how to use it correctly.

wujimin:
> looks like that all things are fine. but do not know the details about annotation `@InjectProperty` and how to use it correctly.

document task traced by https://issues.apache.org/jira/browse/SCB-1162

yhs0092:
Hello. There seems pretty good about this PR. But I get some other doubts about our new config model mechanism.
1. Is the instance of `ConfigObjectFactory` reusable?
Currently the instance of `ConfigObjectFactory` stores the information about the instance of `PriorityProperty` created by it. Although it's okay to reuse the `ConfigObjectFactory` instance, it still seems a little bit strange if the instance of `ConfigObjectFactory` is reused.  
If we don't want people create multiple different `PriorityProperty` instances by one `ConfigObjectFactory` instance, maybe we can make the `create` method `static`, so we can avoid exposing the `ConfigObjectFactory` instance outside.

2. Is the `PriorityProperty` instances created by `ConfigObjectFactory` garbage collected by JVM?
There seems no reference to the instances of `PriorityProperty`. Will they be garbage collected by JVM unexpectly?

wujimin:
1. ConfigObjectFactory is not support reusable, Creating a configuration module is a small probability event, and create ConfigObjectFactory  is very light, no need to reuse it
2. about GC
- com.netflix.config.PropertyWrapper#prop is referenced by global variable com.netflix.config.DynamicProperty#ALL_PROPS
- com.netflix.config.PropertyWrapper#prop is type of com.netflix.config.DynamicProperty, com.netflix.config.DynamicProperty#callbacks reference callback from org.apache.servicecomb.config.priority.PriorityProperty
- so PriorityProperty will not be GC.

PR #1101, opened 2019-02-21
 [SCB-1156] avoid collect wrong invocation stage data when rest invoke failed

wujimin:


wujimin:
CI maven cache become invalid, close and reopen can not resolve, need to submit another PR, so close this one.

Issue #1104, opened 2019-02-22
生成契约时没有考虑继承泛型接口时生成的桥方法


demonbug:
https://github.com/apache/servicecomb-java-chassis/blob/f3ec3889cedbb3c55be2d1dd60bf646da26ff3a5/swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/SwaggerGenerator.java#L263

`public interface A extends B<C>`
会生成
A.xxx(C)和A.xxx(Object)两个方法，导致契约注册冲突

wujimin:
yes, should filter such methods

wujimin:
do you mind submit a PR for this?

wujimin:
https://issues.apache.org/jira/browse/SCB-1165

wujimin:
do you have a jira account?
i will set "Assignee" to you, and close the task.

https://issues.apache.org/jira/browse/SCB-1165

PR #1105, opened 2019-02-22
[SCB-1165] skip bridge method when generate operation swagger

demonbug:
fix issus #1104


wujimin:
thanks for your PR.

wujimin:
do you have a jira account?
i will set "Assignee" to you, and close the task.

https://issues.apache.org/jira/browse/SCB-1165

PR #1113, opened 2019-03-01
[SCB-1179]Optimize the mainclass auto-discovery logic to cover more scenes.

yangshaoqin1:

---

### DIFF
--- a/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
+++ b/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
@@ -22,6 +22,7 @@ import java.net.URL;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
```diff
@@ -39,24 +40,47 @@ private JvmUtils() {
...
+    if (stackTrace != null && stackTrace.length > 0) {
+      for (StackTraceElement stackTraceElement : stackTrace) {
+        if ("main".equals(stackTraceElement.getMethodName())) {
+          mainClass = stackTraceElement.getClassName();
+          break;
+        }
+      }
+    }
+    try {
+      Class<?> cls = Class.forName(mainClass);
```
liubao68: It's better to check if mainClass is null

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
```diff
@@ -78,6 +102,10 @@ private static String readFromJar(String mainClassOrJar) {
       URL url = new URL(manifestUri);
       try (InputStream inputStream = url.openStream()) {
         Manifest manifest = new Manifest(inputStream);
+        String startClass = manifest.getMainAttributes().getValue("Start-Class");
```
liubao68: Class<?> cls = Class.forName(mainClass);
mainClass may be null

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/JvmUtils.java
yangshaoqin1: In the try code block, it will catch a null pointer exception and print the exception stack and the method returns null. I started to write the same way as you. Later, I saw my colleague's original writing and found that it can work normally. However, an unnecessary exception stack log may be generated.
```Java
try {
      Class<?> cls = Class.forName(mainClass);
      LOGGER.info("Found main class \"{}\" by stackTrace.", mainClass);
      return cls;
    } catch (Throwable e) {
      LOGGER.warn("\"{}\" is not a valid class.", mainClass, e);
      return null;
    }
```

liubao68: Yes. Do not use RuntimeException(e.g. NullPointerException) to do logic control. It's better to check null reference.

PR #1123, opened 2019-03-10
[SCB-1188] View convert download schemas

wujimin:
access microservice instance rest port directly, eg:
http://localhost:8080/inspector/index.html
currently only test with latest chrome and firefox, not support IE/Edge

1.view from instance directly
2.view as swaggger, and can test online
3.view as html
4.download as swagger or html for each schema
5.download all as swagger or html with a zip file

remember status between refresh page
1.active node
2.expanded status
3.schema format

screenshot


inspector/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+  <artifactId>inspector</artifactId>
+  <name>Java Chassis::Inspector</name>
+
+  <repositories>
+    <repository>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>jcenter-releases</id>
```
WillemJiang: Why do we need to use jcenter?

wujimin: swagger2markup released in jcenter
most of it's module exist in central repository, but some module not exist
eg: ch.netzwerg:paleo-core:jar:0.11.0

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/resource/ClassPathStaticResourceHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+import java.net.URL;
+
+import javax.servlet.http.Part;
+
+import org.apache.servicecomb.foundation.common.part.InputStreamPart;
+
+public class ClassPathStaticResourceHandler extends StaticResourceHandler {
+  protected Part findResource(String path) throws IOException {
+    URL url = this.getClass().getClassLoader().getResource(path);
```
liubao68: Is it better to constraint the static contents are located in $jar/webapp ? Or we need to check the file suffix to make sure not download unexpected files.

wujimin: already checked in:
org.apache.servicecomb.common.rest.resource.StaticResourceHandler#handle

findResource is a protected method

liubao68: ok, is it better to rename webroot to static? So that we can migrate from spring boot. Or we can add a configuration for webroot.

wujimin: ResourceHandler is a reuseable class, it's not provide service directly
if someone want to change the directory, can do it manually, eg: 
in org.apache.servicecomb.inspector.internal.InspectorImpl, invoke: resourceHandler.setWebRoot("static/")

inspector/src/main/java/org/apache/servicecomb/inspector/internal/InspectorImpl.java
```diff
@@ -0,0 +1,201 @@
...
+  @GET
+  @ApiResponse(code = 200, message = "", response = File.class)
+  public Response downloadSchemas(@QueryParam("format") SchemaFormat format) throws IOException {
+    if (format == null) {
+      format = SchemaFormat.SWAGGER;
+    }
+
+    // normally, schema will not be too big, just save them in memory temporarily
+    ByteArrayOutputStream os = new ByteArrayOutputStream();
+    ZipOutputStream zos = new ZipOutputStream(os);
```
liubao68: It's better to use try with resouce clause to make resource properly closed. And not throw IOException in REST definition.

wujimin: done
add two new commit

PR #1127, opened 2019-03-12
[SCB-1192] Add metrics prefix to slow invocation configuration

wujimin:


solutions/solution-basic/src/main/resources/microservice.yaml
```diff
@@ -41,8 +41,15 @@ servicecomb:
   accesslog:
     enabled: true
   metrics:
+    window_time: 60000
```
heyile: why not use **windowTime**  instead.  maybe there needs a unified style here

wujimin: not sure which style is better, maybe unify them in the furture

heyile: ok.

PR #1150, opened 2019-03-25
[SCB-1220] consumer side support path param regex

yhs0092:

---
See details in [SCB-1220](https://issues.apache.org/jira/browse/SCB-1220)

### DIFF
--- a/common/common-rest/src/test/java/org/apache/servicecomb/common/rest/definition/path/URLPathBuilderTest.java
+++ b/common/common-rest/src/test/java/org/apache/servicecomb/common/rest/definition/path/URLPathBuilderTest.java
@@ -93,6 +93,22 @@ public class URLPathBuilderTest {
         urlPathBuilder.createRequestPath(args));
   }
 
+  @Test
+  public void testRegexPathParam() throws Exception {
+    Map<String, RestParam> paramMap = new LinkedHashMap<>();
+    addParam("p0", int.class, PathParameter::new, paramMap);
+    addParam("p1", String.class, PathParameter::new, paramMap);
+    addParam("q0", int.class, QueryParameter::new, paramMap);
+    addParam("q1", String.class, QueryParameter::new, paramMap);
+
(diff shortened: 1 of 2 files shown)


common/common-rest/src/test/java/org/apache/servicecomb/common/rest/definition/path/URLPathBuilderTest.java
```diff
@@ -93,14 +93,30 @@ public void testMultiQuery() throws Exception {
...
 
+  @Test
+  public void testRegexPathParam() throws Exception {
+    Map<String, RestParam> paramMap = new LinkedHashMap<>();
+    addParam("p0", int.class, PathParameter::new, paramMap);
+    addParam("p1", String.class, PathParameter::new, paramMap);
+    addParam("q0", int.class, QueryParameter::new, paramMap);
+    addParam("q1", String.class, QueryParameter::new, paramMap);
+
+    URLPathBuilder urlPathBuilder = new URLPathBuilder("/path/{p0 : .*}/and/{p1:.*}", paramMap);
```
heyile: what's the means of support regex. do not need check if the value meets the requirements of regex  ?
``` 
URLPathBuilder urlPathBuilder = new URLPathBuilder("/path/{p0 : .*}/and/{p1:[0-10]*}", paramMap);
```

yhs0092: Hello, here is the description about the problem backgroud info: [SCB-1220](https://issues.apache.org/jira/browse/SCB-1220).
Regex path param allows us to specify a more detailed matching rule of request url. Such kind of usage has been applied in our [InspectorImpl.java](https://github.com/apache/servicecomb-java-chassis/blob/master/inspector/src/main/java/org/apache/servicecomb/inspector/internal/InspectorImpl.java).

heyile: you means that **support path param regex**  do not need check param is valid for regex ?
for example 
the url is : **/path/{p0 : [0-9]*}**
what happened if I give a param "abc" which is not valid?

PR #1152, opened 2019-03-26
[SCB-1222] change groupId of "hibernate-validator" from "org.hibernate" to "org.hibernate.validator" to avoid dependency warning

wujimin:


java-chassis-dependencies/default/pom.xml
```diff
@@ -344,6 +349,11 @@
         <artifactId>hibernate-validator-annotation-processor</artifactId>
         <version>6.0.2.Final</version>
       </dependency>
+      <dependency>
+        <groupId>org.hibernate.validator</groupId>
```
liubao68: Shall we delete groupId>org.hibernate</groupId> above?

wujimin: if delete the old one, maybe will cause some problem for old project

PR #1161, opened 2019-03-31
[SCB-1232] make GroupExecutor configuration compatible to old version

wujimin:
expect to work like tomcat thread pool
but jdk standard thread pool's behavior is different:
```
JDK standard thread pool rules:
  1.use core threads.
  2.if all core threads are busy, then queue the request.
  3.if queue is full, then create new thread util reach the limit of max threads.
  4.if queue is full, and threads count is max, then reject the request.
```

to do what we want, need to lock the executor several times for one task, that's too bad

so only make compatible to old version temporary

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/executor/GroupExecutor.java
+++ b/core/src/main/java/org/apache/servicecomb/core/executor/GroupExecutor.java
@@ -40,7 +40,7 @@ public class GroupExecutor implements Executor, Closeable {
   public static final String KEY_GROUP = "servicecomb.executor.default.group";
 
   // Deprecated
-  public static final String KEY_THREAD = "servicecomb.executor.default.thread-per-group";
+  public static final String KEY_OLD_MAX_THREAD = "servicecomb.executor.default.thread-per-group";
 
   public static final String KEY_CORE_THREADS = "servicecomb.executor.default.coreThreads-per-group";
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/servicecomb/core/executor/TestGroupExecutor.java
```diff
@@ -86,46 +75,83 @@ public void maxQueueSize() {
...
     ArchaiusUtils.setProperty(GroupExecutor.KEY_MAX_THREADS, 10);
 
-    LogCollector collector = new LogCollector();
     groupExecutor.initConfig();
     Assert.assertEquals(10, groupExecutor.maxThreads);
-    Assert.assertEquals(
-        "coreThreads is bigger than maxThreads, change from 25 to 10.",
-        collector.getEvents().get(collector.getEvents().size() - 2).getMessage());
-    collector.teardown();
+    Assert.assertEquals(10, groupExecutor.maxThreads);
```
yhs0092: Hello, in the UTs below, the `groupExecutor.maxThreads` is checked twice but the `groupExecutor.maxThreads` is not checked.

core/src/main/java/org/apache/servicecomb/core/executor/GroupExecutor.java
```diff
@@ -106,6 +108,11 @@ public void initConfig() {
     LOGGER.info(
         "executor group={}. per group settings, coreThreads={}, maxThreads={}, maxIdleInSecond={}, maxQueueSize={}.",
         groupCount, coreThreads, maxThreads, maxIdleInSecond, maxQueueSize);
+    LOGGER.info("JDK standard thread pool rules:\n"
+        + "1.use core threads.\n"
+        + "2.if all core threads are busy, then queue the request.\n"
+        + "3.if queue is full, then create new thread util reach the limit of max threads.\n"
+        + "4.if queue is full, and threads count is max, then reject the request.");
```
yhs0092: Do we change the stratege to create new thread first until the limit of max threads is reached. And then queue the requests?

wujimin: yes, i have wrote the reason in the PR comment
and i will copy it to jira task comment

core/src/test/java/org/apache/servicecomb/core/executor/TestGroupExecutor.java
wujimin: done
rewrite the compatible rule:
1.maxThreads always be the max value of (oldMax, core, newMax), if still no value, then default to 100
2.if not configured queue, then coreThreads same to maxThreads
3.if configured queue, and queue size is not max int, then coreThreads default to be 25

core/src/main/java/org/apache/servicecomb/core/executor/GroupExecutor.java
```diff
@@ -83,25 +83,31 @@ public void init() {
...
+    maxThreads = DynamicPropertyFactory.getInstance().getIntProperty(KEY_MAX_THREADS, oldMaxThreads).get();
+    maxThreads = Math.max(coreThreads, maxThreads);
+    maxThreads = maxThreads <= 0 ? 100 : maxThreads;
+
+    maxQueueSize = DynamicPropertyFactory.getInstance().getIntProperty(KEY_MAX_QUEUE_SIZE, Integer.MAX_VALUE).get();
+    if (maxQueueSize == Integer.MAX_VALUE) {
       coreThreads = maxThreads;
+      LOGGER.info("not configured {},  make coreThreads and maxThreads to be {}.", KEY_MAX_QUEUE_SIZE, maxThreads);
+    } else {
+      coreThreads = coreThreads <= 0 ? 25 : coreThreads;
```
liubao68: This statement should be calculated before 
            maxThreads = Math.max(coreThreads, maxThreads);

core/src/test/java/org/apache/servicecomb/core/executor/TestGroupExecutor.java
liubao68: Why you keep this:
  3.if configured queue, and queue size is not max int, then coreThreads default to be 25

I can't figure out the relationship between queue size and max thread size. For example, if users configure queue size to be MAX_INT/2

wujimin: because of jdk thread pool rule and compatible to old version

if upgrade from old version, no queueSize configured, it will be max int, if queueSize is too big, then coreThreads different to maxThreads is no meaning
but we can not know how big is "too big"

if developers configured queueSize， we can only treat the developer know what happened, just use the existing value.

core/src/main/java/org/apache/servicecomb/core/executor/GroupExecutor.java
wujimin: only developers configured a small maxThreads and big coreThreads will cause problem: ThreadPoolExecutor will throw exception in constructor

i think it's developers's duty to make a right configuration
we just handle compatible sense and default value.

PR #1164, opened 2019-04-02
[SCB-1236] bugfix: BufferInputStream#read return value range should b…

jeho0815:
…e -1~255, not the current -128-127

---


transports/transport-highway/src/test/java/org/apache/servicecomb/transport/highway/TestHighwayTransport.java
```diff
@@ -51,6 +52,8 @@ public void testInit() {
       status = false;
     }
     Assert.assertTrue(status);
+
+    VertxUtils.blockCloseVertxByName("transport");
```
WillemJiang: I think we could change the code to put the blockCloseVertxByName code into finally block to close vertx even there is an exception is thrown.
  ```
 } finally {
      VertxUtils.blockCloseVertxByName("transport");
 }   
```

wujimin: not a problem?
because UT failed, CI will terminate at once

PR #1171, opened 2019-04-10
[SCB-1228][WIP][WEAK] swagger generator springmvc not depend on dynamic class

wujimin:


swagger/swagger-generator/generator-springmvc/src/main/java/org/apache/servicecomb/swagger/generator/springmvc/SpringmvcOperationGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  protected void initMethodParameterGenerators(Map<String, List<Annotation>> methodAnnotationMap) {
+    super.initMethodParameterGenerators(methodAnnotationMap);
+
+    parameterGenerators.stream()
+        .filter(pg -> pg.getHttpParameterType() == null)
+        .forEach(pg -> {
+          pg.setHttpParameterType(HttpParameterType.query);
+          pg.setGeneratedParameter(createParameter(HttpParameterType.query));
```
liubao68: Maybe use upper case for enum and give a lower case literal value is better.

wujimin: ok
create a new task for it, because need to modify core/jaxrs/springmvc generators:
https://issues.apache.org/jira/browse/SCB-1252

PR #1172, opened 2019-04-10
[SCB-1246] EventBus subscriber support order

wujimin:


wujimin:
> It seems OK for me. But is this feature really needed? I agree it's indeed an enhancement, but in terms of design patterns, it's not a good idea to take `order` into consideration.

this feature is for 2.0.0, based on https://github.com/apache/servicecomb-java-chassis/pull/1173
MicroserviceMeta managed by SC client
but core/PojoProvider/RestProvider should create additional metas, the modules will subscribe the create event, and they have order dependency.

yhs0092:
OK, I get it.

PR #1196, opened 2019-04-30
[SCB-1274]when service center is unavailable, service should enter sa…

heyile:
…fe mode.

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
@@ -17,11 +17,11 @@
 
 package org.apache.servicecomb.serviceregistry.consumer;
 
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.stream.Collectors;
 
 import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
 import org.apache.servicecomb.foundation.common.utils.SPIServiceUtils;
@@ -34,6 +34,7 @@ import org.apache.servicecomb.serviceregistry.config.ServiceRegistryConfig;
 import org.apache.servicecomb.serviceregistry.definition.DefinitionConst;
 import org.apache.servicecomb.serviceregistry.task.event.MicroserviceNotExistEvent;
 import org.apache.servicecomb.serviceregistry.task.event.PullMicroserviceVersionsInstancesEvent;
+import org.apache.servicecomb.serviceregistry.task.event.SafeModeChangeEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,6 +53,9 @@ public class MicroserviceVersions {
   private String revision = null;
 
   private List<MicroserviceInstance> pulledInstances;
+
+  // in safe mode, instances will never be deleted
+  private boolean safeMode = false;
 
   // instances not always equals to pulledInstances
   // in the future:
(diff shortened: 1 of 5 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/ServiceCenterTask.java
```diff
@@ -30,16 +34,25 @@
 
   private int interval;
 
+  private int checkTimes;
+
+  private AtomicLong failedTimes = new AtomicLong();
```
servicestage-samples: this propertie should be named as consecutiveXXX

heyile: done

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/consumer/MicroserviceVersions.java
```diff
@@ -193,20 +197,26 @@ private void setInstances(List<MicroserviceInstance> pulledInstances, String rev
 
   private List<MicroserviceInstance> mergeInstances(List<MicroserviceInstance> pulledInstances,
       List<MicroserviceInstance> inUseInstances) {
-    List<MicroserviceInstance> upInstances = pulledInstances
-        .stream()
-        .collect(Collectors.toList());
+    List<MicroserviceInstance> upInstances = new ArrayList<>(pulledInstances);
```
liubao68: Can you describe why we need to add this feature? I think this logic is quite a bit confusing. 1. when service center is down, we can't get any instances; 2. service center status has no relationship with microservice instance status. Now no instances will beremoved when no instances queried from service center.

heyile: when service center status changes from down to up. It may take some time for provider to registe itself in the service center. so before provider registes itself, the instances get from service center is wrong.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/ServiceCenterTask.java
```diff
@@ -62,7 +75,27 @@ public void onMicroserviceInstanceHeartbeatTask(MicroserviceInstanceHeartbeatTas
     if (task.getHeartbeatResult() != HeartbeatResult.SUCCESS) {
       LOGGER.info("read MicroserviceInstanceHeartbeatTask status is {}", task.taskStatus);
       onException();
+      if (!safeMode && consecutiveFailedTimes.incrementAndGet() > checkTimes) {
+        LOGGER.warn("service center is unavailable, enter safe mode");
+        eventBus.post(new SafeModeChangeEvent(true));
+      }
+      if (consecutiveSucceededTimes.get() != 0) {
```
liubao68: directly use consecutiveSucceededTimes.set(0) is simpler.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/ServiceCenterTask.java
```diff
@@ -62,7 +75,27 @@ public void onMicroserviceInstanceHeartbeatTask(MicroserviceInstanceHeartbeatTas
...
+      LOGGER.warn("service center is recovery, exit safe mode");
+      eventBus.post(new SafeModeChangeEvent(false));
     }
+    if (consecutiveFailedTimes.get() != 0) {
+      consecutiveFailedTimes.set(0);
+    }
+  }
+
+  @Subscribe
+  public void onSafeModeChanged(SafeModeChangeEvent modeChangeEvent) {
```
liubao68: It's simpler to set safeMode=true before sending event, no need to subscript event here.

heyile: done

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/ServiceCenterTask.java
heyile: done

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/ServiceCenterTask.java
heyile: oh, yes. I agree with you

PR #1209, opened 2019-05-17
[SCB-1278] Not send x-cse* header to third party service by default

yhs0092:

---
See details in [SCB-1278](https://issues.apache.org/jira/browse/SCB-1278)

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
@@ -62,6 +62,13 @@ public class OperationConfig {
   private long msRequestTimeout;
 
   /**
+   * whether to remove certain headers from the 3rd party invocations
+   */
+  @InjectProperty(keys = {"request.${op-any-priority}.clientRequestHeaderFilterEnabled",
+      "request.clientRequestHeaderFilterEnabled"}, defaultValue = "true")
+  private boolean clientRequestHeaderFilterEnabled = true;
+
+  /**
    * producer wait in thread pool timeout
    */
   @InjectProperty(keys = {
@@ -131,5 +138,13 @@ public class OperationConfig {
 
   public long getNanoRestRequestWaitInPoolTimeout() {
     return nanoRestRequestWaitInPoolTimeout;
+  }
+
+  public boolean isClientRequestHeaderFilterEnabled() {
+    return clientRequestHeaderFilterEnabled;
+  }
+
+  public void setClientRequestHeaderFilterEnabled(boolean clientRequestHeaderFilterEnabled) {
+    this.clientRequestHeaderFilterEnabled = clientRequestHeaderFilterEnabled;
   }
 }
--- a/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
+++ b/core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
@@ -64,8 +64,7 @@ public class OperationConfig {
   /**
    * whether to remove certain headers from the 3rd party invocations
    */
-  @InjectProperty(keys = {"request.${op-any-priority}.clientRequestHeaderFilterEnabled",
-      "request.clientRequestHeaderFilterEnabled"}, defaultValue = "true")
+  @InjectProperty(keys = {"request.clientRequestHeaderFilterEnabled${consumer-op-priority}"}, defaultValue = "true")
   private boolean clientRequestHeaderFilterEnabled = true;
 
   /**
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/apache/servicecomb/core/Endpoint.java
```diff
@@ -61,6 +61,13 @@ public Object getAddress() {
     return address;
   }
 
+  public boolean isThirdPartyEndpoint() {
+    if (null == this.instance) {
```
wujimin: seems not so good.
we also need to support register 3rd services to SC, sometimes maybe also not allow send context header to that target.

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
```diff
@@ -61,6 +61,20 @@
...
+  @InjectProperty(keys = {"request.${op-any-priority}.clientRequestHeaderFilterEnabled",
+      "request.clientRequestHeaderFilterEnabled"}, defaultValue = "true")
+  private boolean clientRequestHeaderFilterEnabled = true;
+
+  /**
+   * the matched headers should be removed, in invocations to 3rd party services
+   */
+  @InjectProperty(keys = {"request.${op-any-priority}.clientRequestHeaderFilterPattern",
+      "request.clientRequestHeaderFilterPattern"}, defaultValue = "^x-cse.*")
+  private String clientRequestHeaderFilterPattern = "^x-cse.*";
```
wujimin: if need to compile pattern, then we can do that in setter method
that can avoid check changes during main business logic.

yhs0092: The logic is simplified. Now just remove all of the related headers, no regex pattern.

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
```diff
@@ -61,6 +61,13 @@
   @InjectProperty(keys = {"request.${op-any-priority}.timeout", "request.timeout"}, defaultValue = "30000")
   private long msRequestTimeout;
 
+  /**
+   * whether to remove certain headers from the 3rd party invocations
```
wujimin: priority keys in the middle, maybe will cause conflict?
better to move to the tail?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/Microservice.java
```diff
@@ -70,6 +70,8 @@
   @JsonIgnore
   private MicroserviceInstance instance;
 
+  private boolean thirdPartyService;
```
wujimin: currently, SC not support this
need to comment

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
yhs0092: I haven't got it why here may cause confict. Since there is always a tail `.clientRequestHeaderFilterEnabled` in this config key, no matter users name their service/schema/operation as "clientRequestHeaderFilterEnabled", the config key seems no problem?

yhs0092: fixed

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/api/registry/Microservice.java
yhs0092: Comment added, please review it

core/src/main/java/org/apache/servicecomb/core/definition/OperationConfig.java
```diff
@@ -64,7 +64,7 @@
   /**
    * whether to remove certain headers from the 3rd party invocations
    */
-  @InjectProperty(keys = {"request.${op-any-priority}.clientRequestHeaderFilterEnabled",
+  @InjectProperty(keys = {"request.clientRequestHeaderFilterEnabled.${op-any-priority}",
       "request.clientRequestHeaderFilterEnabled"}, defaultValue = "true")
```
wujimin: `{"request.clientRequestHeaderFilterEnabled.${op-any-priority}","request.clientRequestHeaderFilterEnabled"}`

can be simpler:
`{"request.clientRequestHeaderFilterEnabled${op-priority}"}`

yhs0092: fixed

PR #1231, opened 2019-05-31
[SCB-1304]FallbackPolicy#getFallbackResponse add Throwable

wuliuqi666:

---

### DIFF
--- a/handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/FallbackPolicy.java
+++ b/handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/FallbackPolicy.java
@@ -22,7 +22,7 @@ import org.apache.servicecomb.swagger.invocation.Response;
 public interface FallbackPolicy {
   String name();
 
-  Response getFallbackResponse(Invocation invocation);
+  Response getFallbackResponse(Invocation invocation, Throwable error);
 
   default void record(Invocation invocation, Response response, boolean isSuccess) {
 
--- a/handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/ThrowExceptionFallbackPolicy.java
+++ b/handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/ThrowExceptionFallbackPolicy.java
@@ -30,11 +30,11 @@ public class ThrowExceptionFallbackPolicy implements FallbackPolicy {
   }
 
   @Override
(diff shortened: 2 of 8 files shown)


handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/ThrowExceptionFallbackPolicy.java
```diff
@@ -30,7 +30,7 @@ public String name() {
   }
 
   @Override
-  public Response getFallbackResponse(Invocation invocation) {
+  public Response getFallbackResponse(Invocation invocation, Throwable error) {
```
wujimin: need change second parameter from null to error of createBizkeeperException, or not?

handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/FallbackPolicy.java
```diff
@@ -22,7 +22,7 @@
 public interface FallbackPolicy {
   String name();
 
-  Response getFallbackResponse(Invocation invocation);
+  Response getFallbackResponse(Invocation invocation, Throwable error);
```
wujimin: seems not compatible?

liubao68: I think the change is fine. Users can change their implementations easily.

handlers/handler-bizkeeper/src/main/java/org/apache/servicecomb/bizkeeper/ThrowExceptionFallbackPolicy.java
liubao68: It's better to use error as the parameter, I think

handlers/handler-bizkeeper/src/test/java/org/apache/servicecomb/bizkeeper/TestFallbackPolicyManager.java
```diff
@@ -33,6 +33,36 @@ public void testFallbackPolicyManager(final @Mocked Configuration config, final
...
+      private static final String CUSTOM = "custom";
+
+      @Override
+      public String name() {
+        return CUSTOM;
+      }
+
+      @Override
+      public Response getFallbackResponse(Invocation invocation, Throwable error) {
+        return Response.succResp("test");
```
liubao68: This test case is better to include some message from error and check it.

PR #1250, opened 2019-06-30
[SCB-1336][WIP][WEAK] should not lost @ApiParam description when wrap parameter to body at swagger generator pojo mode

wujimin:


swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/SwaggerUtils.java
```diff
@@ -115,18 +115,21 @@ public static Swagger parseAndValidateSwagger(String swaggerContent) {
...
+
+    for (Path path : paths.values()) {
+      Operation operation = path.getPost();
+      if (operation == null) {
+        continue;
+      }
+
+      for (Parameter parameter : operation.getParameters()) {
+        if (BodyParameter.class.isInstance(parameter) &&
+            !isRawJsonType(parameter) &&
```
liubao68: add a test case for raw java type is preferred.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/pojo/PojoOperationGenerator.java
```diff
@@ -83,9 +83,14 @@ private void wrapParametersToBody(List<ParameterGenerator> bodyFields) {
...
     for (ParameterGenerator parameterGenerator : bodyFields) {
-      SwaggerUtils.addDefinitions(swagger, parameterGenerator.getGenericType());
+      parameterGenerator.setHttpParameterType(HttpParameterType.BODY);
+      scanMethodParameter(parameterGenerator);
+
       Property property = ModelConverters.getInstance().readAsProperty(parameterGenerator.getGenericType());
+      property.setDescription(parameterGenerator.getGeneratedParameter().getDescription());
       bodyModel.addProperty(parameterGenerator.getParameterName(), property);
+
+      parameterGenerator.setHttpParameterType(null);
```
liubao68: This code is a bit confusing. Should HttpParameterType set to the original one or it is always BODY?

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/SwaggerUtils.java
wujimin: oh, belongs to another PR actually, :)
i will add test case for it.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/pojo/PojoOperationGenerator.java
wujimin: yes
should not keep as "BODY", that will cause multiple bodies
no way to let swagger read ApiParam value, except we copy related logic

i will add some comment for this special logic.

PR #1275, opened 2019-07-22
[SCB-1388]test on "handler-loadbalance" project,cannot all success

ltp217:

---

### DIFF
--- a/handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalanceHandler2.java
+++ b/handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalanceHandler2.java
@@ -19,6 +19,8 @@ package org.apache.servicecomb.loadbalance;
 
 import static org.mockito.Mockito.when;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.EventListener;
 import java.util.HashMap;
@@ -70,9 +72,21 @@ public class TestLoadBalanceHandler2 {
   @BeforeClass
   public static void beforeClass() {
     //prepare for defineEndpointAndHandle
-    ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true");
+    try {
+    Field field = LoadbalanceHandler.class.getDeclaredField("supportDefinedEndpoint");
+    field.setAccessible(true);
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    int modifiers = modifiersField.getInt(field);
+    modifiers &= ~Modifier.FINAL;
+    modifiersField.setInt(field, modifiers);
+    field.set(null, true);
+    
+    } catch (Exception e) {
+    
+	  }
     // avoid mock
-
+    
   }
 
   @Before


handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalanceHandler2.java
```diff
@@ -70,9 +72,21 @@
...
-    ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true");
+    try {
+    Field field = LoadbalanceHandler.class.getDeclaredField("supportDefinedEndpoint");
+    field.setAccessible(true);
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    int modifiers = modifiersField.getInt(field);
+    modifiers &= ~Modifier.FINAL;
+    modifiersField.setInt(field, modifiers);
+    field.set(null, true);
```
yhs0092: Hello, maybe it's better to set field `supportDefinedEndpoint` by `ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true")` instead of reflection, since changing field by reflection makes the UT cannot cover the related configuration, and may cause other side effect.

yhs0092:
Hello, thank you for your pull request. Could you please reformat the code? The code template can be found [in our project](https://github.com/apache/servicecomb-java-chassis/tree/master/etc).
And you can change the assignee of the jira issue [SCB-1388](https://issues.apache.org/jira/browse/SCB-1388) to yourself.

handlers/handler-loadbalance/src/test/java/org/apache/servicecomb/loadbalance/TestLoadBalanceHandler2.java
ltp217: @yhs0092 yeah, int the master branch, it uses `ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true")`.

There are two test classes in the `handler-loadbalance` project, one is `TestLoadBalanceHandler`, the other is `TestLoadBalanceHandler2`.  

when first execute `TestLoadBalanceHandler` , in this class, it will init the  `supportDefinedEndpoint` to `false`, after that when it execute `TestLoadBalanceHandler2`, it called `ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true")`, however, **the `supportDefinedEndpoint` will never changed, it will be always `false`, because of this attribute is `final static`**, after that the test case in `TestLoadBalanceHandler2` may have some exceptions.

wujimin: better to call `ArchaiusUtils.resetConfig();` both in `@BeforeClass` and `@AfterClass`

wujimin: i remember after i add this, random fail in this class gone

yhs0092: @wujimin  Both in `TestLoadbalanceHandler` and `TestLoadbalanceHandler2`, the `ArchaiusUtils.resetConfig()` is called in `@After` method. But the field `supportDefinedEndpoint` doesn't hold the `DynamicBooleanProperty` object, it is a boolean value instead.
Maybe we should change the `supportDefinedEndpoint` into `DynamicBooleanProperty` type, like below?
```java
  public static final DynamicBooleanProperty supportDefinedEndpoint =
      DynamicPropertyFactory.getInstance()
          .getBooleanProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", false);
```

wujimin: after resetConfig, DynamicXXXProperty will be invalid
LoadBalanceHandler will not create too many instance, so no need to make configuration variables static, static will make test too difficult

ltp217: so is there any advices?

yhs0092: @wujimin  Do you mean removing the `static` modifier of `org.apache.servicecomb.loadbalance.LoadbalanceHandler#supportDefinedEndpoint`?
That indeed works. And the operation `ArchaiusUtils.setProperty("servicecomb.loadbalance.userDefinedEndpoint.enabled", "true");` should be moved from `@BeforeClass` method to `@Before` method, since in `@After` method, the configurations are reset, once other test case is run before `testConfigEndpoint`, there is still the problem.

wujimin: yes

yhs0092: @ltp217  Hi, as is discussed above, could you please modify the pull request? And please format the code : )

yhs0092:
Hello, is there any update for the PR?

yhs0092:
Hi, this PR seems a duplicate of https://github.com/apache/servicecomb-java-chassis/pull/1313 . Since that PR has been merged, I'll close this PR. Please feel free to reopen it if necessary.
Thank you for your contribution : )

PR #1277, opened 2019-07-25
[SCB-1392]Mutiple classes have the same APIModel,and class will occur…

GuoYL123:
… when generating schemaid.

---

### DIFF
--- a/swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
+++ b/swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
@@ -25,6 +25,8 @@ import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.servicecomb.swagger.generator.core.OperationGenerator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.DefaultParameterNameDiscoverer;
 import org.springframework.core.MethodParameter;
 
@@ -45,6 +47,8 @@ import io.swagger.models.properties.StringProperty;
 
 public final class ParamUtils {
   private static DefaultParameterNameDiscoverer parameterNameDiscoverer = new DefaultParameterNameDiscoverer();
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ParamUtils.class);
 
   private ParamUtils() {
 
@@ -115,9 +119,17 @@ public final class ParamUtils {
   public static void addDefinitions(Swagger swagger, Type paramType) {
     Map<String, Model> models = ModelConverters.getInstance().readAll(paramType);
     for (Map.Entry<String, Model> entry : models.entrySet()) {
+      if (null != swagger.getDefinitions()) {
+        Model tempModel = swagger.getDefinitions().get(entry.getKey());
+        if (null != tempModel && !tempModel.equals(entry.getValue())) {
+          LOGGER.warn("duplicate param model: " + entry.getKey());
+          throw new IllegalArgumentException("duplicate param model: " + entry.getKey());
+        }
+      }
       swagger.addDefinition(entry.getKey(), entry.getValue());
     }
   }
+
 
   public static void setParameterType(Swagger swagger, Method method, int paramIdx,
       AbstractSerializableParameter<?> parameter) {
--- a/swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
+++ b/swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
@@ -17,12 +17,22 @@
 
 package org.apache.servicecomb.swagger.generator.core;
 
+import static junit.framework.TestCase.fail;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.when;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
+import io.swagger.models.Swagger;
+
+import org.apache.servicecomb.swagger.generator.core.pojo.TestType1;
+import org.apache.servicecomb.swagger.generator.core.pojo.TestType2;
 import org.apache.servicecomb.swagger.generator.core.utils.ClassUtils;
 import org.apache.servicecomb.swagger.generator.core.utils.ParamUtils;
 import org.junit.Assert;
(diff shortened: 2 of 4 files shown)


swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
```diff
@@ -113,12 +111,16 @@ public static BodyParameter createBodyParameter(Swagger swagger, String paramNam
   }
 
   public static void addDefinitions(Swagger swagger, Type paramType) {
+    if (swagger.getDefinitions() == null) {
+      ModelResolverExt.refreshClassNameMap();
```
wujimin: this makes swagger generator mechanism from thread safe to not thread safe

swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
```diff
@@ -95,4 +104,34 @@ public void setParameterTypeByTypeOnComplexType() {
...
+
+    List<TestType1> t3;
+
+    Map<String, TestType2> t4;
+
+    Map<TestType1, TestType2> t5;
+  }
+
+  @Test
+  public void testAddDefinitions() {
```
wujimin: expect not throw exception?
just generate swagger, and check the result?

GuoYL123: considering coexists with previous demo,only print log

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
GuoYL123: can i use  ThreadLocal fix?

swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
wujimin: better to check the result.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/extend/ModelResolverExt.java
```diff
@@ -129,13 +135,32 @@ public Model resolve(JavaType type, ModelConverterContext context, Iterator<Mode
...
     if (ModelImpl.class.isInstance(model) && !StringUtils.isEmpty(((ModelImpl) model).getName())) {
       setType(type, model.getVendorExtensions());
     }
     return model;
   }
 
+  private void checkRepeat(JavaType type, ModelImpl model) {
+    if (classNameMap.containsKey(model.getName())) {
+      if (!classNameMap.get(model.getName()).equals(type)) {
+        LOGGER.warn("The class " + type + " and the class " + classNameMap.get(model.getName())
```
wujimin: this will cause problem, so should throw exception?

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
wujimin: seems we can check this depend on swagger instance, no need to modify ModelResolverExt.
we can save a instance in swagger.vendorExtensions, note: key should not start with "x-"

GuoYL123: According to my understanding，if use swaggger we must get the type then store in vendorExtensions.  the way we get the type is very complex ,cause we must handle like list<T>, map<T>, nesting and so on . you can see my first commit '9a2b56a1b2d8535ee0d211ee990fe0957f028825'. i found the ModelResolverExt can help me get type later , and i take the way you see now.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/extend/ModelResolverExt.java
GuoYL123: there is a previos demo 'demo-springmvc' always fail if i throw exception, so don't throw exception is for Compatibility.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
yhs0092:
```java
  public static void addDefinitions(Swagger swagger, Type paramType) {
    Map<String, Model> models = ModelConverters.getInstance().readAll(paramType);
    if (null!=swagger.getDefinitions()) {
      for (Entry<String, Model> modelEntry : models.entrySet()) {
        Model tempModel = swagger.getDefinitions().get(modelEntry.getKey());
        if (null != tempModel && !tempModel.equals(modelEntry.getValue())) {
          throw new IllegalArgumentException("duplicate param model: " + modelEntry.getKey());
        }
      }
    }
    for (Map.Entry<String, Model> entry : models.entrySet()) {
      swagger.addDefinition(entry.getKey(), entry.getValue());
    }
  }
```
Seems the logic like above can works? It's in `org.apache.servicecomb.swagger.generator.core.utils.ParamUtils#addDefinitions` method. As long as all of the subclasses of `Model` implement `equals` properly, we can just check the definition map.

GuoYL123: ok, i refactory my code

swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
```diff
@@ -95,4 +104,39 @@ public void setParameterTypeByTypeOnComplexType() {
...
+
+
+  @Test
+  public void testAddDefinitions() {
+    Field[] fields = AllTypeTest.class.getDeclaredFields();
+    for (int i = 0; i < fields.length; i++) {
+      for (int j = 0; j < fields.length; j++) {
+        if (i != j) {
+          try {
+            testExcep(fields[i].getGenericType(), fields[j].getGenericType());
```
yhs0092: Hi, here you can add a `fail` assertion here to ensure there must be an Exception. Like below:
```java
    try {
      PojoService.test.testException(557);
      fail("InvocationException expected, but threw nothing");
    } catch (InvocationException e) {
      assertThat(e.getStatusCode(), is(557));
      assertThat(e.getErrorData().toString(), is("[[557 error]]"));
    } catch (Exception e) {
      fail("InvocationException expected, but threw " + e);
    }
```

swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
```diff
@@ -95,4 +105,49 @@ public void setParameterTypeByTypeOnComplexType() {
...
+    TestType2[] t4;
+  }
+
+  @Test
+  public void testAddDefinitions() {
+    Field[] fields1 = AllTypeTest1.class.getDeclaredFields();
+    Field[] fields2 = AllTypeTest2.class.getDeclaredFields();
+    for (int i = 0; i < fields1.length; i++) {
+      for (int j = 0; j < fields2.length; j++) {
+        if(fields1[i].isSynthetic()||fields2[j].isSynthetic()){
```
GuoYL123: jacoco will use ASM injecting bytecode into field,this caused a new field named JacocoData will not throw exception when execute  addDefinitions.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
```diff
@@ -17,14 +17,13 @@
 
 package org.apache.servicecomb.swagger.generator.core.utils;
 
-import java.lang.reflect.Executable;
-import java.lang.reflect.Method;
-import java.lang.reflect.Type;
-import java.util.List;
-import java.util.Map;
+import java.lang.reflect.*;
+import java.util.*;
```
liubao68: Better not use * in import

swagger/swagger-generator/generator-core/src/test/java/org/apache/servicecomb/swagger/generator/core/TestParamUtils.java
liubao68: This line of code is not formatted.

swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ParamUtils.java
GuoYL123: done

PR #1281, opened 2019-07-29
[SCB-1402] BeanUtils.getImplClassFromBean can not return correctly in…

AngLi2:
…fo in cglib proxy situation

---

### DIFF
--- a/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/BeanUtils.java
+++ b/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/BeanUtils.java
@@ -17,15 +17,15 @@
 
 package org.apache.servicecomb.foundation.common.utils;
 
-import java.util.LinkedHashSet;
-import java.util.Set;
-
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/BeanUtils.java
```diff
@@ -110,10 +116,36 @@ public static void setContext(ApplicationContext applicationContext) {
   }
 
   public static Class<?> getImplClassFromBean(Object bean) {
```
liubao68: Why not call AopProxyUtils.ultimateTargetClass directly?

AngLi2: > Why not call AopProxyUtils.ultimateTargetClass directly?

I am not sure whether should we deal with the 'Advised' case later, so the 'getSingletonTarget' method is copied from AopProxyUtils and put here for later processing.
Do you think using AopProxyUtils#ultimateTargetClass directly is better? I can change this PR if needed.

liubao68: I think use directly is better, if we got specific reasons not use it.

PR #1283, opened 2019-07-30
[SCB-1407] Add new field when print provideCache log

GuoYL123:

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
@@ -68,6 +68,10 @@ public class MicroserviceInstanceRegisterTask extends AbstractRegisterTask {
     microserviceInstance.getHealthCheck().setTimes(serviceRegistryConfig.getResendHeartBeatTimes());
 
     String instanceId = srClient.registerMicroserviceInstance(microserviceInstance);
+    MicroserviceInstance temIns = srClient.findServiceInstance(microserviceInstance.getServiceId(), instanceId);
+    if (temIns != null) {
+      microserviceInstance.setTimestamp(temIns.getTimestamp());
+    }
     if (StringUtils.isEmpty(instanceId)) {
       LOGGER.error("Register microservice instance failed,will back to register microservice again. microserviceId={}",
           microserviceInstance.getServiceId());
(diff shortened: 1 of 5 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
```diff
@@ -68,6 +68,10 @@ protected boolean doRegister() {
     microserviceInstance.getHealthCheck().setTimes(serviceRegistryConfig.getResendHeartBeatTimes());
 
     String instanceId = srClient.registerMicroserviceInstance(microserviceInstance);
+    MicroserviceInstance temIns = srClient.findServiceInstance(microserviceInstance.getServiceId(),instanceId);
+    if(temIns!=null){
```
liubao68: Use the code formatter in etc folder to format code

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
```diff
@@ -68,6 +68,10 @@ protected boolean doRegister() {
...
+      microserviceInstance.setTimestamp(temIns.getTimestamp());
```
liubao68: Does service-center already add timestamp property and which version？ And if it is neccessary to set timestamp information for the service itself (may be do not need set this property, or we'd better to keep itself reference from service center, not set a single property)?

wujimin: i remember SC provide this property in very very very old version
but better to confirm that.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
GuoYL123: done

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
GuoYL123: it's always provide this property , java-chassis don't use it. we need to display the timestemp

PR #1301, opened 2019-08-20
[SCB-1450] Microservice instance may get stuck when watch connection to sc is recovered and pulling instance action is executed in eventloop thread

HuaiJoe:

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/AbstractClientPool.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/AbstractClientPool.java
@@ -41,17 +41,7 @@ import io.vertx.core.http.HttpClientOptions;
 public abstract class AbstractClientPool implements ClientPool {
   private static final Logger LOGGER = LoggerFactory.getLogger(AbstractClientPool.class);
 
-  protected static final String SSL_KEY = "sc.consumer";
-
-  public static final String PROXY_KEY = "sc.consumer";
-
-  private static final String EVENT_LOOP_POOL_SIZE = "servicecomb.service.registry.client.eventLoopPoolSize";
-
-  private static final String WORKER_POOL_SIZE = "servicecomb.service.registry.client.workerPoolSize";
-
-  private static final String WORKER_POOL_NAME = "registry-vert.x-worker-thread";
-
-  private ClientPoolManager<HttpClientWithContext> clientMgr;
+    private ClientPoolManager<HttpClientWithContext> clientMgr;
 
   public AbstractClientPool() {
     create();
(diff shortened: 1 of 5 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/config/ServiceRegistryConfig.java
```diff
@@ -109,22 +109,23 @@ public HttpVersion getHttpVersion() {
     return HttpVersion.valueOf(property.get());
   }
 
-  public int getWorkerPoolSize() {
-    String workerPoolSizeKey = "servicecomb.service.registry.client.workerPoolSize";
+  public int getInstances() {
+    String instances = "servicecomb.service.registry.client.instances ";
```
liubao68: Define a constant is better

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/AbstractClientPool.java
```diff
@@ -42,27 +45,41 @@
 
   public static final String PROXY_KEY = "sc.consumer";
 
+  private static final String EVENT_LOOP_POOL_SIZE = "servicecomb.service.registry.client.eventLoopPoolSize";
+
```
liubao68: move all configurations to ServiceRegistryConfig is better

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/AbstractClientPool.java
```diff
@@ -42,27 +45,41 @@
...
+  private static final String WORKER_POOL_SIZE = "servicecomb.service.registry.client.workerPoolSize";
```
yhs0092: `servicecomb.service.registry.client.workerPoolSize` is used as the key of another configurations. This is a change not compatible. Could you update the corresponding docs?

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/config/ServiceRegistryConfig.java
yhs0092: There is a redundant space char at the tail of this string

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/config/ServiceRegistryConfig.java
```diff
@@ -109,22 +109,23 @@ public HttpVersion getHttpVersion() {
...
+            DynamicPropertyFactory.getInstance()
+                    .getIntProperty(instances, 1);
+    int deployInstances = property.get();
+    if (deployInstances <= 0) {
       int nAvailableProcessors = Runtime.getRuntime().availableProcessors();
       LOGGER.warn("The property `{}` must be positive integer, fallback to use number of available processors: {}",
-          workerPoolSizeKey,
-          nAvailableProcessors);
+              instances,
+              nAvailableProcessors);
```
yhs0092: The indent of some files seems not consistent with the other code, like here and `AbstractClientPool`. Could you reformat your pull request? You can find the code template [here](https://github.com/apache/servicecomb-java-chassis/tree/master/etc).

liubao68:
Can you please check if config-cc has the same problem?

HuaiJoe:
> Can you please check if config-cc has the same problem?

Config Center client may come across the same problem as Service Center client.I will create a issue and fix it later.

PR #1302, opened 2019-08-20
[SCB-1451] Printing the service information on the log

ZihangJiao:

---

### DIFF
--- a/core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
+++ b/core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
@@ -30,6 +30,7 @@ import org.apache.servicecomb.config.ConfigUtil;
 import org.apache.servicecomb.config.priority.PriorityPropertyManager;
 import org.apache.servicecomb.core.BootListener.BootEvent;
 import org.apache.servicecomb.core.BootListener.EventType;
+import org.apache.servicecomb.core.bootup.BootUpInformationCollector;
 import org.apache.servicecomb.core.definition.MicroserviceMeta;
 import org.apache.servicecomb.core.definition.loader.SchemaListenerManager;
 import org.apache.servicecomb.core.definition.schema.StaticSchemaFactory;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
```diff
@@ -223,10 +228,22 @@ public synchronized void init() {
         }
         status = SCBStatus.FAILED;
         throw new IllegalStateException("ServiceComb init failed.", e);
+      } finally {
+        ServiceInfoPrinter();
       }
     }
   }
 
+  private void ServiceInfoPrinter() {
```
yhs0092: - method name should follow CamelCase
- please use verb as method name, e.g. `printServiceInfo` is better than `ServiceInfoPrinter`

core/src/main/java/org/apache/servicecomb/core/bootup/AddressInformationCollector.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private String getCenterInfo(SystemBootstrapInfo systemBootstrapInfo) {
+        if (systemBootstrapInfo == null) {
+            return ("not exist");
+        } else {
+            return systemBootstrapInfo.getAccessURL().toString();
+        }
+    }
+
+}
```
yhs0092: The indent of some files a different from our code style, please reformat it : )
And you can find code style config files [here](https://github.com/apache/servicecomb-java-chassis/tree/master/etc)

core/src/main/java/org/apache/servicecomb/core/SCBEngine.java
ZihangJiao: The name of the method have been changed to printServiceInfo.

core/src/main/java/org/apache/servicecomb/core/bootup/AddressInformationCollector.java
ZihangJiao: The code is formatted

PR #1326, opened 2019-09-12
[SCB-1513]merge config items according to priority.

zhangchi02:


dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ParseConfigUtils.java
```diff
@@ -130,11 +134,23 @@ private void doRefreshItems() {
   }
 
   private Map<String, Object> mergeDimensionItems(Map<String, Map<String, Object>> items) {
-    Map<String, Object> flatMap = new HashMap<>();
-    return items.values().stream().reduce(flatMap, (result, item) -> {
-      result.putAll(item);
-      return result;
-    });
+	Map<String, Object> flatMap = new HashMap<>();
```
jeho0815: Your code format style is wrong, please use the format file under https://github.com/apache/servicecomb-java-chassis/tree/master/etc

dynamic-config/config-cc/src/test/java/org/apache/servicecomb/config/client/TestParseConfigUtils.java
```diff
@@ -80,4 +81,26 @@ public void testNotifyItemsChangedNeedRefresh() {
...
+	items.put("application", application);
+	items.put("service@app", service);
+	items.put("service@app#version", version);
+	items.put("service@app#version!tag", tag);
+    try {
+      Deencapsulation.invoke(pc, "mergeDimensionItems", items);
+    } catch (Exception e) {
+      status = false;
+    }
+    Assert.assertTrue(status);
```
jeho0815: the dimentions should be out of order and  you can add some key/values and , then check the merged result

liubao68:
Can you please resolve the conflicsts ?

liubao68:
Please retrigger CI after #1340 is merged or some test case will randomly fail.

PR #1334, opened 2019-09-23
[SCB-1494] Optimizing log when starting a new microservice

AngLi2:
[SCB-1494] Optimizing log when starting a new microservice

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
@@ -19,6 +19,7 @@ package org.apache.servicecomb.serviceregistry.client.http;
 
 import static java.util.Collections.emptyList;
 
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -74,6 +75,10 @@ import io.vertx.core.http.HttpClientResponse;
 
 public final class ServiceRegistryClientImpl implements ServiceRegistryClient {
   private static final Logger LOGGER = LoggerFactory.getLogger(ServiceRegistryClientImpl.class);
+
+  private static final String ERROR_CODE = "errorCode";
+  private static final String ERR_SERVICE_NOT_EXISTS = "400012";
+  private static final String ERR_SCHEMA_NOT_EXISTS = "400016";
 
   private IpPortManager ipPortManager;
 
@@ -140,6 +145,22 @@ public final class ServiceRegistryClientImpl implements ServiceRegistryClient {
               holder.setValue((T) bodyBuffer.toString());
               countDownLatch.countDown();
               return;
+            }
+
+            // no need to generate warn log when schema or service not exist
+            if (HttpStatusClass.CLIENT_ERROR.equals(HttpStatusClass.valueOf(response.statusCode()))) {
+              try {
+                Map<String, String> bufferMap = JsonUtils.readValue(bodyBuffer.getBytes(), Map.class);
+                if (bufferMap.containsKey(ERROR_CODE)) {
+                  String errorCode = bufferMap.get(ERROR_CODE);
+                  if (errorCode.equals(ERR_SERVICE_NOT_EXISTS) || errorCode.equals(ERR_SCHEMA_NOT_EXISTS)) {
+                    countDownLatch.countDown();
+                    return;
+                  }
+                }
+              } catch (IOException e) {
+                LOGGER.warn("read value failed from buffer {}", bodyBuffer.toString());
+              }
             }
 
             // no need to support 304 in this place


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -142,6 +143,23 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
...
+                String errorCode = JsonUtils.readValue(bodyBuffer.getBytes(), Map.class).get("errorCode").toString();
+                // 400012: ErrServiceNotExists
+                // 400016: ErrSchemaNotExists
+                if (errorCode.equals("400012") || errorCode.equals("400016")) {
+                  countDownLatch.countDown();
+                  return;
+                }
+              } catch (IOException | NullPointerException ignored) {
+                // Ignore IOException when read value from bodyBuffergit
+                // Ignore NullPointerException when bodyBuffer not contains errorCode
```
liubao68: Does this mean these two exception will never happen? I don't think it's not a good idea to ingore them when the exception will throw.

AngLi2: IOException occures when JsonUtil read value failed and NullPointerException occurs when bodyBuffer not contains errorCode. I will log some warn info to deal with them.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -142,6 +143,24 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
...
+                // 400012: ErrServiceNotExists
+                // 400016: ErrSchemaNotExists
+                if (errorCode.equals("400012") || errorCode.equals("400016")) {
+                  countDownLatch.countDown();
+                  return;
+                }
+              } catch (IOException e) {
+                LOGGER.warn("read value failed from buffer {}", bodyBuffer.toString());
+              } catch (NullPointerException e) {
+                LOGGER.warn("error code not included in buffer {}", bodyBuffer.toString());
```
liubao68: Check  errorCode is null is better, never throw NPE and catch

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
AngLi2: Already modified as suggested

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
AngLi2: Modified, now check errorCode before using it

liubao68:
Please retrigger CI after #1340  is merged or some test case will randomly fail.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -142,6 +144,24 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
               return;
             }
 
+            // no need to generate warn log when schema or service not exist
+            if (HttpStatusClass.CLIENT_ERROR.equals(HttpStatusClass.valueOf(response.statusCode()))) {
+              try {
+                if (JsonUtils.readValue(bodyBuffer.getBytes(), Map.class).containsKey("errorCode")) {
+                  @SuppressWarnings("unchecked")
```
liubao68: @SuppressWarnings("unchecked")  should also cover the above line.   
```
JsonUtils.readValue(bodyBuffer.getBytes(), Map.class)
```
is a big operation, it's better only call it once.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/ServiceRegistryClientImpl.java
```diff
@@ -142,6 +144,24 @@ private void retry(RequestContext requestContext, Handler<RestResponse> response
...
 
+            // no need to generate warn log when schema or service not exist
+            if (HttpStatusClass.CLIENT_ERROR.equals(HttpStatusClass.valueOf(response.statusCode()))) {
+              try {
+                Map<String, String> bufferMap = JsonUtils.readValue(bodyBuffer.getBytes(), Map.class);
+                if (bufferMap.containsKey("errorCode")) {
+                  String errorCode = bufferMap.get("errorCode");
+                  // 400012: ErrServiceNotExists
+                  // 400016: ErrSchemaNotExists
+                  if (errorCode.equals("400012") || errorCode.equals("400016")) {
```
yhs0092: Better to extract constants, like `ERR_SERVICE_NOT_EXISTS`

PR #1340, opened 2019-10-08
[SCB-1511]when use method override, the generated swagger body is wrong

liubao68:

---


swagger/swagger-generator/generator-core/src/main/java/org/apache/servicecomb/swagger/generator/core/utils/ClassUtils.java
```diff
@@ -49,10 +49,10 @@ private ClassUtils() {
 
   // 将一系列body parameter包装成一个class
   public static Class<?> getOrCreateBodyClass(OperationGenerator operationGenerator,
-      List<BodyParameter> bodyParameters) {
+      List<BodyParameter> bodyParameters, String bodyParamName) {
     SwaggerGenerator swaggerGenerator = operationGenerator.getSwaggerGenerator();
     Method method = operationGenerator.getProviderMethod();
```
yhs0092: The variable `method` seems not used. Maybe we can remove the parameter `bodyParamName` and generate it by invoking `ParamUtils.generateBodyParameterName(method)` ?

PR #1346, opened 2019-10-12
[SCB-1522] add "Why use Java Chassis" introduction into README

yhs0092:


liubao68:
Maybe we can also add 中文 README, just as https://github.com/apache/servicecomb-fence do

yhs0092:
That's a good idea. I'll create another PR later.

PR #1349, opened 2019-10-15
[SCB-1525]print codec error stack

liubao68:

---

### DIFF
--- a/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
+++ b/common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
@@ -22,11 +22,14 @@ import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.core.Response.Status;
 
+import org.apache.servicecomb.common.rest.RestConst;
 import org.apache.servicecomb.common.rest.definition.RestOperationMeta;
 import org.apache.servicecomb.common.rest.definition.RestParam;
 import org.apache.servicecomb.swagger.invocation.exception.InvocationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.netflix.config.DynamicPropertyFactory;
 
 public final class RestCodec {
   private static final Logger LOG = LoggerFactory.getLogger(RestCodec.class);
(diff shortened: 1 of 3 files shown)


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
```diff
@@ -61,7 +64,11 @@ public static void argsToRest(Object[] args, RestOperationMeta restOperation,
       try {
         paramValues[idx] = param.getParamProcessor().getValue(request);
       } catch (Exception e) {
-        // Avoid information leak of user input.
+        // Avoid information leak of user input, and add option for debug use.
+        if (DynamicPropertyFactory.getInstance().getBooleanProperty(
+            RestConst.PRINT_CODEC_ERROR_MESSGAGE, false).get()) {
+          LOG.error("", e);
+        }
```
zhangchi02: I want to get the original exception information in the code. This is just printed out in the log. Can you wrap the exception as cause and throw it together in InvocationException?

liubao68: It's not possible doing that in java-chassis. "Exception" is a complicated object that cannot expressed by swagger, and the serialization/deserialization may couple consumer/provider.  

see:
https://docs.servicecomb.io/java-chassis/zh_CN/general-development/error-handling.html

Developers should not send unkown exceptions to consumers to avoid un-neccessary information leak and try to catch exceptions and return friendly error code and error mesasge is recommented.

yhs0092: Hi, maybe we can set the raw exception to the cause of the InvocationException?
See the constructor `InvocationException(int statusCode, String reasonPhrase, Object errorData, Throwable cause)`

liubao68: This will have drawbacks. We can not expose internal exception trace to client(not recommented). And if we do serialize exception to the client, the client may be not possible to decode the exception.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
```diff
@@ -61,12 +64,19 @@ public static void argsToRest(Object[] args, RestOperationMeta restOperation,
...
+        String message = String.format("Parameter is not valid for operation [%s]. Parameter is [%s]. Processor is [%s].",
+            restOperation.getOperationMeta().getMicroserviceQualifiedName(),
+            param.getParamName(),
+            param.getParamProcessor().getProcessorType());
+        if (DynamicPropertyFactory.getInstance().getBooleanProperty(
+            RestConst.PRINT_CODEC_ERROR_MESSGAGE, false).get()) {
+          LOG.error("", e);
+          throw new InvocationException(Status.BAD_REQUEST.getStatusCode(), "", message, e);
+        } else {
+          LOG.error(message + " Add {}=true to print the details." + RestConst.PRINT_CODEC_ERROR_MESSGAGE);
```
yhs0092: Here should use comma instead of plus. And since `RestConst.PRINT_CODEC_ERROR_MESSGAGE` is a const, why not combine the String by `+` directly?

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
```diff
@@ -61,12 +64,19 @@ public static void argsToRest(Object[] args, RestOperationMeta restOperation,
...
-                param.getParamName(),
-                param.getParamProcessor().getProcessorType()));
+        // Avoid information leak of user input, and add option for debug use.
+        String message = String.format("Parameter is not valid for operation [%s]. Parameter is [%s]. Processor is [%s].",
+            restOperation.getOperationMeta().getMicroserviceQualifiedName(),
+            param.getParamName(),
+            param.getParamProcessor().getProcessorType());
+        if (DynamicPropertyFactory.getInstance().getBooleanProperty(
+            RestConst.PRINT_CODEC_ERROR_MESSGAGE, false).get()) {
+          LOG.error("", e);
```
yhs0092: Maybe it's better to add some key words for the convenience of searching error log?

PR #1351, opened 2019-10-16
[SCB-1523] Stopping retrying another server when no server available

AngLi2:

---

### DIFF
--- a/handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
+++ b/handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
@@ -98,7 +98,10 @@ public class LoadbalanceHandler implements Handler {
       boolean isRetry = null != lastServer;
       for (int i = 0; i < COUNT; i++) {
         Server s = delegate.chooseServer((Invocation) key);
-        if (s != null && !s.equals(lastServer)) {
+        if (s == null) {
+          break;
+        }
+        if (!s.equals(lastServer)) {
           lastServer = s;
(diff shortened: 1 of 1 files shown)


handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -98,6 +98,9 @@ public Server chooseServer(Object key) {
       boolean isRetry = null != lastServer;
       for (int i = 0; i < COUNT; i++) {
         Server s = delegate.chooseServer((Invocation) key);
+        if (s == null) {
+          break;
+        }
         if (s != null && !s.equals(lastServer)) {
```
liubao68: unneccessary check

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/LoadbalanceHandler.java
```diff
@@ -98,6 +98,9 @@ public Server chooseServer(Object key) {
       boolean isRetry = null != lastServer;
       for (int i = 0; i < COUNT; i++) {
         Server s = delegate.chooseServer((Invocation) key);
+        if (s == null) {
+          break;
+        }
```
liubao68: unneccessary null check

AngLi2: Already updated.
break the loop when s == null
only continue the loop and retry another server when s.equals(lastServer)

PR #1353, opened 2019-10-16
[SCB-1528] Optimizing some code implementations

AngLi2:

---


AngLi2:
1. Iterating entrySet() is more efficent than entryKey() when call map.get(key) each time;
2. Collections.isEmpty() is more efficent than collections.size() == 0
3. Util classes should have private constructures
4. String.valueOf() is more efficent then + ""
5. BigDecimal() may cause some precision problems while BigDecimal.valueOf() not
6. Return empty collection instead of null can prevent NPE
7. Objects.equals(a, b) is better than a.equals(b) because of NPE
8. Data in enum shoud be private final

integration-tests/it-edge/src/main/java/org/apache/servicecomb/it/edge/PreLoadBootListener.java
```diff
@@ -1,46 +0,0 @@
...
-  @Override
-  public int getOrder() {
-    return 0;
-  }
-
-  @Override
-  public void onBootEvent(BootEvent bootEvent) {
-    if (bootEvent.getEventType() == EventType.BEFORE_REGISTRY) {
-      MicroserviceVersionRule rule = RegistryUtils.getServiceRegistry().getAppManager()
-          .getOrCreateMicroserviceVersionRule(RegistryUtils.getAppId(), "it-producer", "0+");
```
liubao68: This test case is useful that we must gurrantee we can call getOrCreateMicroserviceVersionRule after BEFORE_REGISTRY without exception throw in bootup.

integration-tests/it-edge/src/main/java/org/apache/servicecomb/it/edge/PreLoadBootListener.java
```diff
@@ -26,21 +26,21 @@
...
-      if (rule.getInstances().size() == 0) {
-        LOGGER.warn("Prefetch not successful, maybe the provider not started.");
-      }
+    @Override
+    public void onBootEvent(BootEvent bootEvent) {
```
liubao68: It seems this code does not use servicecomb recommented formatter.

PR #1355, opened 2019-10-18
[SCB-1489] Use old instanceId to re-register instance to service center

AngLi2:

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
@@ -103,9 +103,6 @@ public class MicroserviceRegisterTask extends AbstractRegisterTask {
         return false;
       }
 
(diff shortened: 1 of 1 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceRegisterTask.java
```diff
@@ -103,9 +103,6 @@ protected boolean doRegister() {
         return false;
       }
 
-      // In re-register microservice case, the old instanceId should not be cached
-      microservice.getInstance().setInstanceId(null);
-
       LOGGER.info(
```
liubao68: Maybe we need to set instance id to null when service id is empty. (When we delete all information in service center, both the service and instance is a new one.)

liubao68: we can think of a process to be an instance, use the old instance id is reasonable

PR #1356, opened 2019-10-23
[SCB-1565] add a Java client of servicecomb-service-center

zaneChou1:
I provide a simple java client, which can operate service-center conveniently.


liubao68:
Hello, thank you for you contribution and the idea provided in this PR. 
However I don't think this implementation is not a good idea to put in servicecomb-java-chassis project, maybe create a new project is better. 
servicecomb-java-chassis already has it's own service-center client, in service-registry module. The best way for servicecomb-java-chassis to provide client api for service-center is to refactor the module code, and let other developers use only the module API. And design an open API set for integration other service-registry implementations, like consule, for example.

liubao68:
Please feel convenient to send a DISCUSSION mail to dev@servicecomb.apache.org

zaneChou1:
> Please feel convenient to send a DISCUSSION mail to [dev@servicecomb.apache.org](mailto:dev@servicecomb.apache.org)

There are some discuss in mail list (https://lists.apache.org/thread.html/e7e66dc75f27a42938b6bfb56e609a296ac5a2ac17c06cc2839e17ca@%3Cdev.servicecomb.apache.org%3E )
and github PR 
(https://github.com/apache/servicecomb-service-center/pull/594)

And the next step is to add kie java client.  

Creating a new project is not easy. It may be appropriate to temporarily store in Java chassis. And this can be discussed again.

liubao68:
You mentioned this components maybe used for spring cloud, and I checked project [spring-cloud-hauwei](https://github.com/huaweicloud/spring-cloud-huawei/blob/master/README_zh.md), this project has already project clients and discovery features for spring cloud.

zaneChou1:
> You mentioned this components maybe used for spring cloud, and I checked project [spring-cloud-hauwei](https://github.com/huaweicloud/spring-cloud-huawei/blob/master/README_zh.md), this project has already project clients and discovery features for spring cloud.

I just want to implement a client for servicecomb server components, such as service-center and kie. 
This client will implement all APIs of service-center in the future, not only registry and discovery.
 And it can be used by other microservice framework, not only spring cloud.

liubao68:
> > You mentioned this components maybe used for spring cloud, and I checked project [spring-cloud-hauwei](https://github.com/huaweicloud/spring-cloud-huawei/blob/master/README_zh.md), this project has already project clients and discovery features for spring cloud.
> 
> I just want to implement a client for servicecomb server components, such as service-center and kie.
> This client will implement all APIs of service-center in the future, not only registry and discovery.
> And it can be used by other microservice framework, not only spring cloud.

This is a good idea. But I have some requirements that must be done to merge the code:
1. this code can replace the service-registry implementation. It's not a good idea to have two sets of code doing the same job. 
2. this code can replace the implementation of spring-cloud-huawei implementation. 
Match the two requirements can veriry your API design is correct for future. For API module, frequently API change is very big problem, must do it very carefully in design and testing before make it visible to users.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/http/TLSConfig.java
```diff
@@ -0,0 +1,92 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.servicecomb.service.center.client.http;
+
+public class TLSConfig {
```
liubao68: You can use SSLCustom & SSLOption to load TLS configurations and will save a lot of work. And see VertxTLSBuilder for an example. Documents are :　https://docs.servicecomb.io/java-chassis/zh_CN/security/tls.html

yhs0092:
Hello, thank you for your contribution.
I've noticed that the Apache httpclient is introduced in this pull request. Because in Java-Chassis, the common transport functions are based on Vert.x, it's recommended that you use the http client of Vert.x to get access to Service-Center instead of the Apache httpclient. And I'm going to refactor the transport related code of Java-Chassis to provide a common http module, maybe we can write a Service-Center later based on that http module.

liubao68:
I created an [issue](https://issues.apache.org/jira/browse/SCB-1565) for this RP

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,378 @@
...
+import org.apache.servicecomb.service.center.client.model.MicroserviceInstanceStatus;
+import org.apache.servicecomb.service.center.client.model.MicroserviceInstancesResponse;
+import org.apache.servicecomb.service.center.client.model.MicroservicesResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Created by   on 2019/10/16.
+ */
+public class ServiceCenterClient {
```
liubao68: Please use formatters in $project/etc to format code

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,378 @@
...
+
+    /**
+     * Customized host,port,domainName,TLSConf, headers and any one parameter can be null.
+     * @param host
+     * @param port
+     * @param domainName
+     * @param tlsConfig
+     * @param extraGlobalHeaders
+     */
+    public ServiceCenterClient(String host, int port, String domainName, TLSConfig tlsConfig,  Map<String, String> extraGlobalHeaders) {
```
liubao68: This module has a lot of same code as kei-client, maybe a common http-client module can be created

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,378 @@
...
+     * Get service-center instances message
+     *
+     * @return MicroserviceInstancesResponse; when error happens,return null
+     */
+    public MicroserviceInstancesResponse getServiceCenterInstances() {
+        try {
+            HttpResponse response = httpClient.getHttpRequest("/registry/health", null, null);
+            if (response.getStatusCode() == HttpStatus.SC_OK) {
+                LOGGER.info("getServiceCenterInstances result = " + response.getContent());
+//                return JSON.parseObject(response.getContent(), MicroserviceInstancesResponse.class);
```
liubao68: health? if this is wrong? And do not log successful response.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
liubao68: This class has a lot of operations. Maybe define an interface like ServiceCenterOperations is better.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
zaneChou1: yes, this is a good idea.

liubao68: Can you fix the format problems and some other problems you want to fix first? And I can merge the two PR's first to help you do refactoring easier.

zaneChou1: ok , thanks. 
The code format is now standardized and some problem is fixed.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
zaneChou1: thanks for reminding, the problem is fixed now.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,381 @@
...
+      HttpResponse response = httpClient
+          .getHttpRequest("/registry/microservices/" + serviceId + "/instances", null, null);
+      if (response.getStatusCode() == HttpStatus.SC_OK) {
+        ObjectMapper mapper = new ObjectMapper();
+        return mapper.readValue(response.getContent(), MicroserviceInstancesResponse.class);
+      } else {
+        LOGGER.error("get service instances list fails, responseStatusCode={}, responseMessage={}, responseContent{}",
+            response.getStatusCode(), response.getMessage(), response.getContent());
+      }
+    } catch (IOException e) {
```
wangqj: I think it would be better to throw these exceptions into the upper layer. The upper layer knows what to do next.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,381 @@
...
+   * @return
+   */
+  public void sendHeartBeats(HeartbeatsRequest heartbeatsRequest) {
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      HttpResponse response = httpClient
+          .putHttpRequest("/registry/heartbeats", null, mapper.writeValueAsString(heartbeatsRequest));
+
+      if (response.getStatusCode() == HttpStatus.SC_OK) {
+        LOGGER.info("SEND HEARTBEATS OK");
```
wangqj: log is not true, it has been successful here.

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -0,0 +1,381 @@
...
+      } else {
+        LOGGER.error("send heartbeats fails, responseStatusCode={}, responseMessage={}, responseContent{}",
+            response.getStatusCode(), response.getMessage(), response.getContent());
```
wangqj: the same as above

client/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/http/HttpTransportImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    HttpDelete httpDelete = new HttpDelete(request.getUrl());
+
+    return doRequest(httpDelete);
+  }
+
+  public HttpResponse doRequest(HttpUriRequest httpRequest) throws IOException {
+    //add header
+    httpRequest.addHeader(HEADER_CONTENT_TYPE, "application/json");
+    httpRequest.addHeader(HEADER_USER_AGENT, "cse-serviceregistry-client/1.0.0");
+    httpRequest.addHeader(HEADER_TENANT_NAME, "default");
```
wangqj: "default" is just a default value, you need to transfer parameter tenant name

zaneChou1: Thanks for comments. And I will create a new issue to follow up and fix it.

zaneChou1:
@wangqj  Thanks for comments. 
This PR may not be modified for the time being. 
After this PR is merged, I will create a new issue to track your problem.

PR #1366, opened 2019-10-28
[SCB-1550] add a java client for ServiceComb-Kie

zaneChou1:
Add a sample Java client of ServiceComb-Kie to ServiceComb project ( ServiceComb-Java-chassis ).
Reference go-client ( https://github.com/apache/servicecomb-kie/blob/master/client/client.go)
Implement the Java client of ServiceComb-Kie.


client/kie-client/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.apache.servicecomb</groupId>
+        <artifactId>java-chassis-parent</artifactId>
+        <version>1.3.0-SNAPSHOT</version>
```
liubao68: Lastest has changed to 2.0.0-SNAPSHOT

liubao68:
```
  <parent>
    <artifactId>java-chassis-parent</artifactId>
    <version>2.0.0-SNAPSHOT</version>
    <relativePath>../parents/default</relativePath>
  </parent>
```

relativePath is required for compiling.

client/kie-client/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.7</version>
```
liubao68: 3td parties should managed by dependency management. see java-chassis-dependencies and take a look at other modules.

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/KieClient.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public String putKeyValue(String key, KVBody kvBody) {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            HttpResponse response = httpClient.putHttpRequest("/kie/kv/" + key, null, mapper.writeValueAsString(kvBody));
+            if (response.getStatusCode() == HttpStatus.SC_OK) {
+                LOGGER.info("putKeyValue result:" + response.getContent());
+                return response.getContent();
+            } else {
+                throw new OperationsException(response.getStatusCode() + response.getMessage() + response.getContent());
```
liubao68: Because you have catched the exception, this statement equals log the error.

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/KieClient.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+        } catch (IOException | OperationsException e) {
+            e.printStackTrace();
```
liubao68: use Logger to log the message

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/KieClient.java
```diff
@@ -0,0 +1,161 @@
...
+     * Get get value of a key
+     *
+     * @param key
+     * @return
+     */
+    public List<KVResponse> getValueOfKey(String key) {
+        try {
+            HttpResponse response = httpClient.getHttpRequest("/kie/kv/" + key, null, null);
+            if (response.getStatusCode() == HttpStatus.SC_OK) {
+                LOGGER.info("getKeyValue result:" + response.getContent());
```
liubao68: Do not log the result of successful operation. Will cause a lot of unneccessary logs.

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/KieClient.java
```diff
@@ -0,0 +1,161 @@
...
+        return null;
+    }
+
+    /**
+     * Get get value of a key
+     *
+     * @param key
+     * @return
+     */
+    public List<KVResponse> getValueOfKey(String key) {
```
liubao68: Since this operation will not throw exception, should document this API that when return null, it is some error happens. And maybe define some XXXException for this module is better for users to use this API.

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/model/KVDoc.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private String _id;
+
+    private String check;
+
+    private String domain;
+
+    private String key;
+
+    private String label_id;
```
liubao68: Are these name required？ Can make them to follow java conventions?

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/http/HttpTransportImpl.java
```diff
@@ -0,0 +1,116 @@
...
+    public HttpClient getHttpClient() {
+        return httpClient;
+    }
+
+    public void setHttpClient(HttpClient httpClient) {
+        this.httpClient = httpClient;
+    }
+
+    public HttpTransportImpl() {
+        httpClient = HttpClients.createDefault();
```
liubao68: Need to create a task to implement TLS. A lot of security settings should be set.

pom.xml
```diff
@@ -125,6 +125,7 @@
     <module>java-chassis-spring-boot</module>
     <module>inspector</module>
     <module>solutions</module>
+    <module>client/kie-client</module>
```
liubao68: Create a subtask to implement dynamic config module for kie.

liubao68:
https://docs.servicecomb.io/java-chassis/zh_CN/build-consumer/3rd-party-service-invoke.html 
Can you refer to document and anaylyse if we can use this feature to implement this client.  Because this will add a lot of governance features for invocation, like flow control, metrics ...

zaneChou1:
@liubao68  Thank you for your comments. I optimized the code based on your suggestion. 
ServiceComb-Kie does not provide TLS documentation and samples now. So the client  not yet done support now and it can be increased later. 
About "Create a subtask to implement dynamic config module for kie", the kie client can support dynamic config later， but it should be handled independently in the client directory. 
How about you?

liubao68:
Maybe your IDE formatter is wrong. Please use formatters in etc folder of this project to format code.

zaneChou1:
@liubao68 thanks for reminding, the code format is now standardized.

client/kie-client/src/main/java/org/apache/servicecomb/kie/client/KieRawClient.java
```diff
@@ -0,0 +1,188 @@
...
+      this.port = DEFAULT_PORT;
+      this.domainName = DOMAIN_NAME;
+    }
+
+    public String getDomainName() {
+      return domainName;
+    }
+
+    public Builder setDomainName(String domainName) {
+      if (domainName == null) {
```
liubao68: I think delete this code is better. Port, Host is the same. [minor]

zaneChou1: The domainName means projectName, mayebe this naming is easy to misunderstand. I can change the name to projectName.
The base api path is "/v1/{project}/kie/kv/" in swagger. The default value of {project} is default, which can also be customized by the user.

Issue #1423, opened 2019-11-30
support service center schemas related API in service-center-client


zaneChou1:
The Java client of the service center currently does not support contract-related APIs. This function is also important, and I recommend adding it.

liubao68:
Can you send PR for this? This module is quite independent of this project.

liubao68:
Close old issue, please feel free to create new one if you meet the problem again.

PR #1504, opened 2019-12-30
[SCB-1697] Add schema API in service-center-client

zaneChou1:
#1423  add service center schemas related API in service-center-client

### DIFF
--- a/clients/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
+++ b/clients/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
@@ -19,6 +19,7 @@ package org.apache.servicecomb.service.center.client;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.http.HttpStatus;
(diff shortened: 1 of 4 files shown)


clients/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -422,4 +425,90 @@ public void sendHeartBeats(HeartbeatsRequest heartbeatsRequest) {
...
+  /**
+   * update schema context of service
+   *
+   * @param serviceId
+   * @param schemaId
+   * @param schemaResponse
+   * @return
+   * @throws OperationException
+   */
+  public Boolean updateServiceSchemaContext(String serviceId, String schemaId, GetSchemaResponse schemaResponse) {
```
liubao68: Maybe ```boolean``` is better.

clients/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
```diff
@@ -422,4 +425,90 @@ public void sendHeartBeats(HeartbeatsRequest heartbeatsRequest) {
...
   }
+
+  /**
+   * Get schemas list of service
+   *
+   * @param serviceId
+   * @return
+   * @throws OperationException
+   */
+  public List<GetSchemaResponse> getServiceSchemasList(String serviceId) {
```
liubao68: ```GetSchemaResponse``` is used as an ```request``` in ```updateServiceSchemaContext```, it's better to change to a new name. e.g. ```SchemaInfo```

zaneChou1: Thanks for your suggestion, it has been modified.

clients/service-center-client/src/main/java/org/apache/servicecomb/service/center/client/ServiceCenterClient.java
zaneChou1: Thanks for your suggestion, it has been modified.

PR #1516, opened 2020-01-09
[SCB-1710] catch the throwable from the scheduled pulling task

yhs0092:
https://issues.apache.org/jira/browse/SCB-1710

solution:
- add a wrapper class `SuppressedRunnableWrapper` to decorate the scheduled tasks, to ensure that no `Throwable` is thrown out of the tasks. This wrapper can be reused later.
- wrap the `org.apache.servicecomb.serviceregistry.consumer.AppManager#pullInstances` scheduled task by `SuppressedRunnableWrapper`

### DIFF
--- a/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/concurrency/RunnableWrapperTest.java
+++ b/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/concurrency/RunnableWrapperTest.java
@@ -31,27 +31,36 @@ public class RunnableWrapperTest {
    * should not interrupt the scheduled tasks.
    */
   @Test
-  public void run() throws InterruptedException {
+  public void run() {
     ScheduledThreadPoolExecutor scheduledThreadPoolExecutor = new ScheduledThreadPoolExecutor(1);
     CountDownLatch countDownLatch = new CountDownLatch(3);
-    scheduledThreadPoolExecutor.scheduleAtFixedRate(
-        new SuppressedRunnableWrapper(
-            () -> {
-              countDownLatch.countDown();
-              switch ((int) countDownLatch.getCount()) {
-                case 2:
-                  throw new Error("Normal case, this is a mocked error");
-                case 1:
-                  throw new IllegalStateException("Normal case, this is a mocked exception");
-                default:
-              }
-            }
-        ),
-        1,
-        1,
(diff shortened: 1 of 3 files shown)


foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/concurrency/RunnableWrapperTest.java
```diff
@@ -0,0 +1,57 @@
...
+                default:
+              }
+            }
+        ),
+        1,
+        1,
+        TimeUnit.MILLISECONDS);
+
+    countDownLatch.await(1000, TimeUnit.MILLISECONDS);
+    assertEquals(0, countDownLatch.getCount());
```
liubao68: it's better to call scheduledThreadPoolExecutor.shutDownNow to terminate the thread pool

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/RemoteServiceRegistry.java
```diff
@@ -50,18 +52,16 @@ public RemoteServiceRegistry(EventBus eventBus, ServiceRegistryConfig serviceReg
...
-              try {
-                setName("Service Center Task [" + task.toString() + "[" + this.getId() + "]]");
-                super.run();
-              } catch (Throwable e) {
-                LOGGER.error("task {} execute error.", getName(), e);
-              }
-            }
-          };
+        new ThreadFactory() {
+          private int taskId;
```
liubao68: It's better to assign a default value explicitly

foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/concurrency/RunnableWrapperTest.java
yhs0092: Done. the try catch block is added and the thread pool is shutdown in the `finally` block.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/registry/RemoteServiceRegistry.java
yhs0092: Done. pls check it.

PR #1555, opened 2020-02-03
[SCB-1643] add a slution to integrate with huaweicloud

liubao68:

---

### DIFF
--- a/solutions/solution-basic/pom.xml
+++ b/solutions/solution-basic/pom.xml
@@ -29,6 +29,21 @@
   <artifactId>solution-basic</artifactId>
   <name>Java Chassis::Solutions::BASIC</name>
 
+  <properties>
+    <huaweicloudsdk-version>2.5.0</huaweicloudsdk-version>
+  </properties>
+
+  <repositories>
+    <repository>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>huaweicloudsdk-releases</id>
+      <name>huaweicloudsdk</name>
+      <url>https://repo.huaweicloud.com/repository/maven/huaweicloudsdk/</url>
+    </repository>
+  </repositories>
+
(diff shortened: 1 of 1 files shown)


WillemJiang:
How does the user use it?
It should be optional.

liubao68:
Users use it without notice. By adding this, samples provided in servicecomb-samples can run both in local and cloud

liubao68:
@WillemJiang I check the meaning of 'optional' in maven, this can not fullfil the purpose of the PR. By adding this, I want users include the file without notice to run across local and cloud. If 'optional' is defined, users still need to add this dependency in their pom. This is quite like not have this PR.

WillemJiang:
> @WillemJiang I check the meaning of 'optional' in maven, this can not fullfil the purpose of the PR. By adding this, I want users include the file without notice to run across local and cloud. If 'optional' is defined, users still need to add this dependency in their pom. This is quite like not have this PR.

'optional' means it's not used by default, I'm not talking about the maven optional scope, I'm talking about  if the user want to write the application to run with Huawei Cloud , he can add the dependency pom, but when we do the Apache release, we don't include this dependency pom by default.

liubao68:
> > @WillemJiang I check the meaning of 'optional' in maven, this can not fullfil the purpose of the PR. By adding this, I want users include the file without notice to run across local and cloud. If 'optional' is defined, users still need to add this dependency in their pom. This is quite like not have this PR.
> 
> 'optional' means it's not used by default, I'm not talking about the maven optional scope, I'm talking about if the user want to write the application to run with Huawei Cloud , he can add the dependency pom, but when we do the Apache release, we don't include this dependency pom by default.

```solution-basic``` is optional to users, they may not use this pom. The purpose of adding this pom is for easier of deploy. Or users will always be noticed to include the dependency and configure repositories.

PR #1561, opened 2020-02-07
[SCB-1748] support colon in path value

GuoYL123:

---


GuoYL123:
WIP

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java
```diff
@@ -52,6 +54,9 @@ public Object getValue(HttpServletRequest request) {
       if (value == null) {
         return null;
       }
+      if (value.contains(":")) {
+        return convertValue(URLDecoder.decode(value, "UTF-8"), targetType);
```
liubao68: I think this modification is not correct. You can add a test case for HttpUtils, like 
```
  @Test
  public void uriEncode_sss() {
    String encoded = HttpUtils.uriEncodePath("a:b");
    Assert.assertEquals("a:b", encoded);
    Assert.assertEquals("a:b", HttpUtils.uriDecodePath(encoded));
  }

```

And make this case pass.

liubao68:
[This](https://stackoverflow.com/questions/724043/http-url-address-encoding-in-java) can be a reference

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java
GuoYL123: I notic there is a case org.apache.servicecomb.foundation.common.http.TestHttpUtils#uriDecode_failed
to show a error when the param of HttpUtils.uriDecodePath contain ":".

I think modify  the HttpUtils.uriDecodePat will change the previous  logic.

liubao68: Maybe this test case is not correct. Checked the [RFC](https://www.ietf.org/rfc/rfc2396.txt) 

```
3.3. Path Component

   The path component contains data, specific to the authority (or the
   scheme if there is no authority component), identifying the resource
   within the scope of that scheme and authority.

      path          = [ abs_path | opaque_part ]

      path_segments = segment *( "/" segment )
      segment       = *pchar *( ";" param )
      param         = *pchar

      pchar         = unreserved | escaped |
                      ":" | "@" | "&" | "=" | "+" | "$" | ","

   The path may consist of a sequence of path segments separated by a
   single slash "/" character.  Within a path segment, the characters
   "/", ";", "=", and "?" are reserved.  Each path segment may include a
   sequence of parameters, indicated by the semicolon ";" character.
   The parameters are not significant to the parsing of relative
   references.
```

liubao68: I checked the code again. May be "encodePathParam" and "decodePathParam" should be used here. You need check details usage of "uriEncodePath" and "uriDecodePath". They are different.

GuoYL123: Add decodePathParam method.

common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java
```diff
@@ -17,7 +17,9 @@
 
 package org.apache.servicecomb.common.rest.codec.param;
 
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
+import java.net.URLDecoder;
```
liubao68: remove unused import

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java
```diff
@@ -94,6 +97,14 @@ public static String encodePathParam(String pathParam) {
     return UrlEscapers.urlPathSegmentEscaper().escape(pathParam);
   }
 
+  public static String decodePathParam(String pathParam) throws UnsupportedEncodingException {
+    String res = uriDecodePath(pathParam);
+    if (StringUtils.isEmpty(res)) {
+      return URLDecoder.decode(pathParam, "UTF-8");
```
liubao68: Is ``` URLDecoder.decode(pathParam, "UTF-8");``` is enough?  I think it is enough to use URLEncoder and URLDecoder to encode/decode path segments, and use URI to encode/decode path.

foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java
```diff
@@ -94,6 +97,14 @@ public static String encodePathParam(String pathParam) {
...
+    }
+    return res;
+  }
+
   public static String uriDecodePath(String path) {
     if (path == null) {
```
liubao68: This function seems not correct. According to the java docs, URL(path) will not parse the path, users should use 
```new URI(null, null, path, null);``` to parse the parse

File: foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java (hunk 3 above)
Comment (GuoYL123) [reply foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java]: `URLDecoder.decode(pathParam, "UTF-8")  `can not handle the '+' .
there is a ut show that : org.apache.servicecomb.common.rest.codec.param.TestPathProcessor#testGetPlus.

File: foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java (hunk 4 above)
Comment (GuoYL123) [reply foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/http/HttpUtils.java]: use `new URI(null, null, path, null);` will cause ut failed:
org.apache.servicecomb.common.rest.codec.param.TestPathProcessor#testGetSpaceEncoded

`new URI(null, null, path, null);` can not convert '%20' to ' '

File: common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java (hunk 2 above)
Comment (GuoYL123) [reply common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/param/PathProcessorCreator.java]: done

**Comment by liubao68** [comment]
I think the current modification is not correct, need more analyse. For reference, you can look at 
```org.springframework.web.util.UriUtils```

and This test case seems not correct
```
  @Test
  public void uriDecode_failed() {
    expectedException.expect(IllegalArgumentException.class);
    expectedException
        .expectMessage(Matchers.is("uriDecode failed, path=\":\"."));
    expectedException.expectCause(Matchers.instanceOf(URISyntaxException.class));

    HttpUtils.uriDecodePath(":");
  }
```

GuoYL123:
Thanks for the suggestion .And I found ` StringUtils.uriDecode(value, StandardCharsets.UTF_8)` can resolve the problem. 
Maybe `HttpUtils.uriDecodePath` is not correct , but i can avoid to touch it. :)
Should I delete this unit test?

liubao68:
```org.springframework.web.util.UriUtils``` has very nice api design. We can refer this and use the correct one in the right place. ```HttpUtils``` can disign it's API like this class.

PR #1600, opened 2020-02-27
[SCB-1691] Refactor registry module

yhs0092:
main features:

- one microservice instance can hold multiple `ServiceRegistry` instances, which are managed in `RegistryUtils`, so that one instance can connect to multiple service center clusters
- when microservice instance is shutdown, the corresponding instance record in service center is switched to `DOWN` status and the shutdown precedure is blocked for a while, so that consumer can get enough time to get notified that this instance is stopping. The block period length is determined by the config item `servicecomb.boot.turnDown.waitInSeconds`

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
@@ -22,6 +22,8 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.regex.Matcher;
 
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.servicecomb.config.ConfigUtil;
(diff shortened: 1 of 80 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -20,62 +20,120 @@
...
+
+  private static AppManager appManager = new AppManager();
+
+  private static InstanceCacheManager instanceCacheManager = new InstanceCacheManagerNew(appManager);
+
+  private static final Map<String, ServiceRegistryConfig> EXTRA_SERVICE_REGISTRY_CONFIGS = new LinkedHashMap<>();
+
+  private static final Map<String, ServiceRegistry> EXTRA_SERVICE_REGISTRIES = new LinkedHashMap<>();
+
+  static AggregateServiceRegistryCache aggregateServiceRegistryCache;
```
GuoYL123: should be private？

yhs0092: Done, pls review it

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -214,24 +280,163 @@ private static IpPort genPublishIpPort(String schema, IpPort ipPort) {
 
   public static List<MicroserviceInstance> findServiceInstance(String appId, String serviceName,
       String versionRule) {
-    return serviceRegistry.findServiceInstance(appId, serviceName, versionRule);
+    MicroserviceCache serviceCache = aggregateServiceRegistryCache.findServiceCache(
+        MicroserviceCacheKey.builder()
```
Matrix0xCC, line 284: 这个是不兼容的修改？
修改之后`versionRule`被忽略了，为何不直接去掉这个参数？
这个类目前被业务使用，升级之后没有发现这个接口的功能变更，导致了一个问题。

PR #1632, opened 2020-03-14
 [SCB-1797] Fix marker leak problem when using slf4j1.8 or higher with log4j2

Nick-The-Uncharted:

---

### DIFF
--- a/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/log/LogMarkerLeakFixUtils.java
+++ b/foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/log/LogMarkerLeakFixUtils.java
@@ -20,6 +20,8 @@ import java.lang.reflect.Field;
 import java.util.concurrent.ConcurrentMap;
 
 import org.apache.commons.lang3.reflect.FieldUtils;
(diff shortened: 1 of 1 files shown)


liubao68:
Thanks for this fix. Because there are some [pending issues](https://issues.apache.org/jira/browse/SCB-1738) need to handle for working with higher versions of JDK, need find a better solution for this.  

As I know, this fix is because java-chassis using Marker to log invocation data, while log4j2 designed Marker to be static data, thus cause OOM. This usage may not be correct. So we need to find a better solution.

Nick-The-Uncharted:
I am not familiar with higher version of JDK, do you mean better solution that not using reflection?

liubao68:
> I am not familiar with higher version of JDK, do you mean better solution that not using reflection?

Yes. I am thinking some others ways to adress this problem tomorrow.  The recent PR has removed log4j2 fixs to a seperate module.

liubao68:
I submit a PR for this : https://github.com/apache/servicecomb-java-chassis/pull/1642/files
Please add your review comments if some problems.

Nick-The-Uncharted:
@liubao68 Your PR use MDC instead of Marker to trace user, I dont actully use marker to filter logs so this seems okay to me.

PR #1658, opened 2020-03-23
[SCB-1816] print the instance endpoints in the end of the start up log

yhs0092:


core/src/main/java/org/apache/servicecomb/core/bootup/ServiceInformationCollector.java
```diff
@@ -29,7 +31,16 @@ public String collect() {
...
         + "\n" + "Environment: " + RegistryUtils.getMicroservice().getEnvironment()
         + "\n" + "Service ID: " + RegistryUtils.getMicroserviceInstance().getServiceId()
-        + "\n" + "Instance ID: " + RegistryUtils.getMicroserviceInstance().getInstanceId();
+        + "\n" + "Instance ID: " + RegistryUtils.getMicroserviceInstance().getInstanceId()
+        + "\n" + "Endpoints: " + getEndpoints();
+  }
+
+  private String getEndpoints() {
+    try {
+      return JsonUtils.writeValueAsString(RegistryUtils.getMicroserviceInstance().getEndpoints());
```
wujimin: 换成io.vertx.core.spi.json.JsonCodec#toString(java.lang.Object)，就不用自己写了

yhs0092: Done, please review it

PR #1667, opened 2020-03-27
[SCB-1829] provide integration test for RPC consumer to read response header

wujimin:


liubao68:
Can you add an example(integration tests) for this feature? You can simply add in any demos, they will be run in docker when CI builds. And I'll create an issue to add documents for this feature.

liubao68:
https://issues.apache.org/jira/browse/SCB-1836

wujimin:
> Can you add an example(integration tests) for this feature? You can simply add in any demos, they will be run in docker when CI builds. And I'll create an issue to add documents for this feature.

ok, i will add one IT case in this PR.

integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestResponseHeader.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.apache.servicecomb.it.testcase;
+
+import static org.junit.Assert.assertEquals;
+
+import org.apache.servicecomb.it.Consumers;
+import org.junit.Test;
+import org.springframework.http.ResponseEntity;
+
+public class TestResponseHeader {
```
liubao68: not add `ITJunitUtils.runWithHighwayAndRest(TestReponseHeader.class)` ?

liubao68:
already have this test case in demo-springmvc. close this PR.

PR #1697, opened 2020-04-08
[SCB-1857] RequestLog增加仅针对CONSUMER类型的InvocationFinishEvent处理

whg903:


heyile:
I know why I can not reproduce the problem.

my request log pattern is **%h %SCB-transport - - %t %r %s %D**

when it prints %SCB-transport item, it will throw null pointer exception. as the invocation#referenceConfig is null

PR #1704, opened 2020-04-09
[SCB-1863] toke the revision when pull from kie

GuoYL123:

---

### DIFF
--- a/dynamic-config/config-kie/src/main/java/org/apache/servicecomb/config/kie/client/KieClient.java
+++ b/dynamic-config/config-kie/src/main/java/org/apache/servicecomb/config/kie/client/KieClient.java
@@ -57,6 +57,8 @@ public class KieClient {
 
   private static final int LONG_POLLING_WAIT_TIME_IN_SECONDS = 30;
 
+  private static String revision = "0";
+
   private static final KieConfig KIE_CONFIG = KieConfig.INSTANCE;
 
   private final int refreshInterval = KIE_CONFIG.getRefreshInterval();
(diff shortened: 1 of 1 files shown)


liubao68:
please rebase and fix the problems

dynamic-config/config-kie/src/main/java/org/apache/servicecomb/config/kie/client/KieClient.java
```diff
@@ -115,8 +117,8 @@ void refreshConfig(CountDownLatch latch) {
       String path = "/v1/"
           + KieConfig.INSTANCE.getDomainName()
           + "/kie/kv?label=app:"
-          + KieConfig.INSTANCE.getAppName();
-
+          + KieConfig.INSTANCE.getAppName()
+          + "&revision=" + revision;
```
liubao68, line 121: revision is a static member, it is not good.

dynamic-config/config-kie/src/main/java/org/apache/servicecomb/config/kie/client/KieClient.java
```diff
@@ -57,6 +57,8 @@
 
   private static final int LONG_POLLING_WAIT_TIME_IN_SECONDS = 30;
 
+  private static String revision = "0";
```
liubao68, line 60: IS_FIRST_PULL  revision  is better to use instance property, not static. When KieClient has only one instance, instance property is better. When KieClient has many instances(not possible), static is not good eitheir.

PR #1769, opened 2020-05-21
[SCB-1922]Use DynamicPropertyFacotry way to read environment variables

jungan21:

---

### DIFF
--- a/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
+++ b/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
@@ -16,7 +16,7 @@
  */
 package org.apache.servicecomb.zeroconfig;
 
-import org.apache.servicecomb.foundation.common.utils.BeanUtils;
+import com.netflix.config.DynamicPropertyFactory;
 import org.apache.servicecomb.serviceregistry.ServiceRegistry;
 import org.apache.servicecomb.serviceregistry.client.ServiceRegistryClient;
 import org.apache.servicecomb.serviceregistry.config.ServiceRegistryConfig;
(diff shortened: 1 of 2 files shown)


service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
```diff
@@ -52,24 +52,23 @@
...
 
     // same mechanism as Local registry to enable the Zero Config registry
-    String flag = System.getProperty(ZeroConfigRegistryConstants.ZERO_CONFIG_REGISTRY_FLAG);
-    if (flag != null && flag.equals(ZeroConfigRegistryConstants.ZERO_CONFIG_REGISTRY_ENABLE_FLAG)) {
-      this.applicationContext = applicationContext;
-      BeanUtils.setContext(applicationContext);
+    boolean enable = DynamicPropertyFactory.getInstance()
+        .getBooleanProperty("servicecomb.zeroconfig.enabled", false).get();
+
+    if (enable) {
```
wujimin, line 64: better to be: not enabled then return

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
```diff
@@ -52,24 +52,23 @@
...
       ServerUtil.init();
       ClientUtil.init();
 
       try {
         this.multicastSocket = new MulticastSocket();
       } catch (IOException e) {
-        LOGGER.error("Failed create MulticastSocket object", e);
+        LOGGER.error("Failed to create MulticastSocket object", e);
```
wujimin, line 71: better to throw exception and prevent bootup

jungan21: agreed.

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
jungan21: agreed.

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/ZeroConfigRegistryApplicationListener.java
```diff
@@ -52,32 +53,36 @@
...
-        LOGGER.error("Failed create MulticastSocket object", e);
-      }
-
-      Function<ServiceRegistry, ServiceRegistryClient> registryClientConstructor =
-          serviceRegistry -> new ZeroConfigRegistryClientImpl(new ZeroConfigRegistryService(),
-              multicastSocket, new RestTemplate());
-
-      ServiceRegistryConfig.INSTANCE.setServiceRegistryClientConstructor(registryClientConstructor);
+    boolean enable = DynamicPropertyFactory.getInstance()
+        .getBooleanProperty("servicecomb.zeroconfig.enabled", false).get();
```
liubao68, line 63: I think default value to 'true' is better. Because when users use zeroconfig, they have to add pom dependency implicitly. It is falst when zeroconfig module is not included. I set 'local/service center' default to 'true'. see https://github.com/apache/servicecomb-java-chassis/blob/master/service-registry/registry-local/src/main/java/org/apache/servicecomb/localregistry/LocalRegistration.java

jungan21: I see. let me change default value to true

PR #1778, opened 2020-05-26
[SCB-1936] change findbug dependency to provided

wujimin:


liubao68:
Does jsr 305 required for running java-chassis in vert.x mode?

wujimin:
no need for runtime, add them just because compile warning

PR #1794, opened 2020-05-29
[SCB-1947] define Filter and filter chain

wujimin:


core/src/test/java/org/apache/servicecomb/core/filter/SimpleRetryFilter.java
```diff
@@ -0,0 +1,89 @@
...
+  public CompletableFuture<Response> onFilter(Invocation invocation, FilterNode nextNode) {
+    return new RetrySession(invocation, nextNode).run();
+  }
+
+  protected Throwable unwrapException(Throwable throwable) {
+    return Exceptions.unwrapIncludeInvocationException(throwable);
+  }
+
+  protected boolean isRetryException(Throwable throwable) {
+    return !(throwable instanceof IOException);
```
liubao68, line 45: How do you judge which is reatrable？ This is different than current DefaultRetryExtensionsFactory.isRetriableException implementation.

core/src/main/java/org/apache/servicecomb/core/filter/FilterNode.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private final Filter filter;
+
+  private FilterNode nextNode;
+
+  public FilterNode(Filter filter) {
+    this.filter = filter;
+  }
+
+  public void setNextNode(FilterNode nextNode) {
```
liubao68, line 53: It's better to be private.

wujimin: done

PR #1800, opened 2020-05-30
[SCB-1961] add rest server codec filter

wujimin:


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/AbstractRestInvocation.java
```diff
@@ -268,16 +268,8 @@ protected void sendResponseQuietly(Response response) {
...
-      for (Entry<String, List<Object>> entry : response.getHeaders().getHeaderMap().entrySet()) {
-        for (Object value : entry.getValue()) {
-          if (!entry.getKey().equalsIgnoreCase(HttpHeaders.CONTENT_LENGTH)
-              && !entry.getKey().equalsIgnoreCase("Transfer-Encoding")) {
-            responseEx.addHeader(entry.getKey(), String.valueOf(value));
-          }
-        }
-      }
-    }
+    RestServerCodecFilter.encodeHeader(response.getHeaders().getHeaderMap(), responseEx);
```
liubao68, line 271: `encodeHeader` name is not good. It's do not do any encoding.

wujimin: done

PR #1816, opened 2020-06-04
[SCB-1982] add validator filter

wujimin:


core/src/main/java/org/apache/servicecomb/core/filter/impl/ParameterValidatorFilter.java
```diff
@@ -0,0 +1,88 @@
...
+  private static final String PARAM_VALIDATION_ENABLED = "servicecomb.filters.validator.enabled";
+
+  private ExecutableValidator validator;
+
+  private final DynamicBooleanProperty paramValidationEnabled = DynamicPropertyFactory.getInstance()
+      .getBooleanProperty(PARAM_VALIDATION_ENABLED, false);
+
+  public ParameterValidatorFilter() {
+    paramValidationEnabled
+        .addCallback(() -> LOGGER.info("{} changed to {}", PARAM_VALIDATION_ENABLED, paramValidationEnabled.get()));
```
liubao68, line 59: seams no need to add callback. paramValidationEnabled only used in initialization(Filter enable apply to initinization too).

wujimin: ok, only add when necessary

wujimin: done

PR #1825, opened 2020-06-09
[SCB-1995]Fix incorrect message issue identifed through testing

jungan21:

---

### DIFF
--- a/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
+++ b/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
@@ -16,6 +16,7 @@
  */
 package org.apache.servicecomb.zeroconfig.server;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.servicecomb.foundation.common.concurrent.ConcurrentHashMapEx;
 import org.apache.servicecomb.zeroconfig.client.ClientUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
```diff
@@ -179,12 +179,12 @@ private static ServerMicroserviceInstance buildServerMicroserviceInstanceFromMap
...
-      byte[] buffer = new byte[DATA_PACKET_BUFFER_SIZE];
       multicastSocket = new MulticastSocket(PORT);
       group = InetAddress.getByName(GROUP);
       multicastSocket.joinGroup(group); // need to join the group to be able to receive the data
 
       while (true) {
+        byte[] buffer = new byte[DATA_PACKET_BUFFER_SIZE];
         DatagramPacket receivePacketBuffer = new DatagramPacket(buffer, buffer.length);
         multicastSocket.receive(receivePacketBuffer);
         String receivedPacketString = new String(receivePacketBuffer.getData(), ENCODE);
```
liubao68, line 190: This logic seams not correct. Maybe you'd check the length of data actually received and using the correct data.

jungan21: good point. Unfortunately, based on the Multicast working mechanism, we have to create a buffer before data is received.  After data is received, then buffer is filled with the received data.   So we can't create a buffer with size exactly same as the received data.

By the way, it's also mentioned in the JDK source code comments: 
https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/net/DatagramSocket.java#L698-L700

liubao68: Maybe you misunderstanding my point. You should do this 

```
 multicastSocket.receive(receivePacketBuffer);
int len = multicastSocket.getLength() ;
if(len > 0) {
   String receivedPacketString = new String(receivePacketBuffer.getData(), 0, len, ENCODE);
}
```

PR #1827, opened 2020-06-10
[SCB-1997] Small improvement when Muticastscoket receive data

jungan21:

---

### DIFF
--- a/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
+++ b/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
@@ -172,57 +172,36 @@ public class ServerUtil {
     return map;
   }
 
+  private static void initMulticastSocket() throws IOException {
+    multicastSocket = new MulticastSocket(PORT);
+    group = InetAddress.getByName(GROUP);
+    multicastSocket.joinGroup(group); // need to join the group to be able to receive the data
+  }
+
   private static void startListenerForRegisterUnregisterEvent() {
     try {
       byte[] buffer = new byte[DATA_PACKET_BUFFER_SIZE];
-      multicastSocket = new MulticastSocket(PORT);
-      group = InetAddress.getByName(GROUP);
-      multicastSocket.joinGroup(group); // need to join the group to be able to receive the data
-
+      initMulticastSocket();
       while (true) {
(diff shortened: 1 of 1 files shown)


service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
```diff
@@ -172,16 +172,29 @@ private static ServerMicroserviceInstance buildServerMicroserviceInstanceFromMap
...
       byte[] buffer = new byte[DATA_PACKET_BUFFER_SIZE];
-      multicastSocket = new MulticastSocket(PORT);
-      group = InetAddress.getByName(GROUP);
-      multicastSocket.joinGroup(group); // need to join the group to be able to receive the data
-
+      initMulticastSocket();
       while (true) {
         DatagramPacket receivePacketBuffer = new DatagramPacket(buffer, buffer.length);
-        multicastSocket.receive(receivePacketBuffer);
+        try {
```
liubao68, line 187: Maybe need look at the outer `try', if `IOException` is thrown, the tast is exit.

jungan21: Correct. But the outer try/catch(IOException) is for initializing the Multicastsocket object ......the failure should rarely happen. I don't think we need to handle this for now

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/server/ServerUtil.java
```diff
@@ -172,16 +172,29 @@ private static ServerMicroserviceInstance buildServerMicroserviceInstanceFromMap
...
 
+  private static void initMulticastSocket() throws IOException {
+    multicastSocket = new MulticastSocket(PORT);
+    group = InetAddress.getByName(GROUP);
+    multicastSocket.joinGroup(group); // need to join the group to be able to receive the data
+  }
+
   private static void startListenerForRegisterUnregisterEvent() {
     try {
       byte[] buffer = new byte[DATA_PACKET_BUFFER_SIZE];
```
liubao68, line 183: cyclomatic complexity of `startListenerForRegisterUnregisterEvent` is a bit too high, can be refactored.

PR #1869, opened 2020-07-07
[SCB-2035]Use vertx Json and Pojo to replace own serizalization mechanism

jungan21:

---

### DIFF
--- a/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
+++ b/service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
@@ -16,6 +16,7 @@
  */
 package org.apache.servicecomb.zeroconfig.client;
 
+import io.vertx.core.json.Json;
 import org.apache.servicecomb.registry.api.registry.Microservice;
 import org.apache.servicecomb.registry.api.registry.MicroserviceInstance;
 import org.apache.servicecomb.registry.api.registry.MicroserviceInstanceStatus;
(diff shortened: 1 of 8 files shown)


service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
```diff
@@ -41,86 +39,68 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ClientUtil.class);
 
-  private static ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+  public static ClientUtil INSTANCE = new ClientUtil();
```
liubao68, line 42: public static field should be final.

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
```diff
@@ -41,86 +39,68 @@
...
 
-  private static Microservice microserviceSelf = new Microservice();
+  private static ServerMicroserviceInstance serviceInstanceForHeartbeat = null;
```
liubao68, line 44: This is a sgingletan class, a static member seems a little bit hard to figure out if it is correct.

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
jungan21: agreed.

service-registry/registry-zero-config/src/main/java/org/apache/servicecomb/zeroconfig/client/ClientUtil.java
jungan21: changed to non-static ....

PR #1913, opened 2020-08-08
[SCB-2062] jaxrs mode support PATCH

wujimin:


swagger/swagger-generator/generator-jaxrs/src/test/resources/schemas/patch.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
liubao68, line 1: add license

wujimin: done

PR #1936, opened 2020-09-01
[SCB-2077] fix instance isolation trying chance not released problem

yhs0092:
Please see the details in https://issues.apache.org/jira/browse/SCB-2077

### DIFF
--- a/handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
+++ b/handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
@@ -18,9 +18,10 @@
 package org.apache.servicecomb.loadbalance;
 
 import java.time.Clock;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.servicecomb.core.Invocation;
 import org.apache.servicecomb.foundation.common.utils.TimeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


yhs0092:
Current fix solution:
- change the type of `ServiceCombServerStats#globalAllowIsolatedServerTryingFlag` from `AtomicBoolean` to `AtomicReference<TryingIsolatedServerMarker>`. The newly defined `TryingIsolatedServerMarker` holds the `Invocation` instance which is holding the trying chance and the timestamp.
- in the normal case, when the `Invocation` is finished(checked by `Invocation#isFinished`), the trying chance is released.
- in case that some unknown error occurs, causing the invocation procedure ends with the `Invocation#finished` flag not set to true, an new configuration `servicecomb.loadbalance.isolation.maxSingleTestWindow` is added to limit the max occupation period of the global trying chance.

yhs0092:
I've also considered about other solutions.
An alternative choice is to use `SoftReference` to hold the `Invocation` instance in `TryingIsolatedServerMarker`. In case that the `Invocation#finished` is not set to true properly, the trying chance can be released when the `SoftReference` returns `null` (which means the `Invocation` instance is garbage collected --- the invocation is over).
The advantage of this solution is the new config item `servicecomb.loadbalance.isolation.maxSingleTestWindow` is not needed. While there is also the disadvantage. Since the result of `SoftReference#get` gets null after the garbage collection happens, this insurance policy depends on the timing of GC. The result seems uncontrollable.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
```diff
@@ -35,15 +36,14 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ServiceCombServerStats.class);
 
+  private final Object lock = new Object();
```
liubao68, line 39: lock is never read

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
```diff
@@ -84,20 +84,35 @@ private void init(){
...
   }
 
   /**
    * Applying for a trying chance for the isolated server. There is only 1 trying chance globally concurrently.
    *
    * @return true if the chance is applied successfully, otherwise false
    */
-  public static boolean applyForTryingChance() {
-    return isolatedServerCanTry() && globalAllowIsolatedServerTryingFlag.compareAndSet(true, false);
+  public static boolean applyForTryingChance(Invocation invocation) {
```
liubao68, line 99: This method is not thread safe.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
yhs0092: The `lock` is used in method `ensureWindow`. It does not get involved in this issue.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServiceCombServerStats.java
yhs0092: The thread safty is ensured by the CAS operation `AtomicReference#compareAndSet`

PR #1953, opened 2020-09-14
[SCB-2082] add separated switches for metricsEndpoint and healthEndpoint

yhs0092:
- add config item `servicecomb.health.endpoint.enabled` as the switch of healthEndpoint
- add config item `servicecomb.metrics.endpoint.enabled` as the switch of metricsEndpoint
- the default value of these two item is `true`

### DIFF
--- a/metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HealthBootListener.java
+++ b/metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HealthBootListener.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.servicecomb.metrics.core;
+
+import org.apache.servicecomb.core.BootListener;
+import org.apache.servicecomb.metrics.core.publish.HealthCheckerRestPublisher;
+import org.springframework.stereotype.Component;
+
+import com.netflix.config.DynamicPropertyFactory;
+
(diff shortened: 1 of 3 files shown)


metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/MetricsBootListener.java
```diff
@@ -43,13 +42,10 @@ public SlowInvocationLogger getSlowInvocationLogger() {
 
   @Override
   public void onBeforeProducerProvider(BootEvent event) {
-    if (!DynamicPropertyFactory.getInstance().getBooleanProperty("servicecomb.metrics.endpoint.enabled", true).get()) {
+    if (!DynamicPropertyFactory.getInstance().getBooleanProperty("servicecomb.metrics.endpoint.enabled", false).get()) {
```
liubao68, line 45: Maybe keep default value to true is better?

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HealthBootListener.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.servicecomb.metrics.core.publish.HealthCheckerRestPublisher;
+import org.springframework.stereotype.Component;
+
+import com.netflix.config.DynamicPropertyFactory;
+
+@Component
+public class HealthBootListener implements BootListener {
+  @Override
+  public void onBeforeProducerProvider(BootEvent event) {
+    if (!DynamicPropertyFactory.getInstance().getBooleanProperty("servicecomb.health.endpoint.enabled", false).get()) {
```
liubao68, line 29: Maybe keep default value to true is better?

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/MetricsBootListener.java
yhs0092: Under the security sensitive circumstances, the users only want to expose their business REST interfaces. If we keep the default value to be `true`, they may expose these two REST interfaces by mistake.

liubao68: I think these two interfaces do not have any sensitive information. And if the application is secrured , the two interfaces will be secured too. I don't know why explode the two interface will cause security issue.

yhs0092: Done, please review

metrics/metrics-core/src/main/java/org/apache/servicecomb/metrics/core/HealthBootListener.java
yhs0092: Done, please review

liubao68 (approved):

PR #1994, opened 2020-10-12
[SCB-2089] 支持服务中心rbac认证

jeho0815:

---

### DIFF
--- a/dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
+++ b/dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
@@ -31,7 +31,6 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.ServiceLoader;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -45,6 +44,7 @@ import org.apache.servicecomb.foundation.common.event.EventManager;
 import org.apache.servicecomb.foundation.common.net.IpPort;
 import org.apache.servicecomb.foundation.common.net.NetUtils;
 import org.apache.servicecomb.foundation.common.utils.JsonUtils;
+import org.apache.servicecomb.foundation.common.utils.SPIServiceUtils;
 import org.apache.servicecomb.foundation.vertx.client.http.HttpClientWithContext;
 import org.apache.servicecomb.foundation.vertx.client.http.HttpClients;
 import org.slf4j.Logger;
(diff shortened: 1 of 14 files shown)


foundations/foundation-auth/src/main/java/org/apache/servicecomb/foundation/auth/rbac/Cipher.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
```
liubao68, line 3: This class and implementation can move to foundation-common, do not need module foundation-auth

foundations/foundation-auth/src/main/java/org/apache/servicecomb/foundation/auth/rbac/TokenAuthHeaderProvider.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
liubao68, line 2: This class and others move to huawei-cloud/servicestage module, do not need create a new module.

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
```diff
@@ -95,8 +95,8 @@
 
   private boolean isWatching = false;
 
-  private final ServiceLoader<AuthHeaderProvider> authHeaderProviders =
-      ServiceLoader.load(AuthHeaderProvider.class);
+  private final List<AuthHeaderProvider> authHeaderProviders =
+      SPIServiceUtils.getSortedService(AuthHeaderProvider.class);
```
liubao68, line 99: configure center is initialized before service center, will this cause problem?

foundations/foundation-auth/src/main/java/org/apache/servicecomb/foundation/auth/rbac/Cipher.java
jeho0815: has removed to foundation-common

foundations/foundation-auth/src/main/java/org/apache/servicecomb/foundation/auth/rbac/TokenAuthHeaderProvider.java
jeho0815: has removed to registry-service-center module, they has more closer relationship

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterClient.java
jeho0815: the dependency has some problem, we can refact later

PR #2007, opened 2020-10-21
SCB-2097 Synchronize some bug fix codes

develpoerX:

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceStatusSyncTask.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceStatusSyncTask.java
@@ -28,64 +28,64 @@ import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 
 public class MicroserviceInstanceStatusSyncTask extends AbstractTask {
-	private static final Logger LOGGER = LoggerFactory.getLogger(MicroserviceInstanceStatusSyncTask.class);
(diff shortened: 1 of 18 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceStatusSyncTask.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.servicecomb.serviceregistry.task;
```
liubao68, line 1: add license to to file

liubao68:
Please create more specific issues for this PR. 

e.g

1. upgrade xxx from x to y. e.g.  https://issues.apache.org/jira/browse/SCB-2018
2. fix bug: xxxx

You can still use one PR to fix them all. These issues will used in Release Notes.

service-registry/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceStatusSyncTask.java
develpoerX: Ok, Thanks for you suggestion.

develpoerX:
> Please create more specific issues for this PR.
> 
> e.g
> 
> 1. upgrade xxx from x to y. e.g.  https://issues.apache.org/jira/browse/SCB-2018
> 2. fix bug: xxxx
> 
> You can still use one PR to fix them all. These issues will used in Release Notes.

develpoerX:
I have modified it by putting the sub-problem description and linking in the description

PR #2013, opened 2020-10-26
[SCB-2099]Solve the problem of watch fail.

develpoerX:

---

### DIFF
--- a/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/WebsocketUtils.java
+++ b/service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/WebsocketUtils.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpMethod;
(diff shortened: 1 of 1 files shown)


service-registry/src/main/java/org/apache/servicecomb/serviceregistry/client/http/WebsocketUtils.java
```diff
@@ -52,18 +53,19 @@ public static void open(IpPort ipPort, String url, Handler<Void> onOpen, Handler
           onConnectFailed.handle(asyncResult.cause());
         } else {
           onOpen.handle(null);
-          asyncResult.result().exceptionHandler(v -> {
+          WebSocket webSocket = asyncResult.result();
```
liubao68, line 56: This PR seems nothing chaned. Can you describe when will fail of old code?

develpoerX: Yes, just some refactoring. I found that the process of handling this problem is simply refactored, so there is no synchronization process here.

liubao68: I close the issue. code from master is same.

PR #2154, opened 2020-12-21
Edge's 490 exception is thrown and can be used for business processing

liuguangrong:

---

### DIFF
--- a/transports/transport-rest/transport-rest-client/src/main/java/org/apache/servicecomb/transport/rest/client/http/RestClientInvocation.java
+++ b/transports/transport-rest/transport-rest-client/src/main/java/org/apache/servicecomb/transport/rest/client/http/RestClientInvocation.java
@@ -35,6 +35,7 @@ import org.apache.servicecomb.foundation.common.net.IpPort;
 import org.apache.servicecomb.foundation.common.net.URIEndpointObject;
 import org.apache.servicecomb.foundation.common.utils.ExceptionUtils;
 import org.apache.servicecomb.foundation.common.utils.JsonUtils;
+import org.apache.servicecomb.foundation.common.utils.SPIServiceUtils;
 import org.apache.servicecomb.foundation.vertx.client.http.HttpClientWithContext;
 import org.apache.servicecomb.foundation.vertx.http.HttpServletRequestEx;
 import org.apache.servicecomb.foundation.vertx.http.HttpServletResponseEx;
@@ -47,6 +48,7 @@ import org.apache.servicecomb.swagger.invocation.AsyncResponse;
 import org.apache.servicecomb.swagger.invocation.Response;
 import org.apache.servicecomb.swagger.invocation.exception.CommonExceptionData;
 import org.apache.servicecomb.swagger.invocation.exception.InvocationException;
+import org.apache.servicecomb.transport.rest.client.HttpRestClientInvocationHook;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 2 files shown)


transports/transport-rest/transport-rest-client/src/main/java/org/apache/servicecomb/transport/rest/client/http/RestClientInvocation.java
```diff
@@ -116,12 +118,17 @@ public void invoke(Invocation invocation, AsyncResponse asyncResp) throws Except
...
     }
+    List<HttpRestClientInvocationHook> httpRestClientInvocationHooks =
+            SPIServiceUtils.getAllService(HttpRestClientInvocationHook.class);
 
     clientRequest.exceptionHandler(e -> {
       invocation.getTraceIdLogger()
           .error(LOGGER, "Failed to send request, alreadyFailed:{}, local:{}, remote:{}, message={}.",
               alreadyFailed, getLocalAddress(), ipPort.getSocketAddress(),
               ExceptionUtils.getExceptionMessageWithoutTrace(e));
+      httpRestClientInvocationHooks.forEach(httpServerExceptionHandler -> {
```
liubao68, line 129: Maybe you can add a custom handler to fillful your requirements. 

e.g. 'MyHandler implements Handler'

servicecomb.handler.chain.Consumer.defaut:  your_current_handlers,myhandler

myhandler can process the exception thrown in RestClientInvocation

liuguangrong: https://docs.servicecomb.io/java-chassis/zh_CN/references-handlers/intruduction/

I found it didn't meet my requirements through experiments.

 My handler executes before the exception is returned, and does not return to handler after the exception is thrown.

I can't get this exception message.

Is there any other way

liubao68:
```
public class MyHandler implements Handler {

  private static final Logger LOGGER = LoggerFactory.getLogger(MyHandler.class);

  public static final String SPLITPARAM_RESPONSE_USER_SUFFIX = "(modified by MyHandler)";

  @Override
  public void handle(Invocation invocation, AsyncResponse asyncResp) throws Exception {
    LOGGER.info("before");

    invocation.next(response -> {
         LOGGER.info("after");
    });
  }
}
```

You can find many Handler implementations in servicecomb-java-chassis project.

Volcannozzz:
I think the handler example given by @liubao68  cannot meet @liuguangrong 's needs. Because the transportHandler is the last handler in the source code.

```java
public class ConsumerHandlerManager extends AbstractHandlerManager {
  @Override
  protected String getName() {
    return "Consumer";
  }

  @Override
  protected String getInnerDefaultChainDef() {
    return "simpleLB";
  }

  @Override
  protected Handler getLastHandler() {
    return TransportClientHandler.INSTANCE;
  }
}
```
and in `AbstractHandlerManager.java`

```java
  private List<Handler> createHandlerChain(String chainDef) {
    List<Class<Handler>> chainClasses = convertToChainClass(chainDef);

    List<Handler> handlerList = new ArrayList<>();
    for (Class<Handler> cls : chainClasses) {
      try {
        handlerList.add(cls.newInstance());
      } catch (Exception e) {
        // 在启动阶段直接抛异常出来
        throw new Error(e);
      }
    }
    handlerList.add(getLastHandler());
    return handlerList;
  }
```
In addition, when the transportHandler calls an exception, it will return the exception information to the caller. Can the exception information be handled by the caller? Hope you provide the calling way and error message stack.

liubao68:
This handler throw exception , will hanldled in code block `after` in my example code

Volcannozzz:
Sorry, I have  got the answer from line 86, 125, 544  in `RestClientInvocation.java`.

PR #2253, opened 2021-02-25
[SCB-2187]support dynamic key values as yaml source 

kaister3:


demo/demo-springmvc/springmvc-server/src/main/resources/microservice.yaml
```diff
@@ -47,10 +47,11 @@ servicecomb:
   # can download config center from https://cse-bucket.obs.myhwclouds.com/LocalCSE/Local-CSE-1.0.0.zip to test dynamic config
   config:
     client:
-  #    serverUri: http://127.0.0.1:30113
+      serverUri: http://127.0.0.1:30115
```
liubao68, line 50: This test is an auto-test in docker and no config center installed. So you can not submit this test code.  
After you finished your local test, you can add test cases without config center. For example, add yaml files to `microservice.yaml`

```
mytest:
  yaml: |
      hostName: wyk
```

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterConfig.java
```diff
@@ -17,13 +17,21 @@
 
 package org.apache.servicecomb.config.client;
 
-import java.util.List;
+import com.google.common.base.Joiner;
+import com.netflix.config.ConcurrentCompositeConfiguration;
+
+import java.util.*;
+import java.io.*;
+import java.net.*;
```
liubao68, line 25: Using `ctrl+alt+o` to organize imports. Do not use imortp x.x.*;

dynamic-config/config-cc/src/main/java/org/apache/servicecomb/config/client/ConfigCenterConfig.java
```diff
@@ -71,6 +81,40 @@ private ConfigCenterConfig() {
...
     finalConfig = config;
+    readFileSourceConfig();
+  }
+
+  @SuppressWarnings("unchecked")
+  public List<String> getFileSource() {
+    return (List<String>) finalConfig.getProperty(FILE_SOURCE);
+  }
+
+  private static void readFileSourceConfig() {
```
liubao68, line 92: This PR is not correct. We are not going to support config yaml files. But config yaml contents in config center. Using your test case for example. 

mytest.yaml is a key in config center, and the value of mytest.yaml is the file content.

kaister3: roger that

PR #2317, opened 2021-03-25
[SCB-2224] BeanUtils.getBeanOfType return null when aksk enabled

kaister3:


huawei-cloud/servicestage/src/main/java/org/apache/servicecomb/huaweicloud/servicestage/AKSKAuthHeaderProvider.java
```diff
@@ -142,8 +144,8 @@ private Cipher findCipher() {
       return DefaultCipher.getInstance();
     }
 
-    Map<String, Cipher> cipherBeans = BeanUtils.getBeansOfType(Cipher.class);
-    return cipherBeans.values().stream().filter(c -> c.name().equals(getCipher())).findFirst()
+    List<Cipher> ciphers = SPIServiceUtils.getOrLoadSortedService(Cipher.class);
```
liubao68, line 147: Missed SPI definition file for AKSKCIpher?

kaister3: SPI definition file is added

PR #2352, opened 2021-04-17
update vertx and netty

kaister3:
vertx 3.9.7
netty 4.1.62.Final


kaister3:
[https://github.com/eclipse-vertx/vert.x/issues/3899](url)

Issue #2381, opened 2021-05-13
"message":"Unexpected consumer error, please check logs for details"


hypggg:
调用场景：路由服务转发到目标服务。
报错情况：重新发布目标服务后，就报"message":"Unexpected consumer error, please check logs for details"，需要重启路由服务才能好使。
错误信息如下：
2021-05-13 12:15:58,604Z(65581731):DEBUG{}[vert.x-eventloop-thread-1#15]-->Using retry handler org.apache.servicecomb.loadbalance.DefaultRetryExtensionsFactory$1 for microservice cmconsoleservice.
2021-05-13 12:15:58,604Z(65581731):INFO{}[vert.x-eventloop-thread-1#15]-->Found SPI service org.apache.servicecomb.router.custom.RouterHeaderFilterExt, count=0.
2021-05-13 12:15:58,605Z(65581732):DEBUG{}[vert.x-eventloop-thread-1#15]-->route management match rule success: PolicyRuleItem{precedence=1, match=null, route=[RouteItem{weight=100, currentWeight=0, tags={version=2021.05.05.2126}, tagitem=org.apache.servicecomb.router.model.TagItem@2bc555f9}], total=null, weightLess=false}
2021-05-13 12:15:58,605Z(65581732):DEBUG{}[vert.x-eventloop-thread-1#15]-->route management can not match any rule and route the latest version
2021-05-13 12:15:58,605Z(65581732):DEBUG{}[vert.x-eventloop-thread-1#15]-->route management distribute rule success: []
2021-05-13 12:15:58,605Z(65581732):WARN{}[vert.x-eventloop-thread-1#15]-->There are not servers exist after filtered by class org.apache.servicecomb.router.custom.RouterServerListFilter.
2021-05-13 12:15:58,605Z(65581732):ERROR{SERVICECOMB_TRACE_ID=609d187e56f0bbb1-16348}[vert.x-eventloop-thread-1#15]-->Invoke all server failed. Operation CONSUMER  xxxService.xxxController.xxxMethod, e=cause:ClientException,message:Load balancer does not have available server for client: default

liubao68:
AbstractRouterDistributor

```
  public void initLatestVersion(String serviceName, List<T> list) {
    if (RouterRuleCache.getServiceInfoCacheMap().get(serviceName).getLatestVersionTag() != null) {
      return;
    }
   ... ....
    TagItem tagitem = new TagItem(latestVersion);
[...]

PR #2384, opened 2021-05-14
[#2381]fix initLatestVersion method

kaister3:


handlers/handler-router/src/main/java/org/apache/servicecomb/router/distribute/AbstractRouterDistributor.java
```diff
@@ -140,9 +145,17 @@ public TagItem getFiltedServerTagItem(PolicyRuleItem rule, String targetServiceN
...
+
+    if (latestVersionTag != null) {
+      for (T server : list) {
+        E ms = getIns.apply(server);
+        if (getVersion.apply(ms).equals(latestVersionTag.getVersion())) {
+          return;
+        }
+      }
     }
+
```
liubao68, line 158: Maybe it's better to using API's in foundation-registry to find latest version. 
While now, line 148~158 can be deleted, since if you check version match, will call getIns always, it it some as delete all these lines.

liubao68 (approved):

PR #2387, opened 2021-05-18
[SCB-2267]add registryName api

kaister3:


service-registry/registry-service-center/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -253,6 +253,9 @@ public static void executeOnEachServiceRegistry(Consumer<ServiceRegistry> action
 
   public static void addExtraServiceRegistry(ServiceRegistry serviceRegistry) {
     Objects.requireNonNull(serviceRegistry);
+    if (EXTRA_SERVICE_REGISTRIES.containsKey(serviceRegistry.getName())) {
```
liubao68, line 256: and name can not be the default regitry name

service-registry/registry-service-center/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -253,6 +253,9 @@ public static void executeOnEachServiceRegistry(Consumer<ServiceRegistry> action
...
+      throw new IllegalArgumentException("Register name duplicated!");
```
liubao68, line 257: reigstriy name {name} is duplicated, please set different names for each implementation.

kaister3: when duplicated, I can only get the new client name, cannot get old client name from EXTRA_SERVICE_REGISTRY. How to set different implmentation?

service-registry/registry-service-center/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
kaister3: getName() here returns custom registry name, not default name

liubao68 (approved):

PR #2390, opened 2021-05-20
Set limitProperty default value

fanjiang-2021:
When qpsStrategy is deleted from the configuration center, the service will report a null pointer exception, so set a default value


liubao68:
Can you describe the problem for this change ? Seems this code is not neccessary.

PR #2538, opened 2021-08-31
A request without any Accept header field implies that the user agent…

fanjiang-2021:
A request without any Accept header field implies that the user agent will accept any media type in response。

when consumer sends a request to provider without without any Accept header field, serviceomb-java-chassis SDK will set Accept header empty.

According to RFC regulation, A request without any Accept header field implies that the user agent will accept any media type in response,
so if a consumer sends a request to provider without without any Accept header field,  it is more reasonable that serviceomb-java-chassis SDK not set Accept header.


liubao68:
Thanks for this PR. But I don't understand `when consumer sends a request to provider without without any Accept header field, serviceomb-java-chassis SDK will set Accept header empty.`  and this PR. I can't find where java chassis set Accept header when consumer sends a request. 

Can you provide detail information about this?

PR #2545, opened 2021-09-03
[SCB-2331] swagger generator support NotBlank and NotEmpty annotations

fanjiang-2021:
Current swagger version is 1.6.2, this version do not recognize  NotBlank and NotEmpty annotations. In some cases, it is misunderstand that ome fields Using  NotBlank and NotEmpty annotations are not equired.


liubao68:
retrigger ci

PR #2578, opened 2021-09-14
[SCB-2337] Add timing verification for MicroserviceVersions

maxucheng0:
More info about see issue #2574

### DIFF
--- a/foundations/foundation-registry/src/main/java/org/apache/servicecomb/registry/consumer/MicroserviceManager.java
+++ b/foundations/foundation-registry/src/main/java/org/apache/servicecomb/registry/consumer/MicroserviceManager.java
@@ -60,8 +60,8 @@ public class MicroserviceManager {
         microserviceVersions = versionsByName.get(microserviceName);
         if (microserviceVersions == null) {
           microserviceVersions = new MicroserviceVersions(appManager, appId, microserviceName);
-          microserviceVersions.pullInstances();
           versionsByName.put(microserviceName, microserviceVersions);
+          microserviceVersions.pullInstances();
         }
       }
     }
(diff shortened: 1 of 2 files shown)


foundations/foundation-registry/src/main/java/org/apache/servicecomb/registry/consumer/MicroserviceManager.java
```diff
@@ -60,8 +60,8 @@ public MicroserviceVersions getOrCreateMicroserviceVersions(String microserviceN
         microserviceVersions = versionsByName.get(microserviceName);
         if (microserviceVersions == null) {
           microserviceVersions = new MicroserviceVersions(appManager, appId, microserviceName);
-          microserviceVersions.pullInstances();
           versionsByName.put(microserviceName, microserviceVersions);
+          microserviceVersions.pullInstances();
```
liubao68, line 64: Maybe you can add a synchronized to `pullInstances`  and `onMicroserviceInstanceChanged` can solve this problem. 
And your modification seems not correct and may cause some other concurent problems and add add a cyclic dependency between foundation-registry and registry-service-center.

```
  public void pullInstances() {
    for (MicroserviceVersions microserviceVersions : versionsByName.values()) {
      microserviceVersions.pullInstances();
      tryRemoveInvalidMicroservice(microserviceVersions);
    }
  }
  public void onMicroserviceInstanceChanged(MicroserviceInstanceChangedEvent changedEvent) {
    for (MicroserviceVersions microserviceVersions : versionsByName.values()) {
      microserviceVersions.onMicroserviceInstanceChanged(changedEvent);
      tryRemoveInvalidMicroservice(microserviceVersions);
    }
  }
```

maxucheng0: 好的，稍后我修改并验证下

maxucheng0:
> BTW, I see you add an JIRA issue [SCB-2337], but I can't find this in JIRA https://issues.apache.org/jira/projects/SCB/versions/12350513
> 
> Do you have an JIRA account or have mistake in the issue ID?
> I can add you to JIRA group if you do not have one, see https://servicecomb.apache.org/cn/developers/use-jira/

已创建

liubao68:
fixed in https://github.com/apache/servicecomb-java-chassis/pull/2593

PR #2628, opened 2021-10-30
Use final to clear the test property

Shoothzj:
Use final to clear the test property

### DIFF
--- a/transports/transport-rest/transport-rest-vertx/src/test/java/org/apache/servicecomb/transport/rest/vertx/TestTransportConfig.java
+++ b/transports/transport-rest/transport-rest-vertx/src/test/java/org/apache/servicecomb/transport/rest/vertx/TestTransportConfig.java
@@ -46,8 +46,12 @@ public class TestTransportConfig {
 
   @Test
   public void testGetAddressNormal() {
-    ArchaiusUtils.setProperty("servicecomb.rest.address", "1.1.1.1");
-    Assert.assertEquals("1.1.1.1", TransportConfig.getAddress());
+    try {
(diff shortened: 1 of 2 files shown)


transports/transport-rest/transport-rest-vertx/src/test/java/org/apache/servicecomb/transport/rest/vertx/TestTransportConfig.java
```diff
@@ -63,174 +67,252 @@ int availableProcessors() {
...
   @Test
   public void testGetThreadCountNormal() {
-    ArchaiusUtils.setProperty("servicecomb.rest.server.thread-count", 10);
-    Assert.assertEquals(10, TransportConfig.getThreadCount());
+    try {
+      ArchaiusUtils.setProperty("servicecomb.rest.server.thread-count", 10);
+      Assert.assertEquals(10, TransportConfig.getThreadCount());
+    } finally {
+      ArchaiusUtils.setProperty("servicecomb.rest.server.thread-count", null);
+    }
```
liubao68, line 75: @After
  public void after() {
    ArchaiusUtils.resetConfig();
  }


May have done the job, do you find any problem this do not have this change?

Shoothzj: I don't know this method. You mean all our tests don't rely on `config`. They can be cleared?

liubao68: yes.

PR #2704, opened 2022-01-20
[SCB-2377] expose update microservice properties API

fanjiwang1992:
update microservice properties


service-registry/registry-service-center/src/main/java/org/apache/servicecomb/serviceregistry/RegistryUtils.java
```diff
@@ -160,6 +160,16 @@ public static MicroserviceInstance getMicroserviceInstance() {
         null : serviceCache.getInstances();
   }
 
+  // update microservice  properties
+  public static boolean updateMicroserviceProperties(Map<String, String> microserviceProperties) {
```
liubao68, line 164: RegistryUtils maybe changed in future, do not use API as possible.

fanjiwang1992: ok

PR #2722, opened 2022-02-18
[SCB-2387] optimize abnormal print information of SwaggerProducerOper…

david6969xin:
…ation

---

### DIFF
--- a/swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/engine/SwaggerProducerOperation.java
+++ b/swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/engine/SwaggerProducerOperation.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 
-import org.apache.servicecomb.foundation.common.utils.ExceptionUtils;
 import org.apache.servicecomb.foundation.common.utils.SPIServiceUtils;
 import org.apache.servicecomb.swagger.invocation.AsyncResponse;
 import org.apache.servicecomb.swagger.invocation.Response;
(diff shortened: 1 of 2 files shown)


swagger/swagger-invocation/invocation-core/src/main/java/org/apache/servicecomb/swagger/engine/SwaggerProducerOperation.java
```diff
@@ -153,7 +152,7 @@ public void doCompletableFutureInvoke(SwaggerInvocation invocation, AsyncRespons
     } catch (Throwable e) {
       if (shouldPrintErrorLog(e)) {
         LOGGER.error("unexpected error operation={}, message={}",
-            invocation.getInvocationQualifiedName(), ExceptionUtils.getExceptionMessageWithoutTrace(e));
+            invocation.getInvocationQualifiedName(), e.getMessage());
```
liubao68, line 155: Why modify this ?  There is no SCB-2387 for this.

david6969xin: [This code will print too much sensitive information when JSON parsing fails](https://github.com/apache/servicecomb-java-chassis/issues/2679)

yhs0092: Not all of the sensitive contents can be blocked by this way. For example, the Jackson may put raw JSON string into exception message.

liubao68: Personally I'd prefer not modify this. This is not an actual serious information leak problem. And if the user's do need fix it, they can add try-catch block to avoid throw JsonParseException. 

This information is quite useful for problem analysis and have modified many times.

liubao68:
不处理

PR #2743, opened 2022-03-23
[SCB-2426] upgrade springboot version to 2.6.6

Shoothzj:
upgrade springboot version to 2.6.2
springboot version 2.6.2 depends on spring version 5.3.14, also updates


liubao68:
We may use springn boot 2.6.x when 2.7 is ready. Now it is not a good idea to upgrade.

liubao68 (approved):

PR #2745, opened 2022-03-23
[SCB-2428] add jakarta dependency in foundation-config

Shoothzj:
java later version has removed `javax.annotation` dependency. If we need to support multi java versions, we need to add `jakarta` dependecy

### DIFF
--- a/foundations/foundation-config/pom.xml
+++ b/foundations/foundation-config/pom.xml
@@ -52,6 +52,10 @@
       <artifactId>snakeyaml</artifactId>
     </dependency>
     <dependency>
+      <groupId>jakarta.annotation</groupId>
+      <artifactId>jakarta.annotation-api</artifactId>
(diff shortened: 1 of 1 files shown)


foundations/foundation-config/pom.xml
```diff
@@ -51,6 +51,10 @@
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
     </dependency>
+    <dependency>
+      <groupId>jakarta.annotation</groupId>
+      <artifactId>jakarta.annotation-api</artifactId>
+    </dependency>
```
liubao68, line 57: Why add this dependency? Can you provide the problem?

Shoothzj:
@liubao68 when compile using java 17
image

liubao68 (approved):

PR #2761, opened 2022-03-30
[SCB-2434] log number not correct

Shoothzj:


liubao68:
releasing is in progress, will check it after released.

liubao68 (approved):

PR #2762, opened 2022-03-30
[SCB-2435] protected field/method in final class

Shoothzj:


liubao68:
releasing is in progress, will check it after released.

liubao68 (approved):

PR #2789, opened 2022-04-19
[SCB-2456] introduce maven spotbugs plugin

Shoothzj:


pom.xml
```diff
@@ -295,6 +296,14 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.github.spotbugs</groupId>
+        <artifactId>spotbugs-maven-plugin</artifactId>
+        <configuration>
+          <xmlOutput>true</xmlOutput>
+          <failOnError>false</failOnError>
```
liubao68, line 304: It's better enabled. We can add exculde rules to avoid changing too much in the first place.

Shoothzj: @liubao68 you are right, I will fix it this weekend.

governance/src/main/java/org/apache/servicecomb/governance/policy/CircuitBreakerPolicy.java
```diff
@@ -136,9 +136,6 @@ public SlidingWindowType getSlidingWindowTypeEnum() {
     if (StringUtils.isEmpty(slidingWindowType)) {
       return SlidingWindowType.TIME_BASED;
     }
-    if (SlidingWindowType.COUNT_BASED.equals(slidingWindowType)) {
-      return SlidingWindowType.COUNT_BASED;
-    }
     return SlidingWindowType.TIME_BASED;
```
liubao68, line 139: Why  modify this? seems not correct.

liubao68 (approved):

PR #2799, opened 2022-04-25
[#2795]Fixed the issue that the dynamic modification configuration of gray release is invalid. 

m13922236379:
The new constructor makes the original constructor invalid.Combine the two constructors so that the class is loaded correctly.

### DIFF
--- a/governance/src/main/java/org/apache/servicecomb/router/cache/RouterRuleCache.java
+++ b/governance/src/main/java/org/apache/servicecomb/router/cache/RouterRuleCache.java
@@ -47,7 +47,7 @@ public class RouterRuleCache {
 
   private static final String ROUTE_RULE = "servicecomb.routeRule.%s";
 
-  private Environment environment;
+  private final Environment environment;
 
   private final ConcurrentHashMap<String, ServiceInfoCache> serviceInfoCacheMap = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


governance/src/main/java/org/apache/servicecomb/router/cache/RouterRuleCache.java
```diff
@@ -49,20 +49,17 @@ public class RouterRuleCache {
...
   private final Object lock = new Object();
 
   private final Representer representer = new Representer();
 
-  public RouterRuleCache() {
+  @Autowired
+  public RouterRuleCache(Environment environment) {
     representer.getPropertyUtils().setSkipMissingProperties(true);
     GovernanceEventManager.register(this);
+    this.environment = environment;
```
Shoothzj, line 62: maybe we should put this line to the first line of consturctor

governance/src/main/java/org/apache/servicecomb/router/cache/RouterRuleCache.java
```diff
@@ -49,20 +49,17 @@ public class RouterRuleCache {
...
-
   private final ConcurrentHashMap<String, ServiceInfoCache> serviceInfoCacheMap = new ConcurrentHashMap<>();
 
   private final Object lock = new Object();
 
   private final Representer representer = new Representer();
 
-  public RouterRuleCache() {
+  @Autowired
+  public RouterRuleCache(Environment environment) {
```
Shoothzj, line 59: can we mark environment final?

liubao68 (approved):

PR #2806, opened 2022-04-27
[SCB-2469] update vertx to 4.2.7

hezhangjian:
## main changes
- cookie changes from map to set, because the rfc allow more than two cookies with same name
- `getMaxHeaderListSize` changes from Integer.MAX to 8192


transports/transport-rest/transport-rest-vertx/src/test/java/org/apache/servicecomb/transport/rest/vertx/TestTransportConfig.java
```diff
@@ -215,7 +215,7 @@ public void testHttp2Setting() {
     ArchaiusUtils.setProperty("servicecomb.rest.server.http2.maxFrameSize", 65535);
     Assert.assertEquals(65535, TransportConfig.getMaxFrameSize());
 
-    Assert.assertEquals(Integer.MAX_VALUE, TransportConfig.getMaxHeaderListSize());
+    Assert.assertEquals(8192, TransportConfig.getMaxHeaderListSize());
```
liubao68, line 218: This change need to add a subtask to update release notes.

Issue #2809, opened 2022-04-28
During cross-application invoking, the routeRule in gray release cannot match the instance.


m13922236379:
Only the service name is used for matching. When the gray release rule is Appname combined with Servicename, the matched instance is 0.

PR #2810, opened 2022-04-28
[#2809]A judgment condition is added when gray rules match instances. This condition applies to scenarios where instances cannot be matched in cross-app invoking.

m13922236379:
A judgment condition is added when gray rules match instances. This condition applies to scenarios where instances cannot be matched in cross-app invoking.
From:#2809

### DIFF
--- a/governance/src/main/java/org/apache/servicecomb/router/distribute/AbstractRouterDistributor.java
+++ b/governance/src/main/java/org/apache/servicecomb/router/distribute/AbstractRouterDistributor.java
@@ -43,6 +43,8 @@ public abstract class AbstractRouterDistributor<T, E> implements
   private Function<E, String> getVersion;
 
   private Function<E, String> getServerName;
+  
+  private Function<E, String> getAlias;
 
   private Function<E, Map<String, String>> getProperties;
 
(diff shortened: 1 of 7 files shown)


governance/src/main/java/org/apache/servicecomb/router/distribute/AbstractRouterDistributor.java
```diff
@@ -146,7 +150,7 @@ public void initLatestVersion(String serviceName, List<T> list) {
     String latestVersion = null;
     for (T server : list) {
       E ms = getIns.apply(server);
-      if (getServerName.apply(ms).equals(serviceName)) {
+      if (getServerName.apply(ms).equals(serviceName) || getAlias.apply(ms).equals(serviceName)) {
         if (latestVersion == null || VersionCompareUtil
```
zhaozw, line 154: getAlias如何可以为null，加上getAlias!=null&&getAlias.apply(ms).equals(serviceName) ?

m13922236379: 同理，那getServerName应该也不能为null吧。

liubao68:
I'd prefer to keep  targetServiceName parameter

```
 public <T, E> List<T> getFilteredListOfServers(List<T> list,
                                                 String targetServiceName, Map<String, String> headers, RouterDistributor<T, E> distributer) {
```

in RouterFilter will always be microservice name (do not contains application name).

m13922236379:
Fix the problem by fixing targetServiceName to the microservice name.

PR #2831, opened 2022-05-08
[SCB-2475] migrate to junit5 registry part 1

Shoothzj:


service-registry/registry-service-center/src/test/java/org/apache/servicecomb/serviceregistry/api/registry/TestMicroServiceInstance.java
```diff
@@ -30,9 +30,9 @@
 import org.apache.servicecomb.serviceregistry.RegistryUtils;
 import org.junit.After;
 import org.junit.AfterClass;
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
```
david6969xin, line 34: @Test、@Before、@After and @AfterClass ····· should migrate to junit5 also

Shoothzj: @david6969xin It's just a part of work, just migrate `Assertions`. Some testcases use `jmockit`, they don't work well with junit5

liubao68 (approved):

PR #2868, opened 2022-05-11
[SCB-2500]remove unused wiremock-standalone maven dependency

tyluffy:


Shoothzj:
@tyluffy could you please test again enable `it` profile?

liubao68:
Seems tracing will not always produce a predictable result. I try to retrigger CI.

PR #2945, opened 2022-05-18
[SCB-2475] migrate to junit5 integration-tests part1

Shoothzj:


integration-tests/it-consumer/src/main/java/org/apache/servicecomb/it/testcase/TestAnnotatedAttribute.java
```diff
@@ -16,15 +16,13 @@
  */
 package org.apache.servicecomb.it.testcase;
 
-import static org.junit.Assert.assertEquals;
```
david6969xin, line 19: maybe you can use org.junit.jupiter.api.Assertions.assertEquals to repalce org.junit.Assert.assertEquals

Shoothzj: @david6969xin Yes, It's a way to migrate to `Assertions`. I didn't choose that way because
`static import is kind of ugly`. And `static org.junit.Assert.assertEquals` is not widely use in this module, even the whole project. It's not a big change.

liubao68 (approved):

PR #2979, opened 2022-05-20
[SCB-2522] Optimize dependency management jersey

Shoothzj:


dependencies/default/pom.xml
```diff
@@ -632,12 +632,6 @@
         <scope>test</scope>
       </dependency>
 
-      <dependency>
-        <groupId>org.glassfish.jersey.core</groupId>
-        <artifactId>jersey-common</artifactId>
-        <version>${jersey.version}</version>
```
liubao68, line 638: jersey.version  properties can be delete too.

Shoothzj:
@liubao68 PTAL again, thanks

dependencies/default/pom.xml
Shoothzj: @liubao68 I didn't mean to delete this properties. We used jersey, It's a miss delete.

PR #3009, opened 2022-05-27
[SCB-2533] [1.3.x] encode invocation context before add to http header

demonbug:
fix issus #3008

add writeUnicodeValueAsString to JsonUtils to support unicode encoding
encode invocation context before add to http header

### DIFF
--- a/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/JsonUtilsTest.java
+++ b/foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/JsonUtilsTest.java
@@ -1,0 +1,29 @@
+package org.apache.servicecomb.foundation.common.utils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import junit.framework.TestCase;
+import org.apache.commons.lang.StringUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class JsonUtilsTest extends TestCase {
+
+    @Test
+    public void testWriteUnicodeValueAsString() throws JsonProcessingException {
+        String unicodeStr = "测试";
+        Map<String, String> inMap = new HashMap<>();
+        inMap.put("key", unicodeStr);
+
+        Assert.assertFalse(StringUtils.isAsciiPrintable(JsonUtils.writeValueAsString(inMap)));
+        String jsonStr = JsonUtils.writeUnicodeValueAsString(inMap);
+        Assert.assertTrue(StringUtils.isAsciiPrintable(jsonStr));
+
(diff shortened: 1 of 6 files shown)


transports/transport-rest/transport-rest-client/src/main/java/org/apache/servicecomb/transport/rest/client/http/RestClientInvocation.java
```diff
@@ -274,7 +274,7 @@ protected void fail(ConnectionBase connection, Throwable e) {
 
   protected void setCseContext() {
     try {
-      String cseContext = JsonUtils.writeValueAsString(invocation.getContext());
+      String cseContext = JsonUtils.writeUnicodeValueAsString(invocation.getContext());
```
liubao68, line 277: Is this compatible with old versions? e.g. new consumer invoke an older version provider.

demonbug: I think it is compatible, Jackson will automatic decode unicode string in deserialization.
In testcase, use writeUnicodeValueAsString to serialize and use the old OBJ_MAPPER to deserialize, the value is not changed

liubao68: Can you add some unit test case for this?

foundations/foundation-common/src/test/java/org/apache/servicecomb/foundation/common/utils/JsonUtilsTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import junit.framework.TestCase;
+import org.apache.commons.lang.StringUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
```
liubao68, line 28: Use junit5 for test cases, and I'd prefer PR for master.

demonbug: the lastest NUWA is still using branch 1.3.x, and this branch has not import junit5

transports/transport-rest/transport-rest-client/src/main/java/org/apache/servicecomb/transport/rest/client/http/RestClientInvocation.java
demonbug: ok, i added 2 testcase in TestRestClientInvocation.java

Shoothzj (changes requested):
Apache commons-lang3 also have `StringEscapeUtils`, why we need `commons-text`?

demonbug:
> Apache commons-lang3 also have `StringEscapeUtils`, why we need `commons-text`?
StringEscapeUtils was deprecated in commons-lang3

Shoothzj (approved):

liubao68 (approved):

PR #3035, opened 2022-06-05
[SCB-2544]Use protobuf-maven-plugin generat java class in foundation-protobuf

fu-turer:
### Motivation
Remove protoc generated java sources file, and use protobuf-maven-plugin generate when complie


foundations/foundation-protobuf/pom.xml
```diff
@@ -67,4 +67,34 @@
...
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>test-compile</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
 </project>
```
Shoothzj, line 100: better to add new line on the end

Shoothzj (approved):

PR #3045, opened 2022-06-08
[SCB-2533] [master] encode invocation context before add to http header

demonbug:
fix issus https://github.com/apache/servicecomb-java-chassis/issues/3008

add writeUnicodeValueAsString to JsonUtils to support unicode encoding
encode invocation context before add to http header


dependencies/default/pom.xml
```diff
@@ -44,6 +44,7 @@
     <commons-io.version>2.11.0</commons-io.version>
     <commons-lang.version>2.6</commons-lang.version>
     <commons-lang3.version>3.12.0</commons-lang3.version>
+    <commons-text.version>1.9</commons-text.version>
```
Shoothzj, line 47: we sort dependencies by alphabeta, it should below `commons-logging`

Shoothzj:
LGTM, one nit.

Shoothzj (approved):

dependencies/default/pom.xml
demonbug: fixed

Shoothzj:
Thanks for your contribution. And welcom to servicecomb community.

Issue #3066, opened 2022-06-13
The application cannot start normally in the jdk 17 environment


david6969xin:
Caused by: java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlRootElement
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:323) ~[swagger-core-1.6.6.jar:1.6.6]
	at org.apache.servicecomb.swagger.extend.ModelResolverExt.resolve(ModelResolverExt.java:160) ~[classes/:na]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:205) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:103) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:289) ~[swagger-core-1.6.6.jar:1.6.6]
	at org.apache.servicecomb.swagger.extend.ModelResolverExt.resolve(ModelResolverExt.java:160) ~[classes/:na]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:205) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:103) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:313) ~[swagger-core-1.6.6.jar:1.6.6]
	at org.apache.servicecomb.swagger.extend.ModelResolverExt.resolve(ModelResolverExt.java:160) ~[classes/:na]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:205) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:103) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.converter.ModelConverters.readAll(ModelConverters.java:98) ~[swagger-core-1.6.6.jar:1.6.6]
	at io.swagger.converter.ModelConverters.readAll(ModelConverters.java:88) ~[swagger-core-1.6.6.jar:1.6.6]
[...]

PR #3067, opened 2022-06-13
[SCB-2565] ClassNotFoundException: javax.xml.bind.annotation:XmlRootE…

david6969xin:
fix [#3066](https://github.com/apache/servicecomb-java-chassis/issues/3066)

### DIFF
--- a/swagger/swagger-generator/generator-core/pom.xml
+++ b/swagger/swagger-generator/generator-core/pom.xml
@@ -45,6 +45,11 @@
     </dependency>
 
     <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+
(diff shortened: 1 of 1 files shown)


swagger/swagger-generator/generator-core/pom.xml
```diff
@@ -72,7 +72,6 @@
     <dependency>
       <groupId>javax.xml.bind</groupId>
       <artifactId>jaxb-api</artifactId>
-      <scope>test</scope>
```
Shoothzj, line 75: Yes, that's the way to fix it. But should we put it below these test dependencies?

david6969xin: yes, It is better to adjust the order of dependencies

Shoothzj (approved):

PR #3068, opened 2022-06-13
[SCB-2566] Remove unused tcnative dependency version define

lovehzj:
It's already defined in netty-bom, and should be keep consistent with netty.


Shoothzj (approved):
LGTM, miss the jira number

liubao68 (approved):

PR #3104, opened 2022-06-21
[SCB-2589] fix spans is null sometimes

david6969xin:

---

### DIFF
--- a/integration-tests/test-common/src/test/java/org/apache/servicecomb/tests/tracing/TracingTestBase.java
+++ b/integration-tests/test-common/src/test/java/org/apache/servicecomb/tests/tracing/TracingTestBase.java
@@ -62,7 +62,11 @@ public class TracingTestBase {
         .map(this::extractIds)
         .collect(Collectors.toList());
 
-    List<Span> spans = zipkin.getTrace(traceId(loggedIds));
+    List<Span> spans;
+    do {
+      spans = zipkin.getTrace(traceId(loggedIds));
+    } while (spans == null);
+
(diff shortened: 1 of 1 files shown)


integration-tests/test-common/src/test/java/org/apache/servicecomb/tests/tracing/TracingTestBase.java
```diff
@@ -62,7 +62,11 @@ protected void assertThatSpansReceivedByZipkin(Collection<String> logs, String..
         .map(this::extractIds)
         .collect(Collectors.toList());
 
-    List<Span> spans = zipkin.getTrace(traceId(loggedIds));
+    List<Span> spans;
+    do {
+      spans = zipkin.getTrace(traceId(loggedIds));
```
Shoothzj, line 67: we need to sleep in case of cpu 100%, also we need a reasonable timeout.
Howerver, instead of write our impl, we can use https://github.com/awaitility/awaitility library
```
Awaitility.await().atMost(10, SECONDS).until(() -> xxxxxx);
```

Shoothzj (changes requested):

integration-tests/test-common/src/test/java/org/apache/servicecomb/tests/tracing/TracingTestBase.java
david6969xin: just testing code, that almost never happens, don't need to think so much.   of course, it's better to be strict !

Shoothzj (approved):

PR #3124, opened 2022-06-24
[SCB-2603]Remove dependency powermock

fu-turer:


Shoothzj (approved):
I think we will use `mockito-inline` to mock static method or constructor eventually.

PR #3184, opened 2022-07-10
[SCB-2634]migrate common-access-log module to mockito

fu-turer:


common/common-access-log/pom.xml
```diff
@@ -47,8 +47,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.jmockit</groupId>
-      <artifactId>jmockit</artifactId>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
```
Shoothzj, line 51: we already have `mockito-core` in project level dependency

Shoothzj (approved):

PR #3223, opened 2022-07-26
[SCB-2643]fix instance isolation filter not work problem

zhaozw:

---


handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/ServerListFilterExt.java
```diff
@@ -29,7 +29,7 @@
  */
 public interface ServerListFilterExt {
   default int getOrder() {
-    return 0;
+    return FilterConstant.ORDER_NORMAL;
```
liubao68, line 32: Do not change default value. ORDER_NORMAL should be 0.

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/FilterConstant.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.servicecomb.loadbalance;
+
+public interface FilterConstant {
+  int ORDER_NORMAL = 100;
```
liubao68, line 21: 0

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/FilterConstant.java
```diff
@@ -0,0 +1,32 @@
...
+
+  int ORDER_ISOLATION = 0;
```
liubao68, line 23: -100

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/FilterConstant.java
```diff
@@ -0,0 +1,32 @@
...
+
+  int ORDER_ZONE_AWARE = 200;
+
+  String EMPTY_INSTANCE_PROTECTION = "servicecomb.loadbalance.filter.isolation.emptyInstanceProtectionEnabled";
```
liubao68, line 27: Maybe can define these constants in ServerListFilterExt

handlers/handler-loadbalance/src/main/java/org/apache/servicecomb/loadbalance/filterext/IsolationDiscoveryFilter.java
```diff
@@ -42,17 +39,15 @@
...
+public class IsolationDiscoveryFilter implements ServerListFilterExt {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(IsolationDiscoveryFilter.class);
 
-  private static final String EMPTY_INSTANCE_PROTECTION = "servicecomb.loadbalance.filter.isolation.emptyInstanceProtectionEnabled";
-
   private final DynamicBooleanProperty emptyProtection = DynamicPropertyFactory.getInstance()
       .getBooleanProperty(EMPTY_INSTANCE_PROTECTION, false);
 
+  public EventBus eventBus = EventManager.getEventBus();
```
liubao68, line 49: use private

Shoothzj: can we mark this field final?

liubao68 (approved):

etteeee:
Now we have changed isolation and zoneaware type from `DiscoveryFilters` to `ServerListFilterEx`, so what is the real difference between this 2 types filter except for the filtering timing? What is the **RULE** when adding a new filter policy? @liubao68 @zhaozw 

btw, it's better to create a new class with new name rather than moving to a new package, the class name may cause unnecessary confusion.

liubao68:
@etteeee DiscoveryFilters may have better performance but can only be used when filter results not change only when instance lists changed, while ServerListFilterEx can be used for general purpose.

PR #3453, opened 2022-11-04
[SCB-2717] add exception retry for SslHandshakeTimeoutException

yanghao605:

---

### DIFF
--- a/governance/src/main/java/org/apache/servicecomb/governance/handler/ext/FailurePredictor.java
+++ b/governance/src/main/java/org/apache/servicecomb/governance/handler/ext/FailurePredictor.java
@@ -43,8 +43,8 @@ public interface FailurePredictor {
           .put(IOException.class, Collections.singletonList("Connection reset by peer"))
           .put(VertxException.class, Collections.singletonList("Connection was closed"))
           .put(NoRouteToHostException.class, Collections.emptyList())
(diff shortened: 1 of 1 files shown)


liubao68:
Issue is not replied. Seems retry for timeout exception and handshake exception will not make things better if retried on same server. But if retry on other instances is a bit useful.

liubao68:
Seems retry for timeout exception and handshake exception have more problems than benefit, how do you think so ? @yanghao605

yanghao605:
ok,by default  I will send the retry request to another node

yanghao605:
retry for handshaketimeout exception can improve the reliability and reduce the probability of request loss
Maybe, we can apply this issue, then emphasize the retry strategy that user used should be retried on other instances

liubao68:
Can you provide more information about why SSL handshake or timeout?  

Most of time retry is not a good idea, event  retry on another instance. For example, ssl handshake fail for unknow certificate. 
Or ssl handshake because two many concurrent SSL handshakes.

yanghao605:
fullgc耗时过长导致SSL握手超时异常

yanghao605:
SSL handshake timeout exception caused by Long lasting full gc

liubao68 (approved):

PR #3877, opened 2023-07-25
[#3865] show interface parameter info when throw InvocationException

chengyouling:


common/common-rest/src/main/java/org/apache/servicecomb/common/rest/codec/RestCodec.java
```diff
@@ -57,16 +60,19 @@ public static Map<String, Object> restToArgs(HttpServletRequest request,
     for (RestParam param : paramList) {
       try {
         paramValues.put(param.getParamName(), param.getParamProcessor().getValue(request));
-      } catch (InvocationException e) {
-        throw e;
       } catch (Exception e) {
         // Avoid information leak of user input, and add option for debug use.
```
liubao68, line 64: remove this notes

liubao68 (approved):

PR #4057, opened 2023-11-26
[#4035] fix invocation context loss issue

yanghao605:

---

### DIFF
--- a/handlers/handler-tracing-zipkin/src/main/java/org/apache/servicecomb/tracing/zipkin/ZipkinConsumerDelegate.java
+++ b/handlers/handler-tracing-zipkin/src/main/java/org/apache/servicecomb/tracing/zipkin/ZipkinConsumerDelegate.java
@@ -50,7 +50,7 @@ class ZipkinConsumerDelegate implements ZipkinTracingDelegate {
   }
 
   @Override
-  public Span createSpan(Invocation invocation) {
+  public synchronized Span createSpan(Invocation invocation) {
     return handler.handleSend(requestWrapper.invocation(invocation));
   }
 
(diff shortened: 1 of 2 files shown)


handlers/handler-tracing-zipkin/src/main/java/org/apache/servicecomb/tracing/zipkin/ZipkinConsumerDelegate.java
```diff
@@ -50,7 +50,7 @@ class ZipkinConsumerDelegate implements ZipkinTracingDelegate {
   }
 
   @Override
-  public Span createSpan(Invocation invocation) {
+  public synchronized Span createSpan(Invocation invocation) {
     return handler.handleSend(requestWrapper.invocation(invocation));
   }
```
liubao68, line 55: This fix seems not correct. HttpClientResponseWrapper/HttpClientRequestWrapper should be invocation scope, not bean scope.  Synchronization can not fix this problem .

yanghao605: Based on my understanding of the brave's source code, I think only this part can set context to invocation.
image
When the handler. handleSend() method is called, it will insert the context related to trace into the invocation.
image
image

And after I added synchronized and tested again, there was no further loss of context

liubao68: synchronization maybe very pool performance. Event your notice is correct, the modification is not good.

PR #4095, opened 2023-12-06
[#4074] optimize cors model init failed unclear prompt problem.

chengyouling:


foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/VertxUtils.java
```diff
@@ -85,16 +86,16 @@ public static <CLIENT_POOL> DeploymentOptions createClientDeployOptions(
   }
 
   // deploy Verticle and wait for its success. do not call this method in event-loop thread
-  public static <VERTICLE extends Verticle> boolean blockDeploy(Vertx vertx,
+  public static <VERTICLE extends Verticle> Map<String, String> blockDeploy(Vertx vertx,
```
liubao68, line 89: return `Map<String, Object>` to avoid data type convert

transports/transport-highway/src/main/java/org/apache/servicecomb/transport/highway/HighwayClient.java
```diff
@@ -63,7 +64,10 @@ public void init(Vertx vertx) throws Exception {
 
     DeploymentOptions deployOptions = VertxUtils.createClientDeployOptions(clientMgr,
         HighwayConfig.getClientThreadCount());
-    VertxUtils.blockDeploy(vertx, ClientVerticle.class, deployOptions);
+    Map<String, String> result = VertxUtils.blockDeploy(vertx, ClientVerticle.class, deployOptions);
+    if (!Boolean.parseBoolean(result.get("code"))) {
+      throw new Exception(result.get("message"));
```
liubao68, line 69: throw new IllegalStageException

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/VertxUtils.java
chengyouling: fixed

transports/transport-highway/src/main/java/org/apache/servicecomb/transport/highway/HighwayClient.java
chengyouling: fixed

PR #4103, opened 2023-12-07
[#4053] add cache for get bean instance by BeanUtils#getBean.

chengyouling:


foundations/foundation-common/src/main/java/org/apache/servicecomb/foundation/common/utils/BeanUtils.java
```diff
@@ -147,7 +150,10 @@ public static <T> T getBean(Class<T> type) {
       // for some test case
       return null;
     }
-    return context.getBean(type);
+    if (beanClass.get(type) == null) {
```
liubao68, line 153: It is not good to modify BeanUtils.  Just fix InvokerUtils is enough.

PR #4185, opened 2024-01-04
[#4184] Router add some log for deal problem.

chengyouling:


governance/src/main/java/org/apache/servicecomb/router/RouterFilter.java
```diff
@@ -58,7 +58,7 @@ public <T> List<T> getFilteredListOfServers(List<T> list,
     PolicyRuleItem invokeRule = routerRuleMatcher.match(targetServiceName, extractor);
 
     if (invokeRule == null) {
-      LOGGER.debug("route management match rule failed");
+      LOGGER.warn("route management hava no rule matched service {}", targetServiceName);
```
liubao68, line 61: Too many logs.

chengyouling: fixed

PR #4424, opened 2024-07-26
[#4423] add switch to control return empty instance when router rule not matched

chengyouling:


demo/demo-cse-v1/consumer/src/main/resources/application.yml
```diff
@@ -75,6 +75,54 @@ servicecomb:
...
+        match:
+          headers:
+            canary:
+              exact: force100
+        route:
+          - weight: 100
+            tags:
+              version: 0.0.3
+      - precedence: 5
+        forceEnabled: true
```
liubao68, line 89: Find a better name, e.g. `emptyProtection` and default to true. if `emptyProtection` is false, return empty list.

liubao68 (approved):

demo/demo-cse-v1/consumer/src/main/resources/application.yml
chengyouling: fixed

PR #4471, opened 2024-08-15
fix: shutdown nacos client when discovery destroy

karsonto:
[ISSUE #4470] Shutdown nacos client properly.


shoothzj (approved):
Nice Catch

service-registry/registry-nacos/src/main/java/org/apache/servicecomb/registry/nacos/NacosDiscovery.java
```diff
@@ -137,7 +138,13 @@ public void run() {
 
   @Override
   public void destroy() {
-
+    if (namingService != null) {
+      try {
+        namingService.shutDown();
+      } catch (NacosException e) {
+        throw new IllegalStateException("destroy process is interrupted.");
```
liubao68, line 145: Do not throw exception, just add some logs and ignore.

PR #4533, opened 2024-09-26
[#4532] Fixed when the servicecomb engine storage pool is faulty, the registration and configuration centers cannot perform HA switchover problem

chengyouling:


clients/http-client-common/src/main/java/org/apache/servicecomb/http/client/common/AbstractAddressManager.java
```diff
@@ -221,42 +194,7 @@ private List<String> getZoneOrRegionAddress() {
...
-    try {
-      return new URI(address);
-    } catch (URISyntaxException e) {
-      LOGGER.error("parse address [{}] failed.", address, e);
-      return null;
-    }
-  }
-
-  protected void findAndRestoreAddress(String address) {
+  public void findAndRestoreAddress(String address) {
```
liubao68, line 197: `findAndRestoreAddress` maybe can change to a better name to `recoverIsolatedAddress`

chengyouling: fixed

PR #4595, opened 2024-11-11
[SCB-2892]add dynamic config for etcd

SweetWuXiaoMei:
add dynamic config for etcd


demo/demo-etcd/provider/pom.xml
```diff
@@ -41,12 +41,28 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.servicecomb</groupId>
+      <artifactId>config-etcd</artifactId>
+      <exclusions>
+        <exclusion>
+          <artifactId>commons-logging</artifactId>
+          <groupId>commons-logging</groupId>
+        </exclusion>
+      </exclusions>
```
liubao68, line 65: Exclude commons-logging in registry-etcd & registry-config

demo/demo-etcd/test-client/pom.xml
```diff
@@ -45,6 +45,22 @@
     <dependency>
       <groupId>org.apache.servicecomb</groupId>
       <artifactId>registry-local</artifactId>
+      <exclusions>
+        <exclusion>
+          <artifactId>commons-logging</artifactId>
+          <groupId>commons-logging</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
liubao68, line 54: registry-local may be not dependent on commons-logging. if it is exclude in registry-local, not here.

demo/demo-etcd/test-client/src/main/java/org/apache/servicecomb/samples/EtcdConfigIT.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(EtcdConfigIT.class);
+
+  @Override
+  public void testRestTransport() throws Exception {
+
+    testEnvironment();
+    testApplication();
+    testService();
+    testVersion();
```
liubao68, line 45: Can you add a test case for priority? e.g. service configuration override application configuration

PR #4781, opened 2025-04-23
Compatible with Spring6.x, Jakarta, JDK21, and Springboot4.x

chengyouling:
Main Adjustment Contents：
1、upgrade spring version to 6.2.5；
2、upgrade spring boot version to 3.4.4；
3、using JDK21；
4、upgrade javax.* to jakarta.* ；


service-registry/registry-service-center/src/main/java/org/apache/servicecomb/serviceregistry/task/MicroserviceInstanceRegisterTask.java
```diff
@@ -113,4 +114,9 @@ public void run() {
       srClient.checkIsolationAddressAvailable();
     }
   }
+
+  @VisibleForTesting
+  public void setAddrCheckInit(boolean addrCheckInit) {
```
yhs0092, line 119: It's not recommended to expose the hiden property for UT purpose. If this method is only used for `TestMicroserviceInstanceRegisterTask`, you can delete the `public` modifier. Just exposing `isAddrCheckInit` property to the classes under the same package seems enough.

chengyouling: fixed

yhs0092 (approved):

PR #4809, opened 2025-05-08
[#4808] fixed log level error cause exception stack cannot be output problem

chengyouling:


foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/client/http/HttpClientPoolFactory.java
```diff
@@ -46,7 +46,7 @@ public HttpClientWithContext createClientPool(Context context) {
                   connection.localAddress(), connection.remoteAddress())
           );
           connection.exceptionHandler(e ->
-              LOGGER.info("http connection exception, local:{}, remote:{}.",
+              LOGGER.error("http connection exception, local:{}, remote:{}.",
```
liubao68, line 49: This may print a lot of logs in situation like: health check, connection pool resize and other.

yhs0092: > This may print a lot of logs in situation like: health check, connection pool resize and other.

Hi, is there real user case? The common health check connection issue is about server side: the Nginx establish a TCP connection to the server and then close it, leaving a connection exception on the backend server side. But this is a client pool, not a server pool.

I guess whether we print the stacktrace depends on whether we have actually seen many such logs. Or at least, we should print the exception filtered by `ExceptionUtils.getExceptionMessageWithoutTrace`.

Currently, the log totally doesn't carry the exception message:
image

liubao68: This log level seems changed several times. A possible solution maybe configure the logger if the user really need it. 
However, I'd like to merge this PR if no other problems found now.

yhs0092: +1, this PR looks good to me

PR #4812, opened 2025-05-13
[#4811] Support logging fileName for first polling configurations

chengyouling:


clients/config-center-client/src/main/java/org/apache/servicecomb/config/center/client/ConfigCenterClient.java
```diff
@@ -66,7 +66,8 @@ public void setEventBus(EventBus eventBus) {
   }
 
   @Override
-  public QueryConfigurationsResponse queryConfigurations(QueryConfigurationsRequest request, String address) {
+  public QueryConfigurationsResponse queryConfigurations(QueryConfigurationsRequest request, String address,
+      boolean isFirstPull) {
     String dimensionsInfo = buildDimensionsInfo(request, true);
     QueryConfigurationsResponse queryConfigurationsResponse = new QueryConfigurationsResponse();
```
liubao68, line 72: I'd prefer not merge this PR because it just add some logs and changed the interface definition. Maybe new solutions can have. e.g. add logs for all config centers and make it optional.

PR #4837, opened 2025-06-09
[#4835] Fixed the fileUpload stream not closed problem

chengyouling:


foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/InputStreamWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    FileUploadStreamRecorder.getInstance().clearRecorder(this);
+    inputStream.close();
+  }
+
+  @Override
+  public synchronized void reset() throws IOException {
```
yhs0092, line 76: `synchronized` on this method and `mark(int readlimit)` seems not necessary, please check whether these `synchronized` words can be removed.

yhs0092 (changes requested):

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadPart.java
```diff
@@ -36,7 +36,10 @@ public FileUploadPart(FileUpload fileUpload) {
 
   @Override
   public InputStream getInputStream() throws IOException {
-    return Files.newInputStream(new File(fileUpload.uploadedFileName()).toPath());
+    final InputStream inputStream = Files.newInputStream(new File(fileUpload.uploadedFileName()).toPath());
+    final InputStreamWrapper inputStreamWrapper = new InputStreamWrapper(inputStream);
+    FileUploadStreamRecorder.getInstance().recordOpenStream(inputStreamWrapper);
```
yhs0092, line 41: How about moving this line into the constructor of `InputStreamWrapper`?
It's not about functionality, but since the `clearRecorder` method is invoked in the `InputStreamWrapper.close()` method, maybe it's better to move the `recordOpenStream` method invocation into the `InputStreamWrapper`, too, so that the `FileUploadStreamRecord` can be encapsulated better.

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+public class FileUploadStreamRecorder {
+  private static final Logger LOGGER = LoggerFactory.getLogger(FileUploadStreamRecorder.class);
+
+  private static final FileUploadStreamRecorder RECORDER = new FileUploadStreamRecorder();
+
+  private static final String STREAM_OPEN_UPPER_LIMIT = "file.upload.stream.operate.upper-limit";
+
+  private static final String STREAM_STACKTRACE_ENABLED = "file.upload.stream.operate.stack-trace-enabled";
+
+  private static final String STREAM_CHECK_INTERVAL = "file.upload.stream.operate.check-interval";
```
yhs0092, line 47: To keep consistent with other Java-Chassis properties, please make these property keys start with `servicecomb.` prefix.

yhs0092 (changes requested):

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private final Map<InputStreamWrapper, StreamOperateEvent> wrapperRecorder = new ConcurrentHashMap<>();
+
+  private final EventBus eventBus;
+
+  private final ScheduledExecutorService streamCheckExecutor;
+
+  private FileUploadStreamRecorder() {
+    eventBus = EventManager.getEventBus();
+    streamCheckExecutor = Executors.newScheduledThreadPool(1, (t) -> new Thread(t, "stream-operate-check"));
```
yhs0092, line 61: Maybe the thread name can be more specific, like "upload-file-stream-operate-check"?

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  private int getStreamOpenUpperLimit() {
+    return DynamicPropertyFactory.getInstance()
+        .getIntProperty(STREAM_OPEN_UPPER_LIMIT, DEFAULT_STREAM_OPEN_UPPER_LIMIT).get();
+  }
+
+  private boolean getStackTraceEnabled() {
+    return DynamicPropertyFactory.getInstance().getBooleanProperty(STREAM_STACKTRACE_ENABLED, true).get();
```
yhs0092, line 139: Better to make it false by default, to avoid performance impact.

yhs0092 (changes requested):

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadPart.java
chengyouling: fixed

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+  public static FileUploadStreamRecorder getInstance() {
+    return RECORDER;
+  }
+
+  public void recordOpenStream(final InputStreamWrapper wrapper) {
+    checkAndRemoveOldestStream();
+    wrapperRecorder.put(wrapper, new StreamOperateEvent(wrapper));
+  }
+
+  private void checkAndRemoveOldestStream() {
```
yhs0092, line 81: According to the current `checkAndRemoveOldestStream` logic, some cornor cases may not be handled well.

For example,
- if the business thread pool is large enough,
- or the controller is developed in the reactive mode, opening file stream and scheduling the stream into a backend thread pool to process (I mean the response `CompletableFuture` is completed when the backend task is finished),

in these two cases, it's possible that more than 1000 file streams are opened in a short period, and `checkAndRemoveOldestStream` method closes the previously opened stream, and the business logic is affected. From the perspective of business developers, these are normal usecases but we break their business logic.

To solve the problem perfectly, it seems necessary to subscribe the `InvocationFinishEvent` and establish the mapping relationships between `Invocation` and `StreamOperateEvent`, which increases the complexity.
Or as a simpler solution, can we just set the default value of the max recorders size bigger ?

yhs0092 (changes requested):

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
chengyouling: change to 5000

foundations/foundation-vertx/src/main/java/org/apache/servicecomb/foundation/vertx/http/FileUploadStreamRecorder.java
```diff
@@ -0,0 +1,191 @@
...
+    int maxSize = getStreamRecorderMaxSize();
+    if (streamWrapperRecorder.size() < maxSize) {
+      return;
+    }
+    synchronized (lock) {
+      StreamOperateEvent oldestEvent = getOldestOperateEvent(streamWrapperRecorder.values());
+      LOGGER.warn("reached recorder maxSize [{}] of file stream, delete oldest stream, operate time [{}], stackTrace {}",
+          maxSize, oldestEvent.getOpenStreamTimestamp(), oldestEvent.getInvokeStackTrace());
+      closeStreamWrapper(oldestEvent.getInputStreamWrapper());
```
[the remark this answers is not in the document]
chengyouling: add eventType for distinguish

yhs0092 (approved):

PR #4849, opened 2025-06-18
[#4848] fixed run the maven-source-plugin command twice problem

chengyouling:


pom.xml
```diff
@@ -520,7 +520,7 @@
               </execution>
             </executions>
             <configuration>
-              <source>1.8</source>
+              <source>21</source>
```
liubao68, line 523: why change this?

chengyouling: The 2.9 branch uses JDK21.

Issue #4946, opened 2025-10-01
[BUG] - Nondeterminism in unit test TestInvocationPublishModelFactory due to JSON parameter ordering


anthonyx24:
### Steps to Reproduce

The method `createDefaultPublishModel()` in `TestInvocationPublishModelFactory` compares JSONs to verify that the default object is properly created. However, since JSONs are unordered collections, when converted to strings the parameter order may not be preserved. This test uses a hard-coded JSON string as the expected correct value, but the test could fail if the parameters from the test object are stringified in a different order than the expected string. For another example of this issue being addressed, see this [previous merged PR](https://github.com/apache/servicecomb-java-chassis/pull/4633).

This test was flagged via the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which detects potentially unreliable tests due to underlying Java API assumptions. To see the Nondex output for this test, you can run:

```
mvn -pl metrics/metrics-core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest="org.apache.servicecomb.metrics.core.publish.TestInvocationPublishModelFactory#createDefaultPublishModel"
```

### Servicecomb Version

3.3.0

### Additional Context

I have a potential fix in a PR, I'll be adding that to the discussion.

PR #4633, opened 2024-12-05
Fixed flaky tests in ExceptionsTest.java with ObjectMapper

mumbler6:
Hello, I apologize for not creating and linking a JIRA issue for this PR yet, though I did request for a JIRA account on the self-serve page if needed. This PR is fairly trivial though and does not change much code. 

### Purpose of this PR
This PR fixes the flaky tests:
* [`org.apache.servicecomb.core.exception.ExceptionsTest.should_convert_unknown_client_exception_to_invocation_exception`](https://github.com/apache/servicecomb-java-chassis/blob/fc7c01668fbb992e015458a86bc4366a00dc5878/core/src/test/java/org/apache/servicecomb/core/exception/ExceptionsTest.java#L46)
* [`org.apache.servicecomb.core.exception.ExceptionsTest.should_convert_unknown_server_exception_to_invocation_exception`](https://github.com/apache/servicecomb-java-chassis/blob/fc7c01668fbb992e015458a86bc4366a00dc5878/core/src/test/java/org/apache/servicecomb/core/exception/ExceptionsTest.java#L59)

The mentioned tests may non-deterministically pass or fail without changes made to the source code when it is run in different JVMs due to the inherent unordered nature of JSON properties in the JSON strings during comparison. The tests expect: 

* `"{\"code\":\"SCB.00000000\",\"message\":\"Unexpected exception when processing none. msg\"}"`
* `"{\"code\":\"SCB.50000000\",\"message\":\"Unexpected exception when processing none. msg\"}"`

But `Json.encode(invocationException.getErrorData())` will sometimes return: 

* `"{\"message\":\"Unexpected exception when processing none. msg\",\"code\":\"SCB.00000000\"}"`
* `"{\"message\":\"Unexpected exception when processing none. msg\",\"code\":\"SCB.50000000\"}"`

### Steps to Reproduce

This change was confirmed by running the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which explores and reports errors in different behaviors of under-determined Java APIs.
```
# Compile the module in project directory
mvn clean install -pl core -am -DskipTests

# Run the unit tests using NonDex
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -pl core -Dtest=org.apache.servicecomb.core.exception.ExceptionsTest#should_convert_unknown_client_exception_to_invocation_exception \
[...]

PR #4947, opened 2025-10-01
[#4946] Change TestInvocationPublishModelFactory to deterministic implementation

anthonyx24:
**Issue:** The method `createDefaultPublishModel()` in `TestInvocationPublishModelFactory` compares JSONs to verify that the default object is properly created. However, since JSONs are unordered collections, when converted to strings the parameter order may not be preserved. This test uses a hard-coded JSON string as the expected correct value, but the test could fail if the parameters from the test object are stringified in a different order than the expected string. For another example of this issue being addressed, see this [previous merged PR](https://github.com/apache/servicecomb-java-chassis/pull/4633).

This test was flagged via the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which detects potentially unreliable tests due to underlying Java API assumptions. To see the Nondex output for this test, you can run:

```
mvn -pl metrics/metrics-core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest="org.apache.servicecomb.metrics.core.publish.TestInvocationPublishModelFactory#createDefaultPublishModel"
```

**Fix:** The fix I implemented uses [JSONAssert's](https://jsonassert.skyscreamer.org/apidocs/org/skyscreamer/jsonassert/JSONAssert.html) `assertEquals()` method, using non-strict checking (which ignores parameter ordering). This compares the JSON strings without enforcing parameter ordering, which removes the potentially unreliable nature of these tests. **This library is already included within the spring-boot dependency, which is already used in the project, so there's no need to update the pom file.** Rerunning Nondex shows a passing result.

I believe this is the simplest fix, but it seems someone else has already fixed a similar test in a different way: https://github.com/apache/servicecomb-java-chassis/pull/4633 using JsonNode trees. If it's better to keep code consistency, I can also implement the same methodology here.

One additional change I had to make was to add the `throws Exception` clause onto the test method signature since `JSONAssert.assertEquals()` can throw a `JSONException`. Please let me know if this is an issue; this clause already exists in the same method in `TestThreadPoolPublishModelFactory` within the same module metrics/metrics-core, and the integration tests from `mvn clean install -Pit` pass.

**PR Checklist:**
 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/SCB) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes. (I created a Github Issue instead since I noticed that the JIRA seems to be inactive)

---

### DIFF
--- a/metrics/metrics-core/src/test/java/org/apache/servicecomb/metrics/core/publish/TestInvocationPublishModelFactory.java
+++ b/metrics/metrics-core/src/test/java/org/apache/servicecomb/metrics/core/publish/TestInvocationPublishModelFactory.java
@@ -30,7 +30,6 @@ import org.apache.servicecomb.metrics.core.InvocationMetersInitializer;
 import org.apache.servicecomb.metrics.core.publish.model.DefaultPublishModel;
 import org.apache.servicecomb.swagger.invocation.InvocationType;
 import org.apache.servicecomb.swagger.invocation.Response;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 import org.springframework.core.env.Environment;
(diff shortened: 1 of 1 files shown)


SweetWuXiaoMei:
Thanks,that's great

Issue #4987, opened 2025-10-21
[BUG] - Nondeterminism in unit test within HighwayServerCodecFilterTest due to JSON parameter ordering


anthonyx24:
### Steps to Reproduce

Similar to #4970 , the unit test `should_convert_exception_to_response_when_decode_request_failed` within `HighwayServerCodecFilterTest` compares JSON strings nondeterministically. In summary, since JSONs are unordered, after converting them to strings, the parameter ordering is not guaranteed, which can lead to the tests failing occasionally.

This test was flagged via the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which detects potentially unreliable tests due to underlying Java API assumptions. To see the Nondex output for `HighwayServerCodecFilterTest`, you can run:
```
mvn -pl transports/transport-highway edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest="org.apache.servicecomb.transport.highway.HighwayServerCodecFilterTest"
```

### Servicecomb Version

3.3.0

PR #4988, opened 2025-10-21
Change unit test in HighwayServerCodecFilterTest to deterministic implementation

anthonyx24:
**Issue:** See #4987.

**Fix:** The fix here is the same as previously merged PRs such as #4969 and #4971. I use [JSONAssert's](https://jsonassert.skyscreamer.org/apidocs/org/skyscreamer/jsonassert/JSONAssert.html) assertEquals() method, using non-strict checking (which ignores parameter ordering). This compares the JSON strings without enforcing parameter ordering, which removes the potentially unreliable nature of these tests. **This library is already included within in the project, so there's no need to update the pom file.** Rerunning Nondex shows a passing result.

**PR Checklist:**
 - [x] Github Issue: https://github.com/apache/servicecomb-java-chassis/issues/4987

---


SweetWuXiaoMei:
lgtm

Issue #4989, opened 2025-10-23
[BUG] - Flaky test `EndpointTest#should_ignore_endpoint_when_generate_swagger`


pem70:
### Steps to Reproduce

The test `should_ignore_endpoint_when_generate_swagger` in `EndpointTest` is flaky because it compares the full serialized YAML string output of the generated OpenAPI object using strict string equality, while the internal data structures cannot guarantee deterministic key ordering.

This test was flagged via the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which detects potentially unreliable tests due to underlying Java API assumptions. To see the Nondex output for this test, run:

```
mvn -pl core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.apache.servicecomb.core.invocation.endpoint.EndpointTest#should_ignore_endpoint_when_generate_swagger
```

### Servicecomb Version

3.3.0

### Additional Context

I have a potential fix in a PR, and I'll add that to the discussion.

pem70:
PR: https://github.com/apache/servicecomb-java-chassis/pull/4990

PR #4990, opened 2025-10-23
Fix Flaky Test `EndpointTest#should_ignore_endpoint_when_generate_swagger`

pem70:
### Issue
Test `EndpointTest#should_ignore_endpoint_when_generate_swagger` is flaky. See #4989

### Fix
While the original test compares the full serialized YAML string, this fix removes the fragile YAML string comparison and checks key fields in the generated OpenAPI object instead. It keeps the same intent as it verifies the expected endpoint and response, while avoiding failures caused by nondeterministic field order or formatting.

### Checklist

---


pem70:
已rebase, ci问题应已解决

PR #5125, opened 2026-04-16
Skip JSON serialization for String data in SSE event stream

lww97:
When SSE response data is already a String, avoid redundant JSON serialization that would wrap it in extra quotes.


lww97:
应用层返回的String时，框架会包裹多出余引号

